PR #2, opened 2019-06-25
5: Remove pipeline support from process

jeyvison:
Hey Folks :) Is that what you meant with removing pipeline support? If it's not, let me know. Also, I think something must be changed in ProcessBuilder, am i right? And this PR has a test problem i'm not able to fix but if any of you give me some directions i can fix it and resend :)

Progress
--------

Approvers
---------

### DIFF
--- a/process/src/main/java/org/openjdk/skara/process/Execution.java
+++ b/process/src/main/java/org/openjdk/skara/process/Execution.java
@@ -28,11 +28,10 @@ import java.time.*;
 import java.util.*;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Logger;
-import java.util.stream.Collectors;
 
 public class Execution implements AutoCloseable {
 
-    private final List<ProcessBuilder> processBuilders;
+    private final ProcessBuilder processBuilder;
     private final Process.OutputOption outputOption;
     private final Duration timeout;
 
@@ -41,13 +40,12 @@ public class Execution implements AutoCloseable {
     private String cmd;
     private int status = 0;
     private File stdoutFile;
-    private List<File> stderrFiles;
+    private File stderrFile;
 
-    private Instant start;
     private boolean finished;
     private Result result;
     private Throwable exception;
-    private List<java.lang.Process> processes;
+    private java.lang.Process process;
 
     public static class CheckedResult {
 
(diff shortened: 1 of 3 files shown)


rwestberg:
> Hey Folks :) Is that what you meant with removing pipeline support? If it's not, let me know. 

Hi! Thanks for the PR! Yes, this is indeed what we meant. (There's actually even a bit more complexity that can be removed, I'll add a more detailed review).

> Also, I think something must be changed in ProcessBuilder, am i right?

Do you mean java.lang.ProcessBuilder? That should not need any changes, perhaps I'm misunderstanding the question..

> And this PR has a test problem i'm not able to fix but if any of you give me some directions i can fix it and resend :)

All tests pass when I run them in your branch, what kind of problem did you see?

process/src/main/java/org/openjdk/skara/process/Execution.java
```diff
@@ -253,11 +206,7 @@ private Result createResult() {
         finished = false;
 
         try {
-            if (processBuilders.size() == 1) {
```
rwestberg: And without pipes, processBuilders don't even need to be a list.

process/src/main/java/org/openjdk/skara/process/Execution.java
```diff
@@ -138,52 +137,6 @@ private void prepareRedirects() throws IOException {
...
-                }
-                throw new InterruptedException("Command '" + cmd + "' didn't finish in " + timeout + ", terminated");
-            }
-            remainingTimeout = remainingTimeout.minus(Duration.between(start, Instant.now()));
-            start = Instant.now();
-        }
-    }
-
     private void startProcess() throws IOException {
         cmd = String.join(" ", getLastProcessBuilder().command());
```
rwestberg: The getLastProcessBuilder method could be removed as there is never more than one now.

jeyvison:
@rwestberg Done :) Hope it's better now.

process/src/main/java/org/openjdk/skara/process/Execution.java
```diff
@@ -107,92 +105,39 @@ public int status() {
...
-        cmd = String.join(" ", getLastProcessBuilder().command());
+        cmd = String.join(" ", processBuilder.command());
         log.fine("Executing '" + cmd + "'");
 
         prepareRedirects();
-        start = Instant.now();
 
         processes = new LinkedList<>();
-        processes.add(getLastProcessBuilder().start());
+        processes.add(processBuilder.start());
```
edvbld: We can also remove the field `private List<java.lang.Process> processes` now, that can just be turned into `private java.lang.Process process` (the list was only there to support pipelines).

process/src/main/java/org/openjdk/skara/process/Execution.java
```diff
@@ -239,25 +184,20 @@ private Result createResult() {
         }
 
 
-        var command = processBuilders.stream()
-                                     .map(pb -> String.join(" ", pb.command()))
-                                     .reduce("", (res, cmd) -> res.isEmpty() ? cmd : res + " | " + cmd);
+        var command = String.join(" ", processBuilder.command());
```
edvbld: Can't you just the field `cmd` here instead that you updated on line 134?

jeyvison:
@edvbld Third time is the luck one. Done :)

jeyvison:
Should https://github.com/openjdk/skara/blob/d76aaa2ff9e8809833137c899ec819977d5903a4/process/src/main/java/org/openjdk/skara/process/Process.java#L40 be just a single command too?

rwestberg:
Btw, I changed the title of the PR to conform to our preferred format (this will be automated in the future).

rwestberg:
> Should
> 
> https://github.com/openjdk/skara/blob/d76aaa2ff9e8809833137c899ec819977d5903a4/process/src/main/java/org/openjdk/skara/process/Process.java#L40
> 
> be just a single command too?

And no, that one is fine, commands (executable and its arguments) are passed to ProcessBuilder as a list: https://github.com/openjdk/skara/blob/d76aaa2ff9e8809833137c899ec819977d5903a4/process/src/main/java/org/openjdk/skara/process/Process.java#L87

PR #7, opened 2019-06-26
Ignore `path=` lines in git credential fill output.

JornVernee:
Hi,

I'm trying out skara, but when running `git fork https://github.com/openjdk/panama git-panama` I'm getting the following exception:

```
Exception in thread "main" java.io.IOException: 'git credential' returned unexpected line: path=
        at org.openjdk.skara.cli/org.openjdk.skara.cli.GitCredentials.fill(GitCredentials.java:101)
        at org.openjdk.skara.cli/org.openjdk.skara.cli.GitFork.main(GitFork.java:130)
        at org.openjdk.skara.cli/org.openjdk.skara.cli.GitSkara.main(GitSkara.java:130)
```

Looks like `path=` is not handled yet?

This PR adds some handling that ignores lines starting with `path=` for now, but maybe any unknown line should be ignored instead of throwing an exception, what do you think?

Progress
--------

Approvers
---------

### DIFF
--- a/cli/src/main/java/org/openjdk/skara/cli/GitCredentials.java
+++ b/cli/src/main/java/org/openjdk/skara/cli/GitCredentials.java
@@ -28,12 +28,14 @@ import java.util.concurrent.TimeUnit;
 
 class GitCredentials {
     private final String host;
+    private final String path;
     private final String username;
     private final String password;
     private final String protocol;
 
-    GitCredentials(String host, String username, String password, String protocol) {
+    GitCredentials(String host, String path, String username, String password, String protocol) {
         this.host = host;
+        this.path = path;
         this.username = username;
         this.password = password;
         this.protocol = protocol;
@@ -41,6 +43,10 @@ class GitCredentials {
 
     String host() {
         return host;
+    }
+
+    String path() {
+        return path;
     }
 
     String username() {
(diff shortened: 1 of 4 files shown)


JornVernee:
Issue created at: https://bugs.openjdk.java.net/browse/SKARA-21

JornVernee:
webrev is a 404 for me...

edvbld:
Thanks @JornVernee for your contribution! :tada: 

The patch looks good, have you tried it with a credential with a `path`? I'm thinking that if the user has a `path` set for a credential then we should probably pass that `path` back to `git credential approve` when in `GitCredentials::approve`. Won't we otherwise set the credential for the entire host, or am I missing something? Looking at the [man page](https://git-scm.com/docs/git-credential) it seems like we should pass the `path` back to `git credential approve`

JornVernee:
Yeah, that's a good point. I hadn't looked into it that much. I figured since there was no handling for `path` it was safe to ignore.

I usually use TortoiseGit, and I'm not really familiar with the `credentials` CLI. Trying to run `git credential fill` manually from the command line seems to hang forever, so I'm not sure how best to test it with a credential with a path.

JornVernee:
Okay, I've tested with a credential with a path as well (had to set `credential.useHttpPath=true` in git config for this), and it seems to be working. I can use `git fork` to clone a repo like that. I also see the credential (including the path) appearing in wincred.

PR #9, opened 2019-06-27
24: Allow lazy evaluation of the fullName field in HostUserDetails

rwestberg:
Hi all,

Please review the following change that allows lazy evaluation of the fullName field in HostUserDetails. Most GitHub API calls only return the id and username, so to populate the full name we have to do a separate request. However, the full name is rarely used, so we can save a few API calls by only fetching it when needed.

Best regards,
Robin

Progress
--------

Approvers
---------

### DIFF
--- a/host/src/main/java/org/openjdk/skara/host/HostUserDetails.java
+++ b/host/src/main/java/org/openjdk/skara/host/HostUserDetails.java
@@ -28,20 +28,18 @@ import java.util.function.Supplier;
 public class HostUserDetails {
     private final int id;
     private final String username;
-    private String name;
     private final Supplier<String> nameSupplier;
+    private String name;
 
     public HostUserDetails(int id, String username, String name) {
         this.id = id;
         this.username = username;
-        this.name = name;
-        this.nameSupplier = null;
+        this.nameSupplier = () -> name;
     }
 
     public HostUserDetails(int id, String username, Supplier<String> nameSupplier) {
         this.id = id;
         this.username = username;
-        this.name = null;
         this.nameSupplier = nameSupplier;
     }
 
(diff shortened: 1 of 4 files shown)


host/src/main/java/org/openjdk/skara/host/HostUserDetails.java
```diff
@@ -63,6 +72,11 @@ public String userName() {
     }
 
     public String fullName() {
+        if (name != null) {
+            return name;
+        }
+        assert nameSupplier != null;
```
edvbld: Perhaps check an throw an exception here instead? Assertions are not enabled by default in the launchers. You can use `Optional` to make this short and sweet:
```
var supplier = Optional.ofNullable(nameSupplier).orElseThrow(IllegalStateException::new)
name = supplier.get();
return name;
```

rwestberg: This can't actually happen (name or namesupplier must be set), it's mostly to make my IDE stop complaining

host/src/main/java/org/openjdk/skara/host/HostUserDetails.java
```diff
@@ -75,7 +73,6 @@ public String fullName() {
         if (name != null) {
             return name;
         }
```
edvbld: Nice improvement, but `fullName` can now be as simple as:
```java
public String fullName() {
    if (name == null) {
        name = nameSupplier.get();
    }
    return name;
}
```

rwestberg: Sure, can change it

rwestberg:
Thanks for reviewing!

PR #10, opened 2019-06-27
Check if census endpoint is a URL before trying to parse it as a path.

JornVernee:
When running `git jcheck --local` I'm getting the following exception:

```
Exception in thread "main" java.nio.file.InvalidPathException: Illegal char <:> at index 5: https://openjdk.java.net/census.xml
        at java.base/sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)
        at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)
        at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
        at java.base/sun.nio.fs.WindowsPath.parse(WindowsPath.java:92)
        at java.base/sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:229)
        at java.base/java.nio.file.Path.of(Path.java:147)
        at org.openjdk.skara.cli/org.openjdk.skara.cli.GitJCheck.main(GitJCheck.java:159)
        at org.openjdk.skara.cli/org.openjdk.skara.cli.GitSkara.main(GitSkara.java:130)
```

Looking into the source code this seems to be because the fallback endpoint for getting the OpenJDK census is used, a URL, which is not a valid path (at least not on Windows), so the check to see if we're dealing with a path or not fails when parsing the path.

This PR switches to a regular expression for checking whether the endpoint is a URL or not (by checking if it starts with `http(s)://`).

Progress
--------

Approvers
---------

### DIFF
--- a/cli/src/main/java/org/openjdk/skara/cli/GitJCheck.java
+++ b/cli/src/main/java/org/openjdk/skara/cli/GitJCheck.java
@@ -34,10 +34,14 @@ import java.io.IOException;
 import java.net.*;
 import java.nio.file.*;
 import java.util.*;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.logging.Level;
 
 public class GitJCheck {
+
+    private static final Pattern urlPattern = Pattern.compile("^https?://.*", Pattern.CASE_INSENSITIVE);
+
     public static void main(String[] args) throws Exception {
         var flags = List.of(
             Option.shortcut("r")
@@ -156,7 +160,9 @@ public class GitJCheck {
                 return fallback;
             }
         });
-        var census = Files.exists(Path.of(endpoint)) ? Census.parse(Path.of(endpoint)) : Census.from(URI.create(endpoint));
+        var census = !isURL(endpoint)
+                ? Census.parse(Path.of(endpoint))
+                : Census.from(URI.create(endpoint));
         var isLocal = arguments.contains("local");
         if (!isLocal) {
             var lines = repo.config("jcheck.local");
@@ -171,5 +177,9 @@ public class GitJCheck {
                 error.accept(visitor);
             }
         }
+    }
+
+    private static boolean isURL(String pathOrURL) {
+        return urlPattern.matcher(pathOrURL).matches();
     }
 }
--- a/cli/src/main/java/org/openjdk/skara/cli/GitJCheck.java
+++ b/cli/src/main/java/org/openjdk/skara/cli/GitJCheck.java
@@ -179,7 +179,7 @@ public class GitJCheck {
         }
     }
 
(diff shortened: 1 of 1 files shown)


cli/src/main/java/org/openjdk/skara/cli/GitJCheck.java
```diff
@@ -172,4 +178,8 @@ public static void main(String[] args) throws Exception {
             }
         }
     }
+
+    private static boolean isURL(String pathOrURL) {
+        return urlPattern.matcher(pathOrURL).matches();
```
edvbld: The regex solution works fine, I just tend to stick with `String::startsWith` for simple cases :smiley: 
What do you think of `return s.startsWith("http://") || s.startsWith("https://");` ? I'm fine with the regex if you prefer that.

cli/src/main/java/org/openjdk/skara/cli/GitJCheck.java
```diff
@@ -172,4 +178,8 @@ public static void main(String[] args) throws Exception {
             }
         }
     }
+
+    private static boolean isURL(String pathOrURL) {
```
edvbld: Nice to add a static method. Very minor nit, I would probably just have went with `String s` here, but I'm fine with `pathOrURL` as well.

cli/src/main/java/org/openjdk/skara/cli/GitJCheck.java
JornVernee: I was using `s.matches("^http://")` before, but checking on the URI specification https://tools.ietf.org/html/rfc3986#section-3.1

> Although schemes are case-
   insensitive, the canonical form is lowercase

That's why I went with the case insensitive regex. If we don't care about case then it's fine to use `startsWith`, but otherwise we'd need to check every combination of upper and lower case, so in that case the regex seems better.

cli/src/main/java/org/openjdk/skara/cli/GitJCheck.java
JornVernee: Yeah, `String s` seems fine. I'll change it.

cli/src/main/java/org/openjdk/skara/cli/GitJCheck.java
edvbld: Ah, yes, you are right, better stick with the regex :

edvbld:
@JornVernee I just nominated you to become a commiter in Skara, see https://mail.openjdk.java.net/pipermail/skara-dev/2019-June/000109.html for details.

JornVernee:
@edvbld Thanks for the nomination :)

PR #14, opened 2019-07-01
Put the credentials lock in the remote repository

rwestberg:
Hi all,

Please review the following change that puts the credentials lock (used when running integration tests) in the remote repository instead of a local file, to allow the tests to be launched from multiple hosts.

Best regards,
Robin

Progress
--------

Approvers
---------

### DIFF
--- a/test/src/main/java/org/openjdk/skara/test/HostCredentials.java
+++ b/test/src/main/java/org/openjdk/skara/test/HostCredentials.java
@@ -214,7 +214,7 @@ public class HostCredentials implements AutoCloseable {
             localRepo.push(lockHash, repo.getUrl(), "testlock");
             log.info("Obtained credentials lock");
 
-            // If no exception occurs, we have obtained the lock
+            // If no exception occurs (such as the push fails), we have obtained the lock
             return true;
         }
     }
@@ -226,7 +226,7 @@ public class HostCredentials implements AutoCloseable {
             Repository localRepo;
             localRepo = Repository.materialize(repoFolder, repo.getUrl(), "testlock");
             localRepo.remove(lockFile);
-            var lockHash = localRepo.commit("Lock", "test", "test@test.test");
+            var lockHash = localRepo.commit("Unlock", "test", "test@test.test");
             localRepo.push(lockHash, repo.getUrl(), "testlock");
         }
     }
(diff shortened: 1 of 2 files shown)


test/src/main/java/org/openjdk/skara/test/HostCredentials.java
```diff
@@ -175,43 +176,78 @@ public HostCredentials(TestInfo testInfo) throws IOException  {
...
+                    log.info("Active lock encountered - waiting");
+                    return false;
                 }
             }
+
+            // The lock either doesn't exist or is stale, try to grab it
+            Files.writeString(lockFile, ZonedDateTime.now().format(DateTimeFormatter.ISO_DATE_TIME), StandardCharsets.UTF_8);
+            localRepo.add(lockFile);
+            var lockHash = localRepo.commit("Lock", "test", "test@test.test");
+            localRepo.push(lockHash, repo.getUrl(), "testlock");
```
edvbld: Maybe add a comment that `localRepo.push` will throw an `IOException` if the push fails?

test/src/main/java/org/openjdk/skara/test/HostCredentials.java
```diff
@@ -175,43 +176,78 @@ public HostCredentials(TestInfo testInfo) throws IOException  {
...
+    }
+
+    private void releaseLock(HostedRepository repo) throws IOException {
+        try (var tempFolder = new TemporaryDirectory()) {
+            var repoFolder = tempFolder.path().resolve("lock");
+            var lockFile = repoFolder.resolve("lock.txt");
+            Repository localRepo;
+            localRepo = Repository.materialize(repoFolder, repo.getUrl(), "testlock");
+            localRepo.remove(lockFile);
+            var lockHash = localRepo.commit("Lock", "test", "test@test.test");
```
edvbld: Please change the commit message comment to "Unlock" to help future debugging

rwestberg: Good point

test/src/main/java/org/openjdk/skara/test/HostCredentials.java
rwestberg: Can clarify the existing comment

PR #27, opened 2019-07-04
41: Bot launcher should detect ev. http proxy in environment

edvbld:
Hi all,

this patch ensures that `BotLauncher.java` correctly sets up corresponding system properties for `http_proxy`, `https_proxy` and `no_proxy`. The patch became a little larger than I intended since I had to break out `HttpProxy` into a new module and package to share it between `cli` and `bots/cli`. I also updated the Gradle `ProxyPlugin` to align with the new `proxy` module (there is unfortunately now way to use `HttpProxy` from the Gradle `ProxyPlugin`).

Thanks,
Erik

## Testing

- [x] Executed `sh gradlew test` successfully on Linux

Progress
--------

Approvers
---------
 * Robin Westberg ([rwestberg](@rwestberg) - **Reviewer**) **Note!** Review applies to 9584f27f9b76603d5275d748fd12e47bc50d5ee9


proxy/build.gradle
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright (c) 2018, 2019, Oracle and/or its affiliates. All rights reserved.
```
rwestberg:
```suggestion
 * Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
```
:)

edvbld:
Thanks for reviewing @rwestberg!

PR #34, opened 2019-08-01
Fix start of hunk range off-by-one error when parsing range string

JornVernee:
I kept running into off-by-one errors in hunk headers when exporting a patch from a git repo, and then importing into a mercurial repo (for pushing, since git repos are currently read-only)

The problem seems to be that unified diffs generated by git report a line number 1 before the modified line if the line count is 0. For instance, if I just remove a few lines from a file, the hunk destination start line is one before the first deleted line. (similarly when only adding lines). And actually, this seems to be wrong according to the spec: https://www.gnu.org/software/diffutils/manual/html_node/Detailed-Unified.html#Detailed-Unified

> An empty hunk is considered to start at the line that follows the hunk. 

The problem is that when hunks are then coalesced together, the starting line, and computed line counts are off-by-one as well, and as a result, the generated patch fails to apply.

I noticed there was already some code dealing with this in HunkCoalescer, but this wasn't catching all the cases. I first made a version that patched the starting line in the missing cases, but it seemed cleaner to try and fix this at the source, i.e. when parsing the Range string from the unified diff. So, that is what I've done in this PR, as well as removing the previous patching code in HunkCoalescer, which was no longer needed.

If wanted I could also go the other route of patching the starting line in HunkCoalescer in the missing cases (there were 4 of these).

Progress
--------

Approvers
---------

### DIFF
--- a/vcs/src/main/java/org/openjdk/skara/vcs/Range.java
+++ b/vcs/src/main/java/org/openjdk/skara/vcs/Range.java
@@ -49,6 +49,12 @@ public class Range {
         var count =
             countString.equals("18446744073709551615") ?  0 : Integer.parseInt(countString);
 
+        if (count == 0 && start != 0) {
+            // start is off-by-one when count is 0.
+            // but if start == 0, a file was added and we need a 0 here.
+            start++;
+        }
+
         return new Range(start, count);
     }
 
--- a/webrev/src/main/java/org/openjdk/skara/webrev/HunkCoalescer.java
+++ b/webrev/src/main/java/org/openjdk/skara/webrev/HunkCoalescer.java
@@ -174,14 +174,6 @@ class HunkCoalescer {
         var sourceCount = sourceEnd - sourceStart;
         var destCount = destEnd - destStart;
 
-        // For some reason git wants the start to be +1 when no lines have changed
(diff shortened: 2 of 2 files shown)


edvbld:
Wow, thanks for digging into this! This is indeed a bug in git, as you probably noticed in my comments. There is also another underflow bug in git when generating combined diffs with zero context (my comment and workaround is right above your fix).

I agree with fixing this within the `vcs` module, the `webrev` module should not have to know about this issue.

Regarding your little tool, would you want to upstream it? I was thinking something like

```bash
$ git transplant --target /path/to/hg/repo <commit>
```
(loosely based on `hg transplant`, but with a `--target` flag instead of `--source` flag)

JornVernee:
> Regarding your little tool, would you want to upstream it? I was thinking something like
> 
> ```shell
> $ git transplant --target /path/to/hg/repo <commit>
> ```
> 
> (loosely based on `hg transplant`, but with a `--target` flag instead of `--source` flag)

Not using a tool for that currently, just exporting a webrev (also used in the review), and then importing the generated `.patch` file with `qimport` & `qpush` in the hg repo.

Maybe having a tool would be nice though...

PR #35, opened 2019-08-01
Convert path separators in patch file to unix explicitly

JornVernee:
I ran into trouble when trying to apply a patch file generated with Windows Git to a mercurial repository through cygwin. It wasn't recognizing the files since the paths in the patch file were using Windows path separators.

The spec doesn't seem to mention path separators: https://www.gnu.org/software/diffutils/manual/html_node/Detailed-Unified.html#Detailed-Unified

But, both the Windows Git and Mercurial clients I have installed are handling unix path separators just fine. So, in this PR I've changed the patch file generators to use unix path separators explicitly, which solves the problem.

Progress
--------

Approvers
---------

### DIFF
--- a/vcs/src/main/java/org/openjdk/skara/vcs/Patch.java
+++ b/vcs/src/main/java/org/openjdk/skara/vcs/Patch.java
@@ -104,10 +104,10 @@ public abstract class Patch {
 
     public void write(BufferedWriter w) throws IOException {
         // header
-        var sourcePath = source.path().isPresent() ?
-            source.path().get().toString() : target.path().get().toString();
-        var targetPath = target.path().isPresent() ?
-            target.path().get().toString() : source.path().get().toString();
+        var sourcePath = pathWithUnixSeps(source.path().isPresent() ?
+            source.path().get() : target.path().get());
+        var targetPath = pathWithUnixSeps(target.path().isPresent() ?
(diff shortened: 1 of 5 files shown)


edvbld:
Hi Jorn,

thanks for testing out the tools using Cygwin!

The patch itself looks good, but maybe this patch should take the opportunity to fix https://bugs.openjdk.java.net/browse/SKARA-7 as well? I would prefer moving away from `Webrev` having its own logic for creating `.patch` files, it should use `Diff::write` and `Diff::toString` instead. I suspect we have the same exact issue in `Diff::write`, so please fix the bug there as well, even if you don't refactor `Webrev` to use `Diff::write` :bowing_man:

JornVernee:
@edvbld I'll take a look :)

JornVernee:
I've just fixed the problem for `Patch::write` as well, since switching `Webrev` to use `Diff::write` is not a simple switch. The latter does not have support for printing hunk context lines it seems. Doesn't seem like a quick fix, and I don't really have time to look into it right now :/

I ran the `vcs` module tests, but this is the one that had a lot of failures. I can say that at least there were no additional failures from the `Patch::write` change, but you might want to test on your side as well.

edvbld:
Ok, thanks, I will take the patch for a spin.

PR #36, opened 2019-08-02
Get resource path from URI to avoid getting spurious leading path separator on Windows

JornVernee:
One more little pathing problem I ran into when running GitWebrev through a debugger.

The old code uses `URL::getPath`, but this results in a leading slash before the path, and on Windows absolute paths contain drive letters, so you get something like `/C:/a/b/c`, which is not a valid path syntax.

We can first convert the URL to a URI, for which `Path::of` has a special overload that does the right thing for us.

Progress
--------

Approvers
---------


JornVernee:
Guess it doesn't recognize me as a committer yet :)

edvbld:
@JornVernee You will be considered committer when https://openjdk.java.net/census is updated :+1:

PR #37, opened 2019-08-07
22: Wrong lines for comments

rwestberg:
Hi all,

Please review the following change that translates the position field in a GitHub review comment into the actual line number, which is expected in our common API.

Best regards,
Robin

Progress
--------

Approvers
---------


vcs/src/main/java/module-info.java
```diff
@@ -26,6 +26,7 @@
     requires org.openjdk.skara.encoding;
 
     exports org.openjdk.skara.vcs;
+    exports org.openjdk.skara.vcs.tools;
```
edvbld:
```suggestion
```

PR #39, opened 2019-08-07
50: Fix failing tests after e8d32c78dc

edvbld:
Hi all,

this patch fixes some test failures after e8d32c78dc was integrated. Turns out there are different bugs in git with regards to unified diffs with zero context :cry: @JornVernee fixed the issues with the regular unified diff printer (when there is exactly one parent), this patch works around the issues with git's combined diff printer. The combined diff printer seems to print _correct_ start for empty hunks as long as the file the hunks belong to isn't added.

I also updated a number of tests to take the changes in e8d32c78dc into account.

## Testing
- [x] Run `sh gradlew test` on Linux x86-64

Progress
--------

Approvers
---------
 * Robin Westberg ([rwestberg](@rwestberg) - **Reviewer**) **Note!** Review applies to 0914742fbd34787c12cc2fab29f69a1954fe65b3

### DIFF
--- a/vcs/src/main/java/org/openjdk/skara/vcs/git/GitCombinedDiffParser.java
+++ b/vcs/src/main/java/org/openjdk/skara/vcs/git/GitCombinedDiffParser.java
@@ -43,7 +43,7 @@ class GitCombinedDiffParser {
         this.delimiter = delimiter;
     }
 
-    private List<List<Hunk>> parseSingleFileMultiParentDiff(UnixStreamReader reader) throws IOException {
+    private List<List<Hunk>> parseSingleFileMultiParentDiff(UnixStreamReader reader, List<PatchHeader> headers) throws IOException {
         assert line.startsWith("diff --combined");
 
         while ((line = reader.readLine()) != null &&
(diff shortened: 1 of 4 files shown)


vcs/src/main/java/org/openjdk/skara/vcs/tools/GitRange.java
```diff
@@ -0,0 +1,52 @@
...
+        var separatorIndex = s.indexOf(",");
+
+        if (separatorIndex == -1) {
+            var start = Integer.parseInt(s);
+            return new Range(start, 1);
+        }
+
+        var start = Integer.parseInt(s.substring(0, separatorIndex));
+
+        // Need to work arond a bug in git where git sometimes print -1
```
rwestberg:
```suggestion
        // Need to work around a bug in git where git sometimes print -1
```

vcs/src/main/java/org/openjdk/skara/vcs/git/GitCombinedDiffParser.java
```diff
@@ -64,7 +64,14 @@ public GitCombinedDiffParser(List<Hash> bases, Hash head, String delimiter) {
             assert words[0].startsWith("@@@");
             var sourceRangesPerParent = new ArrayList<Range>(numParents);
             for (int i = 1; i <= numParents; i++) {
-                sourceRangesPerParent.add(Range.fromString(words[i].substring(1))); // skip initial '-'
+                var header = headers.get(i - 1);
+                if (header.status().isAdded()) {
+                    // git reports wrong start for added files, they should
+                    // always have range (0, 0), but git reports (1,0)
```
rwestberg:
```suggestion
                    // always have range (0,0), but git reports (1,0)
```

edvbld:
Thanks @rwestberg for the comments!

PR #40, opened 2019-08-07
49: Add forwarder bot

edvbld:
Hi all,

this patch adds a "forwarder" bot that can forward commits between repositories and or branches. This is useful for e.g. a sandbox use-case where we want to forwards commits from jdk:master to e.g. sandbox:master.

# Testing
- [x] Run `sh gradlew test` on Linux x86-64
- [x] Added two new unit tests for the bot

I also did some manual testing of the bots to ensure it successfully forwarded commits between two remote repositories.

Progress
--------

Approvers
---------

### DIFF
--- a/bots/forward/src/main/java/org/openjdk/skara/bots/forward/ForwardBot.java
+++ b/bots/forward/src/main/java/org/openjdk/skara/bots/forward/ForwardBot.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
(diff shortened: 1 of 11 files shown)


bots/forward/src/main/java/org/openjdk/skara/bots/forward/ForwardBot.java
```diff
@@ -0,0 +1,103 @@
...
+        var otherBot = (ForwardBot) other;
+        return !toHostedRepo.getName().equals(otherBot.toHostedRepo.getName());
+    }
+
+    @Override
+    public void run(Path scratchPath) {
+        try {
+            var sanitizedUrl =
+                URLEncoder.encode(toHostedRepo.getUrl().toString(), StandardCharsets.UTF_8);
+            var toDir = scratchPath.resolve("forward").resolve(sanitizedUrl);
```
rwestberg: Perhaps it would be better to use permanent storage here (like the hgbridge) to avoid excessive copies when running concurrently.

edvbld: Good idea, I pushed a new commit that utilizes shared storage instead.

PR #49, opened 2019-08-16
Proper parsing of git diff --raw lines

edvbld:
Hi,

this patch introduces stricter (and more correct) parsing of the output from `git diff --raw`. The previous parsing code could not handle spaces in file names.

## Testing
- [x] `sh gradlew test`

Progress
--------

Approvers
---------
 * Robin Westberg ([rwestberg](@rwestberg) - **Reviewer**) **Note!** Review applies to e95a4ccd65a55e6b0cf2cea87a127a8e9bbae2c0


edvbld:
Robin only says "Ship it" to be polite, he did tell me to add a test, and so I did! Thanks @rwestberg for enforcing regression tests :+1:

rwestberg:
Even better!

PR #66, opened 2019-08-22
Upgrade JUnit 5 to 5.5.1

sormuras:
Upgrade JUnit 5 to Jupiter 5.5.1 and Platform 1.5.1

- Set Jupiter version number 5.5.1
- Introduce Platform Launcher 1.5.1 test runtime dependency in order to force Gradle to load the module from the module-path
- Add missing "requires 'org.junit.jupiter.params'" directives in modules: jcheck, storage, vcs, and webrev

Solves https://bugs.openjdk.java.net/projects/SKARA/issues/SKARA-69

Progress
--------

Approvers
---------

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -43,9 +43,12 @@ configure(subprojects.findAll() { it.name != 'bots' }) {
     }
 
     dependencies {
-        testImplementation 'org.junit.jupiter:junit-jupiter-api:5.3.1'
-        testImplementation 'org.junit.jupiter:junit-jupiter-params:5.3.1'
-        testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.3.1'
+        testImplementation 'org.junit.jupiter:junit-jupiter-api:5.5.1'
+        testImplementation 'org.junit.jupiter:junit-jupiter-params:5.5.1'
+        testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.5.1'
+        // Force Gradle to load the JUnit Platform Launcher from the module-path, as
+        // configured in buildSrc/.../ModulePlugin.java -- see SKARA-69 for details.
+        testRuntimeOnly 'org.junit.platform:junit-platform-launcher:1.5.1'
     }
 
     test {
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ To check that everything works as expected, run the command `git skara help`.
 
 ## Testing
 
-[JUnit](https://junit.org/junit5/) 5.3.1 or later is required to run the unit
+[JUnit](https://junit.org/junit5/) 5.5.1 or later is required to run the unit
 tests. To run the tests, execute following command from the source tree root:
 
 ```bash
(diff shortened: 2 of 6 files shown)


build.gradle
```diff
@@ -43,9 +43,12 @@ configure(subprojects.findAll() { it.name != 'bots' }) {
...
     dependencies {
-        testImplementation 'org.junit.jupiter:junit-jupiter-api:5.3.1'
-        testImplementation 'org.junit.jupiter:junit-jupiter-params:5.3.1'
-        testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.3.1'
+        testImplementation 'org.junit.jupiter:junit-jupiter-api:5.5.1'
+        testImplementation 'org.junit.jupiter:junit-jupiter-params:5.5.1'
+        testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.5.1'
+        // Force Gradle to load the JUnit Platform Launcher from the module-path, as
+        // configured in buildSrc/.../ModulePlugin.java -- see SKARA-69 for details.
+        testRuntimeOnly 'org.junit.platform:junit-platform-launcher:1.5.1'
```
edvbld: Hmm, I don't really follow here - how does adding a test runtime dependency on org.junit.platform:junit-platform-launcher:1.5.1 force Gradle to load the JUnit Platform launcher from the module path? Did you forgot to `git add` a change made to `buildSrc/.../ModulePlugin.java`?

sormuras: > Did you forgot to `git add` a change...

No. The re-configuration of the Gradle `test` task is good as-is.

IIRC and locally passing tests support that, Gradle uses JUnit Platform Launcher (and its dependencies) from the project dependencies -- and only injects one (on the class-path) if a project doesn't provide it.

With `org.junit.platform:junit-platform-launcher:1.5.1` a) being present on the project's class-path and b) being moved to the module-path by the [`module` plugin](https://github.com/openjdk/skara/blob/master/buildSrc/module/src/main/java/org/openjdk/skara/gradle/module/ModulePlugin.java#L106), things work out again.

Correct me if I'm wrong here, @marcphilipp, @britter, et al.

edvbld: Ah ok, I get, thanks for the explanation @sormuras! I'm pretty busy today, but I will take the patch for a spin early next week :+1:

marcphilipp: Gradle always puts the version of the junit-platform-launcher it ships with on the classpath. Adding `org.junit.platform:junit-platform-launcher:1.5.1` to the module path seems to hide that version from the classloader which now loads the version on the module path instead. This version is a proper module and may access the `org.junit.platform.commons.util` package of the junit-platform-commons module.

sormuras:
Before making this PR integration-ready, shall I update to JUnit 5.5.2 right away? We fixed two minor bugs: https://junit.org/junit5/docs/5.5.2/release-notes/

edvbld:
Lets do that as a separate PR, I will have to test that one as well. Now that I've finally finished testing and reviewing this one lets integrate it :+1:

PR #67, opened 2019-08-22
Make ext.py compatible with hg5 api

JornVernee:
The HG api changes in version 5 to drop the first 2 parameters to `mercurial.match.exact`, which breaks the extension script.

This PR detects whether we're running on hg 5, and if so, calls the function without the first 2 arguments instead.

Progress
--------

Approvers
---------
 * Erik Helin ([ehelin](@edvbld) - **Reviewer**) **Note!** Review applies to 91fb200e300ecabd42980b4fd9add2de7ce40f86

### DIFF
--- a/vcs/src/main/resources/ext.py
+++ b/vcs/src/main/resources/ext.py
@@ -22,8 +22,12 @@
 import mercurial
 import mercurial.patch
 import mercurial.mdiff
+import mercurial.util
 import difflib
 import sys
+
+# space separated version list
+testedwith = '4.9.2 5.0.2'
 
 def mode(fctx):
     flags = fctx.flags()
@@ -51,6 +55,15 @@ def write(s):
 def writeln(s):
     write(s)
     sys.stdout.write(encode('\n'))
+
+def _match_exact(root, cwd, files, badfn=None):
+    """
+    Wrapper for mercurial.match.exact that ignores some arguments based on the used version
+    """
+    if mercurial.util.versiontuple() < 5:
+        return mercurial.match.exact(root, cwd, files, badfn)
+    else:
+        return mercurial.match.exact(files, badfn)
 
 def _diff_git_raw(repo, ctx1, ctx2, modified, added, removed):
     nullHash = '0' * 40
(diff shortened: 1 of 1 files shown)


vcs/src/main/resources/ext.py
```diff
@@ -22,9 +22,13 @@
 import mercurial
 import mercurial.patch
 import mercurial.mdiff
+import mercurial.util
 import difflib
 import sys
 
+# space separated version list
+testedwith = '4.9.2 5.0.2'
+
```
edvbld: Does a space separated string named `testedWith` mean something special in Mercurial? Otherwise I would probably just have this a comment instead of a variable (or if you _really_ want a variable, why not use a `set`?)

JornVernee: It's used to mark supported versions per: https://www.mercurial-scm.org/wiki/WritingExtensions#Marking_compatible_versions

JornVernee: Not super needed, but the back trace I got earlier from the extension didn't include the supported versions. This adds the supported version to the backtrace e.g.:

```
** Unknown exception encountered with possibly-broken third-party extension log-git
** which supports versions 4.9 of Mercurial.
```

Seemed nice to have...

edvbld: Ok, sounds good, lets keep it then :+1:

vcs/src/main/resources/ext.py
```diff
@@ -52,6 +56,15 @@ def writeln(s):
     write(s)
     sys.stdout.write(encode('\n'))
 
+def _match_exact(root, cwd, files, badfn=None):
+    """
+    Wrapper for mercurial.match.exact that ignores some arguments based on the used version
+    """
+    if mercurial.util.versiontuple() < 5:
```
edvbld: Two questions here:
1. How does this check work? `versiontuple()` returns a tuple, e.g. `(4, 9, 2)`, how can you compare a tuple to an `int`?
2. `mercurial.util.versiontuple()` is fairly new (added 2015) whereas `mercurial.util.version()` has been around since 2009. Is it possible to implement the check using `mercurial.util.version()` instead? Like `mercurial.util.version().startswith("5")`?

JornVernee: 1. My bad, supposed to compare the first element, which is not quite what happens when comparing with an `int` (though it works in the specific case I tested).
2. Yeah, I was afraid of that. Will switch to using `version()`. (could also go with the `try ... except` approach if you prefer)

edvbld: 1. No worries, that is why we review :+1:
2. Personally I would prefer something like `mercurial.util.version().startswith("5")` _if_ that approach works, I didn't check too carefully what `mercurial.util.version()` returns

JornVernee: I did (and then I noticed this shiny `versiontuple`, it just returns a string like `5.0.2`

edvbld: Ok, great!

JornVernee:
@edvbld There was one other call to `mercurial.match.exact` which I forgot to change earlier (but was caught by another test)... changed that as well.

Still good to go?

edvbld:
@JornVernee yep, still good!

PR #68, opened 2019-08-22
A repository without refs but with objects is healthy

rwestberg:
Hi all,

Please review this minor fix that makes a repository without refs but with some objects flagged as healthy.

Best regards,
Robin

Progress
--------

Approvers
---------


rwestberg:
Thanks @edvbld

PR #70, opened 2019-08-22
Do not catch NoSuchElementException in HgOpenJDKImport

edvbld:
Hi all,

this patch removes catching `NoSuchElementException` in the `main` method for `HgOpenJDKImport` to ease debugging.

## Testing

Thanks,
Erik

Progress
--------

Approvers
---------

### DIFF
--- a/cli/src/main/java/org/openjdk/skara/cli/HgOpenJDKImport.java
+++ b/cli/src/main/java/org/openjdk/skara/cli/HgOpenJDKImport.java
@@ -36,11 +36,17 @@ import java.util.function.*;
 import java.util.logging.*;
 
 public class HgOpenJDKImport {
-    private static Supplier<NoSuchElementException> error(String fmt, Object... args) {
-        return () -> new NoSuchElementException(String.format(fmt, args));
+    static class ErrorException extends RuntimeException {
+        ErrorException(String s) {
+            super(s);
+        }
     }
 
-    public static void main(String[] args) throws IOException, NoSuchElementException {
+    private static Supplier<ErrorException> error(String fmt, Object... args) {
+        return () -> new ErrorException(String.format(fmt, args));
+    }
+
+    public static void main(String[] args) throws IOException {
         var flags = List.of(
             Switch.shortcut("")
                   .fullname("verbose")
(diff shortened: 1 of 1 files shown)


cli/src/main/java/org/openjdk/skara/cli/HgOpenJDKImport.java
```diff
@@ -36,16 +36,11 @@
 import java.util.logging.*;
 
 public class HgOpenJDKImport {
-    private static void die(Exception e) {
-        System.err.println(e.getMessage());
-        System.exit(1);
-    }
-
     private static Supplier<NoSuchElementException> error(String fmt, Object... args) {
```
rwestberg: Pre-existing, but it's a bit unclear why NoSuchElementException is used - plain RuntimeException (or Exception) would probably make more sense? Or perhaps something like NoSuchFileException?

cli/src/main/java/org/openjdk/skara/cli/HgOpenJDKImport.java
```diff
@@ -36,16 +36,11 @@
...
         return () -> new NoSuchElementException(String.format(fmt, args));
     }
 
-    public static void main(String[] args) throws IOException {
+    public static void main(String[] args) throws IOException, NoSuchElementException {
```
rwestberg: And if you go with a RuntimeException-based one I guess there's no need to declare it here either.

cli/src/main/java/org/openjdk/skara/cli/HgOpenJDKImport.java
edvbld: Yes, that is a bit unclear to me as well, and I even wrote the code in the first place :laughing: I _think_ I wanted my `catch` clause to catch `NoSuchElementException` so I could just do `.orElseThrow()` or `.get()` and still display a somewhat nice error message automatically. Pretty bad idea in retrospect, I will rework this.

cli/src/main/java/org/openjdk/skara/cli/HgOpenJDKImport.java
edvbld: Yes, you are totally right, thanks!

PR #71, opened 2019-08-22
Force unix line endings when writing patch files

JornVernee:
Mercurial needs its patch files to end with unix line endings, so we need to use those explicitly when writing patch files (I've also been told on the hg irc that this is according to patch file spec).

Progress
--------

Approvers
---------


edvbld:
Hi Jorn,

thanks for fixing this! If this patch fixes some tests to pass on Windows, which tests are now passing that didn't before this patch?

JornVernee:
@edvbld Particularly; `GitToHgConverterTests::convertMergeCommitWithP0Diff` and `GitToHgConverterTests::convertMergeCommit`, which both failed while applying the patch.

PR #74, opened 2019-08-23
Expect carriage return to be part of lines when testing on Windows.

JornVernee:
One more patch to address failing `vcs` tests on Windows.

The tests in `RepositoryTest` use `Files::write` to create some test files, but this also means the files will have Windows line endings (i.e. CRLF). When these are parsed by the skara tooling the LF part is dropped as a delimiter, but the CR part ends up being part of the line. This is all correct and as expected, since patch files are supposed to have unix line endings, so the carriage return ends up as part of the content.

The problem comes when testing what is written to the file. The tests are not expecting the Windows line endings to be there, so a lot of them fail.

This PR changes the tests to expect the Windows line endings as well. Or rather, it tries to stick what's in the system line separator (minus a trailing '\n', since that is dropped as a delimiter) at the end of the checked line.

With this patch, I have all the `vcs` tests passing locally.

Progress
--------

Approvers
---------

### DIFF
--- a/vcs/src/test/java/org/openjdk/skara/vcs/RepositoryTests.java
+++ b/vcs/src/test/java/org/openjdk/skara/vcs/RepositoryTests.java
@@ -311,7 +311,7 @@ public class RepositoryTests {
             assertEquals(new Range(0, 0), hunk.source().range());
             assertEquals(new Range(1, 1), hunk.target().range());
 
-            assertEquals(List.of(), hunk.source().lines());
+            assertLinesEquals(List.of(), hunk.source().lines());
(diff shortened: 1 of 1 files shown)


edvbld:
First of all, thank so much @JornVernee for getting the `vcs` tests to pass on Windows!

Regarding your patch, what do you think about the static method `assertLinesEquals` instead of having to use `suffixPlatformLineEndings` everywhere? This way the tests reads a bit nicer, like `assertLinesEquals(List.of("Hello, readme!"), hunk.target().lines())`

PR #78, opened 2019-08-23
Add ReadOnlyRepository::dump

edvbld:
Hi all,

this patch adds the `ReadOnlyRepository::dump` method and a corresponding unit test.

## Testing

Thanks!
Erik

Progress
--------

Approvers
---------


vcs/src/main/java/org/openjdk/skara/vcs/ReadOnlyRepository.java
```diff
@@ -66,6 +66,7 @@
         return files(h, Arrays.asList(paths));
     }
 
+    boolean dump(FileEntry entry, Path to) throws IOException;
```
rwestberg: Is it useful to have a return value? Perhaps you are planning to use it in the future, but if so, it should probably have a test..

PR #86, opened 2019-08-26
Fine-tune testing on Windows

sormuras:
### All Windows issues down

- **Skip all bots tests on Windows**
> [...] and since the bots are not really required to run on Windows any ways, it would also be fine to disable all the bots tests from running on Windows, to avoid any further compatibility problems.

- **Convert relative folder path to be URI-friendly**
On Windows Path::toString generates back slashes as path element separators. This yields invalid URI paths: like `"%s\webrev.%s"`.

Progress
--------

Approvers
---------
 * Jorn Vernee ([jvernee](@JornVernee) - Committer)

### DIFF
--- a/bots/hgbridge/src/test/java/org/openjdk/skara/bots/hgbridge/BridgeBotTests.java
+++ b/bots/hgbridge/src/test/java/org/openjdk/skara/bots/hgbridge/BridgeBotTests.java
@@ -31,6 +31,7 @@ import org.openjdk.skara.vcs.*;
 import org.openjdk.skara.vcs.openjdk.convert.*;
 
 import org.junit.jupiter.api.*;
+import org.junit.jupiter.api.condition.*;
 
 import java.io.*;
 import java.net.URI;
@@ -43,6 +44,7 @@ import java.util.stream.Collectors;
 
 import static org.junit.jupiter.api.Assertions.*;
 
+@DisabledOnOs(OS.WINDOWS)
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 class BridgeBotTests {
     private List<String> runHgCommand(Repository repository, String... params) throws IOException {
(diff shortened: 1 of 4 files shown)


JornVernee:
Thanks for looking into this! With your patch all the `mlbridge` tests are passing for me. I'm not seeing the same failures as you with the `mlbridge` tests...

I'm not a Skara reviewer, so I can't approve this, but the patch looks good to me 👍

---
FWIW, I also got some other failures from some of the bot tests, particularly `IntegrateTests::autorebase`, `MergeTests::branchMergeRebase` and `SponsorTests::autoRebase` in the `bots.pr` module. When looking into it this seemed to be caused by file paths exceeding the maximum path length on Windows. Git rejects them with a specific error message, and HG just throws a file not found error.

Besides that some tests in the `bots.submit` module are failing as well, but I haven't determined the exact reason yet.

I had a chat with Erik about this, and since the bots are not really required to run on Windows any ways, it would also be fine to disable all the `bots` tests from running on Windows, to avoid any further compatibility problems. Or, just aggressively disable any that are failing.

sormuras:
> [...] it would also be fine to disable all the bots tests from running on Windows [...]

As long as some CI service is running those tests on Linux/Mac, fine with me.

I'll extend this PR with to disable all bots tests on Windows.

edvbld:
This looks good, but @sormuras would you mind updating the title of the PR? The title of the PR will be used as the title for the resulting commit, and I think we would want something a bit more descriptive than `Windows again...` :smile:

sormuras:
Sure, Erik.

But I'll update the PR once more to include comments from above:
- Disable all bots test
- Revert not needed optimization -- Gradle does just fine, only IDEA executes disabled tests.
> Due to @BeforeAll void setup() (and teardown()) being called anyway, a second gatekeeper in form an assumption is added to this method as well. This improves the execution duration time of all disabled tests.

edvbld:
Ok, sounds good!

sormuras:
Sneak-peek: https://gradle.com/s/u4a6jzvrpkcvs

sormuras:
Instead of annotating each test class in every `bots` project, I implemented an `ExecutionCondition` and enabled auto-detection of provided Jupiter extension. That implementation is named `DisableAllBotsTestsOnWindows` and resides in the `org.openjdk.skara.test` module. It checks the current test classes' package name. If that package name starts with `"org.openjdk.skara.bots."` and the current operating system is Windows, the execution of all tests within the test class are disabled.

Too much magic?

PR #99, opened 2019-08-28
84: Default bug ID format should match the JDK default

kevinrushforth:
Fix for [SKARA-84](https://bugs.openjdk.java.net/browse/SKARA-84) to change the default issue pattern to require exactly 7 digits with no prefix to match `hg jcheck`.

I updated all of the tests to account for this and added new tests for now-invalid formats.

Progress
--------

### DIFF
--- a/bots/notify/src/test/java/org/openjdk/skara/bots/notify/UpdaterTests.java
+++ b/bots/notify/src/test/java/org/openjdk/skara/bots/notify/UpdaterTests.java
@@ -80,7 +80,7 @@ class UpdaterTests {
             TestBotRunner.runPeriodicItems(notifyBot);
             assertEquals(List.of(), findJsonFiles(jsonFolder, ""));
 
-            var editHash = CheckableRepository.appendAndCommit(localRepo, "One more line", "12345678: Fixes");
+            var editHash = CheckableRepository.appendAndCommit(localRepo, "One more line", "1234567: Fixes");
             localRepo.push(editHash, repo.getUrl(), "master");
             TestBotRunner.runPeriodicItems(notifyBot);
             var jsonFiles = findJsonFiles(jsonFolder, "");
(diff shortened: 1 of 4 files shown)


kevinrushforth:
I filed [SKARA-86](https://bugs.openjdk.java.net/browse/SKARA-86) to associate my GitHub username, `kevinrushforth`, with my OpenJDK username, `kcr`.

edvbld:
Hi Kevin,

first of all, thanks for contributing! Unfortunately there are dependencies on the regex your are changing in the rest of the code, particularly in the `HgToGitConverter` (via `ConverterCommitMessageParser`). That means that your change will cause the conversion for some OpenJDK Mercurial repositories to start to fail, so we can't integrate this patch as is :(

I think a better approach here would be to implement https://bugs.openjdk.java.net/projects/SKARA/issues/SKARA-80, but _not_ at the `vcs` level (`vcs` is one "layer" below the `jcheck` layer). Instead I would recommend enhancing git-jcheck's configuration file to allow customizing the `issue` check, for example:
```
[checks "issues"]
pattern = ^([1-9][0-9]{6}): (\\S.*)$
```
Then you could extend `ChecksConfiguration` to have an `IsssuesConfiguration` member (see for example how `ReviewersConfiguration` or `WhitespaceConfiguration` works). Finally you would then have to update the `IssuesCheck` to read this configuration and the pattern and check that all issues in `message.issues()` conforms to the pattern, for example:
```
var issues = new ArrayList<Issue>();
var patttern = Patter.compile(conf.checks().issues().pattern());
for (var issue : message.issues()) {
    if (!pattern.matcher(issue.toString()).matches()) {
        issues.add(new IssuesIssue(metadata));
    }
}
return issues.iterator();
```
The last part that would then fulfill what you want this patch to do would be have the default value for `pattern` in `IssueConfiguration` be `"^([1-9][0-9]{6}): (\\S.*)$"`.

To try to make the architecture a bit more clear: the `vcs` layer should work for _all_ OpenJDK Projects, not just the JDK Project. Then projects can specify how they want jcheck to work for their project via .jcheck/conf. The good news is that you can still re-use all the updates you made to the unit tests for a patch implementing the above!

kevinrushforth:
Got it. Thanks for the explanation. Yes, this does seem better to be `jcheck`-specific. I'll close this PR then.

PR #102, opened 2019-08-29
TestHostedRepository::getPullRequests should only return open PRs

edvbld:
Hi,

this patch fixes `TestHostedRepository::getPullRequests` so that it only returns pull requests with state `PullRequest.State.OPEN` - this is how GitLab and GitHub works.

Thanks,
Erik

Progress
--------

Approvers
---------
 * Robin Westberg ([rwestberg](@rwestberg) - **Reviewer**) **Note!** Review applies to d3b0c0ab51354afee83b0d0ac8a6b4edf78b43d4

### DIFF
--- a/test/src/main/java/org/openjdk/skara/test/TestPullRequest.java
+++ b/test/src/main/java/org/openjdk/skara/test/TestPullRequest.java
@@ -259,7 +259,7 @@ public class TestPullRequest implements PullRequest {
         data.lastUpdate = ZonedDateTime.now();
     }
 
-    public boolean isOpen() {
+    boolean isOpen() {
         return data.state.equals(PullRequest.State.OPEN);
     }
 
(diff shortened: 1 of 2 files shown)


test/src/main/java/org/openjdk/skara/test/TestPullRequest.java
```diff
@@ -254,9 +255,14 @@ public void updateCheck(Check updated) {
 
     @Override
     public void setState(State state) {
+        data.state = state;
         data.lastUpdate = ZonedDateTime.now();
     }
 
+    public boolean isOpen() {
```
rwestberg:
```suggestion
    boolean isOpen() {
```
Can perhaps keep this one package private.

edvbld:
Thanks for the suggestion!

PR #105, opened 2019-08-29
Add topological merge bot

JornVernee:
This PR adds a bot that does a topological merge of the branches in a repo.

The branches can declare a dependencies file, which lists the branches that they depend on. This bot will crawl the branches, collect the dependencies for each branch, and topologically sort them based on their dependencies. Following that it will attempt to merge each dependency into the dependent in this order (this is mainly done so that we get less merges/failures if one of the root merges fails).

Branches that do not declare a dependency file implicitly depend on the master branch. Therefore the list of branches that the bot considers is passed in during configuration.

---

Aside from that, it also fixes a minor problem with `Repository::clone` on Windows.

Progress
--------

Approvers
---------
 * Maurizio Cimadamore ([mcimadamore](@mcimadamore) - no project role) **Note!** Review applies to 4a8f36108e9084e2e58d8546eb2c6b26687ee816

### DIFF
--- a/bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalBot.java
+++ b/bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalBot.java
@@ -1,0 +1,244 @@
+/*
+ * Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package org.openjdk.skara.bots.topological;
+
+import org.openjdk.skara.bot.*;
+import org.openjdk.skara.host.*;
+import org.openjdk.skara.vcs.*;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.UncheckedIOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
+import java.nio.file.Files;
+import java.net.URLEncoder;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Bot that automatically merges any changes from a dependency branch into a target branch
+ */
+class TopologicalBot implements Bot, WorkItem {
+    private final Logger log = Logger.getLogger("org.openjdk.skara.bots");
+    private final Path storage;
+    private final HostedRepository hostedRepo;
+    private final List<Branch> branches;
+    private final String depsFileName;
+
+    TopologicalBot(Path storage, HostedRepository repo, List<Branch> branches, String depsFileName) {
+        this.storage = storage;
+        this.hostedRepo = repo;
+        this.branches = branches;
+        this.depsFileName = depsFileName;
+    }
+
+    @Override
+    public boolean concurrentWith(WorkItem other) {
+        if (!(other instanceof TopologicalBot)) {
+            return true;
+        }
+        var otherBot = (TopologicalBot) other;
+        return !hostedRepo.getName().equals(otherBot.hostedRepo.getName());
+    }
+
+    @Override
+    public void run(Path scratchPath) {
+        log.info("Starting topobot run");
+        try {
+            var sanitizedUrl = URLEncoder.encode(hostedRepo.getWebUrl().toString(), StandardCharsets.UTF_8);
+            var dir = storage.resolve(sanitizedUrl);
+            Repository repo;
+            if (!Files.exists(dir)) {
+                log.info("Cloning " + hostedRepo.getName());
+                Files.createDirectories(dir);
+                repo = Repository.clone(hostedRepo.getUrl(), dir);
+            } else {
+                log.info("Found existing scratch directory for " + hostedRepo.getName());
+                repo = Repository.get(dir)
+                        .orElseThrow(() -> new RuntimeException("Repository in " + dir + " has vanished"));
+            }
+
+            repo.fetchAll();
+            var depsFile = repo.root().resolve(depsFileName);
+
+            List<Branch> ordered = orderedBranches(repo, depsFile);
+            log.info("Merge order " + ordered);
+            for (Branch branch : ordered) {
+                log.info("Processing branch " + branch + "...");
+                repo.checkout(branch);
+                Set<String> parents = new HashSet<>(
+                        Files.exists(depsFile) ? Files.readAllLines(depsFile) : List.of("master"));
+                List<String> failedMerges = new ArrayList<>();
+                boolean progress;
+                boolean failed;
+                do {
+                    // We need to attempt merge parents in any order that works. Keep merging
+                    // and pushing, until no further progress can be made.
+                    progress = false;
+                    failed = false;
+                    for (var parentsIt = parents.iterator(); parentsIt.hasNext();) {
+                        String parent = parentsIt.next();
+                        try {
+                            mergeIfAhead(repo, branch, parent);
+                            progress = true;
+                            parentsIt.remove(); // avoid doing pointless merges
+                        } catch(IOException e) {
+                            log.severe("Merge with " + parent + " failed. Reverting...");
+                            repo.abortMerge();
+                            failedMerges.add(branch + " <- " + parent);
+                            failed = true;
+                        }
+                    }
+                } while(progress && failed);
+
+                if (!failedMerges.isEmpty()) {
+                    throw new IOException("There were failed merges:\n" + failedMerges);
+                }
+            }
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+        log.info("Ending topobot run");
+    }
+
+    private List<Branch> orderedBranches(Repository repo, Path depsFile) throws IOException {
+        List<Edge> deps = new ArrayList<>();
+        for (Branch branch : branches) {
+            repo.checkout(branch);
+            if (Files.exists(depsFile)) {
+                Files.lines(depsFile)
+                        .forEach(dep -> deps.add(new Edge(new Branch(dep), branch)));
+            } else {
+                deps.add(new Edge(new Branch("master"), branch));
+            }
+        }
+        return tsort(deps).stream()
+            .filter(branch -> !branch.name().equals("master"))
+            .collect(Collectors.toList());
+    }
+
+    private void mergeIfAhead(Repository repo, Branch branch, String parent) throws IOException {
+        if (log(repo, branch.name(), parent).count() != 0) {
+            var fromHash = repo.resolve(parent).orElseThrow();
+            var isFastForward = repo.isAncestor(repo.head(), fromHash);
+            repo.merge(fromHash);
+            if (!isFastForward) {
+                log.info("Merged " + parent + " into " + branch);
+                repo.commit("Automatic merge with " + parent, "topobot", "");
+            } else {
+                log.info("Fast forwarded " + branch + " to " + parent);
+            }
+            log.info("merge with " + parent + " succeeded. The following commits will be pushed:\n"
+                    + log(repo, "origin/" + branch.name(), branch.name()).collect(Collectors.joining("\n", "\n", "\n")));
+            try {
+                repo.push(repo.head(), hostedRepo.getUrl(), branch.name());
+            } catch (IOException e) {
+                log.severe("Pusing failed! Aborting...");
+                repo.abortMerge();
+                throw e;
+            }
+        }
+    }
+
+    private static Stream<String> log(Repository repo, String targetRef, String fromRef) throws IOException {
+        var process = new ProcessBuilder()
+                .command("git", "log", targetRef + ".." + fromRef, "--")
+                .directory(repo.root().toFile())
+                .start();
+        try {
+            int exit = process.waitFor();
+            if (exit != 0) {
+                throw new IOException("Unexpected exit code: " + exit + "\n\n"
+                        + new BufferedReader(new InputStreamReader(process.getErrorStream()))
+                            .lines()
+                            .collect(Collectors.joining("\n")));
+            }
+        } catch (InterruptedException e) {
+            throw new IOException(e);
+        }
+
+        return new BufferedReader(new InputStreamReader(process.getInputStream())).lines();
+    }
+
+    private static class Edge {
+        final Branch from;
+        final Branch to;
+
+        Edge(Branch from, Branch to) {
+            this.from = from;
+            this.to = to;
+        }
+
+        @Override
+        public String toString() {
+            return "Edge{" +
+                    "from='" + from + '\'' +
+                    ", to='" + to + '\'' +
+                    '}';
+        }
+    }
+
+    private static List<Branch> tsort(List<Edge> edges) {
+        List<Edge> eCopy = new ArrayList<>(edges);
+        List<Branch> result = new ArrayList<>();
+        while (!eCopy.isEmpty()) {
+            Set<Branch> orphans = eCopy.stream()
+                    .map(e -> e.from)
+                    .filter(f -> eCopy.stream().map(e -> e.to).noneMatch(f::equals))
+                    .collect(Collectors.toSet());
+            if (orphans.isEmpty()) {
+                throw new IllegalStateException("Detected a cycle! " + edges);
+            }
+            orphans.forEach(o -> {
+                result.add(o);
+                eCopy.removeIf(e -> o.equals(e.from));
+            });
+        }
+
+        // add all leaves
+        edges.stream()
+            .map(e -> e.to)
+            .filter(f -> edges.stream().map(e -> e.from).noneMatch(f::equals))
+            .forEach(result::add);
+
+        return result;
+    }
+
+    @Override
+    public String toString() {
+        return "TopoBot@(" + hostedRepo + ")";
+    }
+
+    @Override
+    public List<WorkItem> getPeriodicItems() {
+        return List.of(this);
+    }
+}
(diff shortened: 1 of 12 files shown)


bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalBot.java
```diff
@@ -0,0 +1,244 @@
...
+            } else {
+                log.info("Fast forwarded " + branch + " to " + parent);
+            }
+            log.info("merge with " + parent + " succeeded. The following commits will be pushed:\n"
+                    + log(repo, "origin/" + branch.name(), branch.name()).collect(Collectors.joining("\n", "\n", "\n")));
+            try {
+                repo.push(repo.head(), hostedRepo.getUrl(), branch.name());
+            } catch (IOException e) {
+                log.severe("Pusing failed! Aborting...");
+                repo.abortMerge();
```
JornVernee: Hmm, don't think this `abortMerge` will actually do anything, since the working tree is clean...

edvbld:
Hi Jorn, thanks for the contribution! :clap:

Just a quick suggestion after a brief look - how about moving `Edge` to its own file, `Edge.java` and also move `tsort` into its own class, `TopologicalSort` (in `TopologicalSort.java`)? That way you should have a fairly easy time to add a new test class, `TopologicalSortTests`, with some unit tests for the topological sort

bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalBot.java
```diff
@@ -0,0 +1,244 @@
...
+                repo = Repository.get(dir)
+                        .orElseThrow(() -> new RuntimeException("Repository in " + dir + " has vanished"));
+            }
+
+            repo.fetchAll();
+            var depsFile = repo.root().resolve(depsFileName);
+
+            List<Branch> ordered = orderedBranches(repo, depsFile);
+            log.info("Merge order " + ordered);
+            for (Branch branch : ordered) {
```
edvbld:
```suggestion
            for (var branch : ordered) {
```

bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalBot.java
```diff
@@ -0,0 +1,244 @@
...
+                log.info("Processing branch " + branch + "...");
+                repo.checkout(branch);
+                Set<String> parents = new HashSet<>(
+                        Files.exists(depsFile) ? Files.readAllLines(depsFile) : List.of("master"));
+                List<String> failedMerges = new ArrayList<>();
```
edvbld: Isn't the format of the `deps.txt` file that the branches are separated with space? So maybe this should be:
```java
Set<String> parents = null;
if (Files.exists(depsFIle)) {
    var lines = Files.readAllLines(depsFile).stream().filter(String::isEmpty).collect(Collectors.toList());
    if (lines.size() > 1) {
        throw new IllegalStateException("Multiple non-empty lines in " + depsFile.toString() + ": " + String.join("\n", lines));
    }
    parents = new HashSet<>(Arrays.asList(lines.get(0).split(" ")));
} else {
    parents = Set.of("master");
}

File: bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalBot.java [hunk 4]
```diff
@@ -0,0 +1,244 @@
...
+        return tsort(deps).stream()
+            .filter(branch -> !branch.name().equals("master"))
+            .collect(Collectors.toList());
+    }
+
+    private void mergeIfAhead(Repository repo, Branch branch, String parent) throws IOException {
+        if (log(repo, branch.name(), parent).count() != 0) {
+            var fromHash = repo.resolve(parent).orElseThrow();
+            var isFastForward = repo.isAncestor(repo.head(), fromHash);
+            repo.merge(fromHash);
```
mcimadamore: All the logic in this routine is very dependent on the behavior of the VCS, and on which actual concrete commands do these API method map to, so it's hard to check whether this code is correct. One issue you already pointed oit yourself: with the abortMerge() method. Another issue I see is the behavior with respect to a merge conflict. In the old script we had to tweak the standard behavior a bit to try and auto-resolve conflicts as much as possible. I don't know what the standard merge behavior is in Git.

bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalBot.java
```diff
@@ -0,0 +1,244 @@
...
+        log.info("Ending topobot run");
+    }
+
+    private List<Branch> orderedBranches(Repository repo, Path depsFile) throws IOException {
+        List<Edge> deps = new ArrayList<>();
+        for (Branch branch : branches) {
+            repo.checkout(branch);
+            if (Files.exists(depsFile)) {
+                Files.lines(depsFile)
+                        .forEach(dep -> deps.add(new Edge(new Branch(dep), branch)));
```
edvbld: May I suggest having a function for reading `depsFile` that return the name of the branches in the depsFile as a `List`. Something like:
```java
private List<Branch> dependencies(Path depsFile) {
    if (Files.exists(depsFile)) {
        // parse despFile
    } else {
        return List.of("master");
    }
}

File: bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalBot.java [hunk 6]
```diff
@@ -0,0 +1,244 @@
...
+        public String toString() {
+            return "Edge{" +
+                    "from='" + from + '\'' +
+                    ", to='" + to + '\'' +
+                    '}';
+        }
+    }
+
+    private static List<Branch> tsort(List<Edge> edges) {
+        List<Edge> eCopy = new ArrayList<>(edges);
```
mcimadamore: It seems to me that this methods is walking all the edges and trying to find cycles, and leaves. This probably works, but can be quadratic in cost. There are better alternatives out there - some of which already tried and tested which are part of the javac compiler itself. I suggest maybe directly using this?

http://hg.openjdk.java.net/jdk/jdk/file/c16208de74da/src/jdk.compiler/share/classes/com/sun/tools/javac/util/GraphUtils.java

There, you will also find interfaces for Node and the likes. The topo sort algo used there is routinely used in javac type inference, so it's been rather heavily tested :-)

edvbld: On the other hand I don't expect a branch to have more than lets say five other branches it depends on, so I think a quadratic cost could work here :smiley:

bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalBot.java
edvbld: @mcimadamore you can find all such details in https://github.com/openjdk/skara/blob/master/vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java

bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalBot.java
```diff
@@ -0,0 +1,244 @@
...
+                deps.add(new Edge(new Branch("master"), branch));
+            }
+        }
+        return tsort(deps).stream()
+            .filter(branch -> !branch.name().equals("master"))
+            .collect(Collectors.toList());
+    }
+
+    private void mergeIfAhead(Repository repo, Branch branch, String parent) throws IOException {
+        if (log(repo, branch.name(), parent).count() != 0) {
```
edvbld: Can't you express this with just `if (!repo.contains(branch, repo.resolve(parent).get())`?

bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalBot.java
```diff
@@ -0,0 +1,244 @@
...
+            var fromHash = repo.resolve(parent).orElseThrow();
+            var isFastForward = repo.isAncestor(repo.head(), fromHash);
+            repo.merge(fromHash);
+            if (!isFastForward) {
+                log.info("Merged " + parent + " into " + branch);
+                repo.commit("Automatic merge with " + parent, "topobot", "");
```
edvbld: Please `"duke"` and `"duke@openjdk.org"` as name and email for bot authors

bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalBot.java
JornVernee: Yep, it's pretty quick and dirty implementation of finding any nodes without incoming edges and then adding them to the sorted list and removing from the graph. Finally, we need to once more find all leaves, since those will always have incoming edges.

It doesn't seem like perf would be a problem, and the GraphUtils impl seems a lot more extensive. So, since it works (just done writing tests), maybe stick with Quick & Dirty™ for now?

bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalBot.java
JornVernee: That works as well, was tinkering with calling `git branch --contains` on the command line before. Didn't notice there was API for that :)

bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalBot.java
JornVernee: You're right, thanks for catching that.

bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalBot.java
mcimadamore: I'm fine with quick and dirty

bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalBot.java
mcimadamore: well, yes and no. I mean, I could read all the code of course - but for instance, it seems like when you do "git merge" you typically do a merge and then a commit (unless I'm mistaken). Here I see a call to merge() then commit(). Are we sure we're not committing twice?

bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalBot.java
mcimadamore: If this translates to "git merge --abort" then this is required, I think, as it will wipe out the repo of any by-product of the (failed) merge attempt (e.g. reset all files to original head).

bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalBot.java
edvbld: Yep, see the `--no-commit` flag passed to `git merge`:
https://github.com/openjdk/skara/blob/master/vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java#L873

bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalBot.java
```diff
@@ -0,0 +1,220 @@
...
+            return Stream.of(lines.get(0).split(" ")).map(Branch::new);
+        } else {
+            return Stream.of(new Branch("master"));
+        }
+    }
+
+    private List<Branch> orderedBranches(Repository repo, Path depsFile) throws IOException {
+        List<Edge> deps = new ArrayList<>();
+        for (var branch : branches) {
+            repo.checkout(branch);
```
edvbld: There is no need to check out the branch just to read the file. You can replace this with for example:
```java
var hash = repo.resolve(branch).orElseThrow();
dependencies(repo.lines(hash, depsFile)).forEach(dep -> deps.add(new Edge(dep, branch)));
```
```

File: bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalBot.java [hunk 10]
```diff
@@ -0,0 +1,220 @@
...
+            dependencies(depsFile).forEach(dep -> deps.add(new Edge(dep, branch)));
+        }
+        return TopologicalSort.tsort(deps).stream()
+            .filter(branch -> !branch.name().equals("master"))
```
edvbld: If you want you can use `.filter(branch -> !branch.equals(repo.defaultBranch()))`

bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalBot.java
```diff
@@ -0,0 +1,220 @@
...
+                log.severe("Pusing failed! Aborting...");
+                hardReset(repo, oldHead);
+                throw e;
+            }
+        }
+    }
+
+    private void hardReset(Repository repo, Hash oldHead) throws IOException {
+        var process = new ProcessBuilder()
+            .command("git", "reset", "--hard", oldHead.hex())
```
edvbld: Please add `Repository::reset(boolean hard)` instead

bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalBot.java
JornVernee: There are 2 kinds of merges that can go on. 1.) where the child branch has not diverted, e.g. by making it's own commits, from the parent branch. In that case a 'fast forward' happens, and no merge commit is made. 2.) In the case the child branch has diverted, we make merge commit that shows a merge from the parent has happened.

I don't think there are cases where we can get 2 commits (since `merge` alone never makes a commit [1](https://github.com/openjdk/skara/blob/master/vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java#L873)), but we could for instance disable the fast forward behaviour and then we always make a commit?

bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalBot.java
```diff
@@ -0,0 +1,220 @@
...
+            var isFastForward = repo.isAncestor(oldHead, fromHash);
+            repo.merge(fromHash);
+            if (!isFastForward) {
+                log.info("Merged " + parent + " into " + branch);
+                repo.commit("Automatic merge with " + parent, "duke", "duke@openjdk.org");
+            } else {
+                log.info("Fast forwarded " + branch + " to " + parent);
+            }
+            log.info("merge with " + parent + " succeeded. The following commits will be pushed:\n"
+                    + log(repo, "origin/" + branch.name(), branch.name()).collect(Collectors.joining("\n", "\n", "\n")));
```
edvbld: You could just do `repo.commits("origin/" + branch.name() + ".." + branch.name())` instead of rolling your own

bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalBot.java
```diff
@@ -0,0 +1,220 @@
...
+            try {
+                repo.push(repo.head(), hostedRepo.getUrl(), branch.name());
+            } catch (IOException e) {
+                log.severe("Pusing failed! Aborting...");
+                hardReset(repo, oldHead);
```
edvbld: So this would become `repo.reset(oldHead, true)` (see my comment further down for details)

bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalBot.java
```diff
@@ -0,0 +1,220 @@
...
+        var process = new ProcessBuilder()
+                .command("git", "log", targetRef + ".." + fromRef, "--")
+                .directory(repo.root().toFile())
+                .start();
+        await(process);
+
+        return new BufferedReader(new InputStreamReader(process.getInputStream())).lines();
+    }
+
+    private static void await(Process process) throws IOException {
```
edvbld: I don't think you need this after you have removed `log` and `hardReset`

bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalSort.java
```diff
@@ -0,0 +1,60 @@
...
+import org.openjdk.skara.vcs.Branch;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+class TopologicalSort {
+
+    static List<Branch> tsort(List<Edge> edges) {
```
edvbld: I think naming the method `sort` is enough given that the class is named `TopologicalSort` :smiley_cat:

vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
```diff
@@ -295,6 +295,21 @@ public void clean() throws IOException {
...
+    @Override
+    public void reset(Hash target, boolean hard) throws IOException {
+        var cmd = new ArrayList<String>();
+        cmd.addAll(List.of("git", "reset"));
+        if (hard) {
+           cmd.add("--hard");
+        }
+        cmd.add(target.hex());
+
+        try (var p = capture(cmd.toArray(String[]::new))) {
```
edvbld: There is no need to do `cmd.toArray`, `capture` has an overload for `List<String>`

vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
```diff
@@ -295,6 +295,21 @@ public void clean() throws IOException {
         }
     }
 
+    @Override
+    public void reset(Hash target, boolean hard) throws IOException {
```
edvbld: Did you add a unit test for `reset` in `RepositoryTests`?

JornVernee: Nope, will do

bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalBot.java
```diff
@@ -0,0 +1,185 @@
...
+            var isFastForward = repo.isAncestor(oldHead, fromHash);
+            repo.merge(fromHash);
+            if (!isFastForward) {
+                log.info("Merged " + parent + " into " + branch);
+                repo.commit("Automatic merge with " + parent, "duke", "duke@openjdk.org");
+            } else {
+                log.info("Fast forwarded " + branch + " to " + parent);
+            }
+            log.info("merge with " + parent + " succeeded. The following commits will be pushed:\n"
+                    + repo.commits("origin/" + branch.name() + ".." + branch.name()).stream()
```
edvbld: Please use `asList()` or close the `stream()`, otherwise this will leak process id:s

edvbld: You might also want to break this out perhaps a few more lines :smile_cat:

bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalSort.java
```diff
@@ -0,0 +1,60 @@
...
+
+import org.openjdk.skara.vcs.Branch;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+class TopologicalSort {
+
```
edvbld: Can skip this empty line

bots/topological/src/main/java/org/openjdk/skara/bots/topological/TopologicalSort.java
```diff
@@ -0,0 +1,60 @@
...
+        // add all leaves
+        edges.stream()
+            .map(e -> e.to)
+            .filter(f -> edges.stream().map(e -> e.from).noneMatch(f::equals))
+            .forEach(result::add);
+
+        return result;
+    }
+
```
edvbld: Can skip this one too :)

PR #107, opened 2019-08-29
jcheck "repository" property in [general] section not read from .jcheck/conf

kevinrushforth:
This PR fixes a typo in the `GeneralConfiguration` class of `jcheck` that causes the `repository` property to be set to the same value as `project` rather than being read from the `.jcheck/conf` file.

Progress
--------

Approvers
---------


kevinrushforth:
I might have expected the `/integrate` in the previous comment to cause the bot to wake up and annotate the PR to indicate that it was now ready to be sponsored. Does it not do that? Should it?

rwestberg:
It most certainly should wake up, I've filed https://bugs.openjdk.java.net/browse/SKARA-91 to track the fix.

PR #113, opened 2019-08-30
Coalesce hunks that have overlapping context

JornVernee:
The `webrev` command can generate patch files with adjacent hunks that have overlapping context. `git apply` rejects these patches.

This PR coalesces adjacent hunks with overlapping context, which is also what's done by `git diff`.

Also added a test for `HunkCoalescer` (since it's a pretty complex class I think it deserved it's own tests).

Resulting webrev applies with both git and hg.

Progress
--------

Approvers
---------
 * Erik Helin ([ehelin](@edvbld) - **Reviewer**) **Note!** Review applies to 6f5f10711384ae9a27f7e8b45e46d455c90bb40d

### DIFF
--- a/webrev/src/main/java/org/openjdk/skara/webrev/HunkCoalescer.java
+++ b/webrev/src/main/java/org/openjdk/skara/webrev/HunkCoalescer.java
@@ -148,8 +148,10 @@ class HunkCoalescer {
             var last = hunksInRange.get(hunksInRange.size() - 1);
             var destEnd = last.target().range().end() + numContextLines;
             var sourceEnd = last.source().range().end() + numContextLines;
-            if (sourceEnd >= next.source().range().start() ||
-                  destEnd >= next.target().range().start()) {
+            var nextDestStart = next.target().range().start() - numContextLines;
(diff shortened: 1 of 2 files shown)


JornVernee:
@edvbld I tested with the following script:

```powershell
for ($i = 1; $i -le 20; $i++) {
	git webrev -r "HEAD~$i" -N
	git checkout "HEAD~$i"
	git apply .\webrev\git-jdk2.patch
	git checkout master --force
}
```
This went alright until about `HEAD~16`, which threw a 'corrupt patch' error. Looking into it, the offending line is:

    diff a/test/hotspot/jtreg/ProblemList-aot.txt b/test/hotspot/jtreg/ProblemList-aot.txt

I thought it might be the hyphen in the name, but `git diff` generates the same, and that patch does apply cleanly. Even when copying over the line from the git diff to the patch file generated by `webrev` it still complains. Not sure what's going on...

However, running that script without this PR fails on the first webrev, so I'd say that is a net improvement :), and since the issue seems unrelated, I think we could still move ahead with this PR?

edvbld:
@JornVernee yeah lets integrate this one since it is an improvement at least

PR #122, opened 2019-09-09
Add support for Mercurial

edvbld:
Hi all,

this patch adds skara.py which makes it quite a bit more convenient to use the Skara tools from Mercurial (hg). I also added `--mercurial` to `git-pr` and `git-fork` so they can be used with [hg-git](https://bitbucket.org/durin42/hg-git). There are still a few minor things to work out, for example `hg pr create` does not work yet and I need to do add some help texts, but it is a start.

## Testing
- [x] Manual testing with git token

Thanks,
Erik

Progress
--------

Approvers
---------
 * Robin Westberg ([rwestberg](@rwestberg) - **Reviewer**) **Note!** Review applies to fadd0cd8aab3d351ec697fd83ab0ba547a6738b0

### DIFF
--- a/cli/src/main/java/org/openjdk/skara/cli/GitFork.java
+++ b/cli/src/main/java/org/openjdk/skara/cli/GitFork.java
@@ -55,21 +55,22 @@ public class GitFork {
         }
     }
 
-    private static Repository clone(URI from, Path dest) throws IOException {
+    private static Repository clone(URI from, Path dest, boolean isMercurial) throws IOException {
         try {
             var to = dest == null ? Path.of(from.getPath()).getFileName() : dest;
             if (to.toString().endsWith(".git")) {
                 to = Path.of(to.toString().replace(".git", ""));
             }
 
-            var pb = new ProcessBuilder("git", "clone", from.toString(), to.toString());
+            var vcs = isMercurial ? "hg" : "git";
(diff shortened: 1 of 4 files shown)


cli/src/main/java/org/openjdk/skara/cli/GitFork.java
```diff
@@ -140,27 +146,39 @@ public static void main(String[] args) throws IOException {
...
         var host = Host.from(uri, new PersonalAccessToken(credentials.username(), credentials.password()));
-        path = credentials.path();
         if (path.endsWith(".git")) {
             path = path.substring(0, path.length() - 4);
         }
+        if (path.startsWith("/")) {
+            path = path.substring(1);
+        }
 
+        System.out.println("path: " + path);
```
rwestberg: Debugging printout?

skara.py
```diff
@@ -0,0 +1,155 @@
...
+# Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+# or visit www.oracle.com if you need additional information or have any
+# questions.
+
+import mercurial
+import os.path
+import subprocess
+import sys
+import shutil
+
```
rwestberg: Add a testedwith line?

PR #148, opened 2019-09-19
Add support for adding an issue when formatting the commit message

rwestberg:
Hi all,

Please review this change that allows setting the issue of a commit through the PR title.

Best regards,
Robin

Progress
--------

Approvers
---------

### DIFF
--- a/host/src/main/java/org/openjdk/skara/host/github/GitHubPullRequest.java
+++ b/host/src/main/java/org/openjdk/skara/host/github/GitHubPullRequest.java
@@ -213,6 +213,11 @@ public class GitHubPullRequest implements PullRequest {
     }
 
     @Override
+    public void setTitle(String title) {
+        throw new RuntimeException("not implemented yet");
+    }
+
+    @Override
     public String getBody() {
         var body = json.get("body").asString();
         if (body == null) {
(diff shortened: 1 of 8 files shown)


host/src/main/java/org/openjdk/skara/host/github/GitHubPullRequest.java
```diff
@@ -212,6 +212,11 @@ public String getTitle() {
         return json.get("title").asString();
     }
 
+    @Override
+    public void setTitle(String title) {
+        throw new RuntimeException("not implemented yet");
```
edvbld: Won't this be an issue if we run the unit tests as integration tests against real hosts?

host/src/main/java/org/openjdk/skara/host/gitlab/GitLabMergeRequest.java
```diff
@@ -254,6 +254,11 @@ public String getTitle() {
         return json.get("title").asString();
     }
 
+    @Override
+    public void setTitle(String title) {
+        throw new RuntimeException("not implemented yet");
```
edvbld: Same issue here?

host/src/main/java/org/openjdk/skara/host/github/GitHubPullRequest.java
rwestberg: You are quite right, I was planning to save it for another task, as there are already existing issues with the real host tests..

edvbld: Ah ok, lets tackle it later then :+1:

PR #153, opened 2019-09-20
Add git-sync command

edvbld:
Hi all,

this commits adds the new `git sync` command which is useful for syncing a personal fork with its upstream repository. `git sync` supports both a remote and URL as a argument, for example:

```bash
$ git sync upstream
Syncing upstream/master with origin/master... done

$ git sync https://github.com/openjdk/skara
Syncing https://github.com/openjdk/skara/master with origin/master... done
```

The remote can also be configured in `.git/config` (or `~/.gitconfig`) as in:

```
[sync]
       remote = upstream
```

I also added two options, `--pull` and `--fetch`, that will pull and/or fetch the current branch. This is a shorthand for `git sync && git pull`. Since `git fork` sets the `sync.remote` in `.git/config` (local for the repository) a user now only has to write to fork an upstream repository, sync the branches and update their local copy:

```bash
$ git fork https://github.com/openjdk/skara
$ cd skara
$ git sync --pull
```

The syncing leaves no trace in the local repository, I'm only using `FETCH_HEAD` in the refspec when pushing to the personal fork.

## Testing
- [x] `sh gradlew reproduce` passes on Linux x86_64
- [x] Added two new unit tests
- [x] Manual testing of `git sync`

Thanks,
Erik

Progress
--------

Approvers
---------
 * Robin Westberg ([rwestberg](@rwestberg) - **Reviewer**) **Note!** Review applies to 62268c27972ce202e4cb0b9a92ebc5078a61440b
 * Jorn Vernee ([jvernee](@JornVernee) - Committer)

### DIFF
--- a/cli/src/main/java/org/openjdk/skara/cli/GitSync.java
+++ b/cli/src/main/java/org/openjdk/skara/cli/GitSync.java
@@ -1,0 +1,118 @@
+/*
+ * Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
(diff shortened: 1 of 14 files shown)


cli/src/main/java/org/openjdk/skara/cli/GitSync.java
```diff
@@ -0,0 +1,164 @@
...
+            if (lines.size() == 1 && remotes.contains(lines.get(0))) {
+                upstream = lines.get(0);
+            } else {
+                die("No remote provided to sync with");
+            }
+        }
+
+        var upstreamPullPath = remotes.contains(upstream) ?
+            Remote.toURI(repo.pullPath(upstream)) : URI.create(upstream);
+        var origin = "origin";
```
JornVernee: I think the fork remote name should be a CLI option as well, maybe in the gitconfig as well (e.g. I have named it `fork` for some of my local repos)

cli/src/main/java/org/openjdk/skara/cli/GitSync.java
```diff
@@ -0,0 +1,164 @@
...
+            }
+        }
+
+        for (var branch : repo.remoteBranches(upstream)) {
+            var name = branch.name();
+            if (!branches.isEmpty() && !branches.contains(name)) {
+                System.out.println("Skipping branch " + name);
+                continue;
+            }
+            System.out.print("Syncing " + upstream + "/" + name + " to " + origin + "/" + name + "... ");
```
rwestberg: Do you need to flush the PrintStream here?

edvbld: Probably, good idea, let me fix!

cli/src/main/java/org/openjdk/skara/cli/GitSync.java
JornVernee: Sounds good!

PR #159, opened 2019-09-24
Add "binary" jcheck check

edvbld:
Hi all,

this patch adds a new jcheck check called "binary" that checks for binary files.

## Testing
- `sh gradlew test` passes on Linux x86_64
- Added two new unit tests

Thanks,
Erik

Progress
--------

Approvers
---------

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
@@ -207,4 +207,9 @@ class PullRequestCheckIssueVisitor implements IssueVisitor {
     public void visit(BlacklistIssue issue) {
         log.fine("ignored: blacklisted commit");
     }
+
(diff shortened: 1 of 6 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
```diff
@@ -207,4 +207,9 @@ public void visit(ExecutableIssue issue) {
     public void visit(BlacklistIssue issue) {
         log.fine("ignored: blacklisted commit");
     }
+
+    @Override
+    public void visit(BinaryIssue issue) {
+        log.fine("ignored: binary file");
```
rwestberg: Should we do something more here? Perhaps something for a follow up..

edvbld: My plan was to leave that to you :wink:

PR #163, opened 2019-09-25
80: Allow valid bug ID format to be configurable

kevinrushforth:
JBS issue: https://bugs.openjdk.java.net/browse/SKARA-80

This provides a configuration option to define the pattern for a valid issue ID as follows. This starts with the idea @edvbld mentioned in PR #99, although I felt it was cleaner to leave the default issue format as is, and use the more restrictive pattern for the "legacy" jdk conf file.

The syntax of the new pattern is:

```
[checks "issues"]
pattern=<PATTERN_STRING>
```

The default pattern is:

```
[checks "issues"]
pattern=^(([A-Z][A-Z0-9]+-)?[0-9]+): (\S.*)$
```

For legacy JDK `.jcheck/conf` files, the pattern is as follows to match the issue ID rules in `jcheck.py`:

```
[checks "issues"]
pattern=^([124-8][0-9]{6}): (\S.*)$
```

Progress
--------

Approvers
---------
 * Robin Westberg ([rwestberg](@rwestberg) - **Reviewer**) **Note!** Review applies to a12ec8ee97b871f7ef91a0a4412d7c5be8da9bdc

### DIFF
--- a/jcheck/src/main/java/org/openjdk/skara/jcheck/IssuesCheck.java
+++ b/jcheck/src/main/java/org/openjdk/skara/jcheck/IssuesCheck.java
@@ -45,7 +45,7 @@ public class IssuesCheck extends CommitCheck {
 
         var metadata = CommitIssue.metadata(commit, message, conf, this);
         if (commit.message().isEmpty() || message.issues().isEmpty()) {
-            log.finer("isuse: no reference to a JBS issue");
+            log.finer("issue: no reference to a JBS issue");
             return iterator(new IssuesIssue(metadata));
         }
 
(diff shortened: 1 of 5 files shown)


jcheck/src/main/java/org/openjdk/skara/jcheck/IssuesCheck.java
```diff
@@ -42,12 +43,19 @@
             return iterator();
         }
 
+        var metadata = CommitIssue.metadata(commit, message, conf, this);
         if (commit.message().isEmpty() || message.issues().isEmpty()) {
-            var metadata = CommitIssue.metadata(commit, message, conf, this);
             log.finer("isuse: no reference to a JBS issue");
```
rwestberg:
```suggestion
            log.finer("issue: no reference to a JBS issue");
```

kevinrushforth:
@rwestberg I fixed the typo you pointed out.

PR #170, opened 2019-09-30
WIP: 108: Command line tools fon't run with Cygwin git

kevinrushforth:
https://bugs.openjdk.java.net/browse/SKARA-108

This is a very preliminary "proof of concept" patch that allows `git jcheck` to work with git from Cygwin. It is incomplete, and is missing the necessary checks to allow it to continue working with native Windows git.

@edvbld @rwestberg - I submitted it this early to get your high-level feedback on whether this support is something that you have thought about already; if not, I'd like your feedback on the approach.

The idea would be to check whether we are are using a Cygwin version of git (this should be trivial), and if so, do the conversion between the `Path` strings and the input/output to the `git` command in all of the necessary places in `GitRepository` (that should be the only class that needs to be modified).

---------
### Progress

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/170/head:pull/170`
`$ git checkout pull/170`

### DIFF
--- a/skara.sh
+++ b/skara.sh
@@ -24,7 +24,14 @@
 DIR=$(dirname "${0}")
 OS=$(uname)
 
-if [ "${OS}" = "Linux" -o "${OS}" = "Darwin" ]; then
+case "${OS}" in
+    Linux) USEBAT=0 ;;
+    Darwin) USEBAT=0 ;;
+    CYGWIN*) USEBAT=0 ;;
+    *) USEBAT=1 ;;
+esac
+
+if [ ${USEBAT} -eq 0 ]; then
     if [ ! -x "${DIR}/bin/bin/git-skara" ]; then
         echo "Compiling ..."
         (cd "${DIR}" && sh gradlew)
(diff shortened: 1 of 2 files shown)


rwestberg:
Looks like a reasonable approach to me, perhaps our most prominent Windows user @jornvernee has considered something in this area.. I usually avoid installing cygwin git to avoid issues like this.

JornVernee:
@rwestberg I'm happily using Windows native git :)

edvbld:
Hi Kevin,

thanks for looking into this and sorry for not responding earlier. I would like to understand your use case a bit more - why are you using Cygwin git over [Git for Windows](https://gitforwindows.org/)? I understand that you need to use Cygwin, but why not use the Windows native git binary in Cygwin (instead of Cygwin's version of git)?

Looking at your changes I can immediately see that adding support for the Cygwin version of git will have far-reaching consequences for almost all our code, so I would really like to understand the advantages of using Cygwin git before pursuing this further.

Thanks,
Erik

rwestberg:
Please disregard this comment, investigating a failed email delivery.

kevinrushforth:
I'm withdrawing this PR. I still think having support for Cygwin git would be a good thing. I tried native Windows git a few months back and had a variety of usage issues, so I'm sticking with Cygwin git. I can live without formal support for the Skara command line tools on Cygwin git (my temporary patch is enough to allow `git jcheck` to work, so this isn't blocking me).

PR #172, opened 2019-10-01
Add Makefile wrapper

edvbld:
Hi all,

this patch adds a Makefile "wrapper" around `gradlew`. This allows the Skara CLI tooling to be installed using the (for Linux/macOS users) familiar commands:

```bash
$ make
$ make install
```

The user can also give a prefix to `make install`, as in `make prefix=/usr install`. The Makefile variables have the same names as for git upstream. One benefit of installing the Skara CLI tooling this way is that the man pages are available via `--help` to various Skara CLI commands.

Please note that this is in _no_ way meant to replace the existing Gradle based build system, this is only a thin wrapper for convenience.

Thanks,
Erik

Progress
--------

Approvers
---------
 * Robin Westberg ([rwestberg](@rwestberg) - **Reviewer**) **Note!** Review applies to c34d755acee1ded4c011dd858ca519c4f5ddbbcc

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,56 @@
+# Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
+# DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+#
+# This code is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 only, as
+# published by the Free Software Foundation.
+#
+# This code is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -0,0 +1,56 @@
...
+	@sh gradlew
+
+check:
+	@sh gradlew test
+
+test:
+	@sh gradlew test
+
+install: all $(LAUNCHERS) $(MANPAGES) $(sharedir)/skara
+
```
rwestberg: Perhaps echo something like "Installed to $(prefix)" here?

edvbld: Good idea, let me fix. Erik Joelsson also suggested offline that I should probably provide an `uninstall` target as well, I will add that too.

rwestberg: Looks good!

PR #173, opened 2019-10-01
Add Host::isMemberOf

edvbld:
Hi all,

this patch adds a new method `Host::isMemberOf` that can be used to query a host (e.g. GitHub or Gitlab) if a specific user is a member of a particular group. I couldn't come up with a positive test for the method, since it is really hard to rely on a test user being a member of a particular group, but I could at least write a negative test.

## Testing

Thanks,
Erik

Progress
--------

Approvers
---------

### DIFF
--- a/host/src/main/java/org/openjdk/skara/host/Host.java
+++ b/host/src/main/java/org/openjdk/skara/host/Host.java
@@ -31,7 +31,7 @@ public interface Host {
     HostUserDetails getUserDetails(String username);
     HostUserDetails getCurrentUserDetails();
     boolean supportsReviewBody();
-    boolean isMemberOf(long groupId, HostUserDetails user);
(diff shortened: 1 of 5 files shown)


host/src/main/java/org/openjdk/skara/host/Host.java
```diff
@@ -31,6 +31,7 @@
     HostUserDetails getUserDetails(String username);
     HostUserDetails getCurrentUserDetails();
     boolean supportsReviewBody();
+    boolean isMemberOf(long groupId, HostUserDetails user);
```
rwestberg: You may want to make the groupId parameter a String instead, there are hosts out there that we may want to support that don't use numerical ids.

edvbld: Thanks, updated!

PR #178, opened 2019-10-03
Use patterns for branches to notify instead of a list

rwestberg:
Hi all,

Please review this change that lets the notifier use a pattern to determine which branches to monitor, instead of a fixed list. Also fetch everything from the remote in order to perform the pattern matching locally.

Best regards,
Robin

## Progress

## Approvers
 * Erik Helin ([ehelin](@edvbld) - **Reviewer**) **Note!** Review applies to 7a6183d0f3a2371ea8a641d9243c7ee252b3e588

### DIFF
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/JNotifyBot.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/JNotifyBot.java
@@ -174,8 +174,7 @@ class JNotifyBot implements Bot, WorkItem {
             handleTags(localRepo, history);
 
             for (var ref : localRepo.remoteBranches("origin")) {
-                var branchMatcher = branches.matcher(ref.name());
-                if (branchMatcher.matches()) {
+                if (branches.matcher(ref.name()).matches()) {
                     var branch = new Branch(ref.name());
                     handleBranch(localRepo, history, branch, ref.hash());
                 }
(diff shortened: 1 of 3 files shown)


bots/notify/src/main/java/org/openjdk/skara/bots/notify/JNotifyBot.java
```diff
@@ -151,20 +150,35 @@ private void handleTags(Repository localRepo, UpdateHistory history) throws IOEx
...
+            var localRepo = fetchAll(path, repository.getUrl());
             var history = UpdateHistory.create(tagStorageBuilder, historyPath.resolve("tags"), branchStorageBuilder, historyPath.resolve("branches"));
             handleTags(localRepo, history);
 
-            for (var branch : branches) {
-                var hash = localRepo.fetch(repository.getUrl(), branch.name());
-                handleBranch(localRepo, history, branch, hash);
+            for (var ref : localRepo.remoteBranches("origin")) {
+                var branchMatcher = branches.matcher(ref.name());
+                if (branchMatcher.matches()) {
```
edvbld: `if (branches.matcher(ref.name()).matches())` ?

rwestberg: Sure, why not!

PR #180, opened 2019-10-03
Make it possible to cancel a Check

edvbld:
Hi all,

this patch became a bit larger than I initially thought. I added a new `CheckState` - `CANCELLED` and a new method to `CheckBuider` - `cancel()`. However to fully support this I needed to quite a bit of refactoring in both `GitHubPullRequest` and `GitLabMergeRequest` so that `createCheck` allows checks with all kinds of states to be created.

## Testing
- [x] Added one new unit test
- [x] `sh gradlew test` with credentials mostly passes against both GitHub and GitLab (there are a few known failures)

Thanks,
Erik

## Progress

## Approvers
 * Robin Westberg ([rwestberg](@rwestberg) - **Reviewer**) **Note!** Review applies to ed66f45c4d5671275e0ed7a5f2e591ae958dece7

### DIFF
--- a/host/src/main/java/org/openjdk/skara/host/gitlab/GitLabMergeRequest.java
+++ b/host/src/main/java/org/openjdk/skara/host/gitlab/GitLabMergeRequest.java
@@ -440,7 +440,7 @@ public class GitLabMergeRequest implements PullRequest {
                 body = ":warning: The merge request check **" + check.name() + "** identified the following issues:";
                 break;
             case CANCELLED:
-                body = ":x: The merge request check **" + check.name() + "** has been cancelled";
+                body = ":x: The merge request check **" + check.name() + "** has been cancelled.";
                 break;
             default:
                 throw new RuntimeException("Unknown check status");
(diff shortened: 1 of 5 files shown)


host/src/main/java/org/openjdk/skara/host/gitlab/GitLabMergeRequest.java
```diff
@@ -408,71 +426,87 @@ private String linkToDiff(String path, Hash hash, int line) {
...
             case SUCCESS:
-                status = "SUCCESS";
+                body = ":tada: The merge request check **" + check.name() + "** completed successfully!";
                 break;
             case FAILURE:
-                status = "FAILURE";
+                body = ":warning: The merge request check **" + check.name() + "** identified the following issues:";
+                break;
+            case CANCELLED:
+                body = ":x: The merge request check **" + check.name() + "** has been cancelled";
```
rwestberg:
```suggestion
                body = ":x: The merge request check **" + check.name() + "** has been cancelled.";
```

PR #231, opened 2019-10-31
Add support for building on AArch64

nick-arm:
I'm working with OpenJDK on Arm machines so I want to be able to run
`git webrev' etc. there. At the moment we can't build the Skara tools on
AArch64 due to an implicit assumption that Linux means X86. I tried to
refactor the build scripts support multiple architectures on Linux. Only
X86 and AArch64 for now but should be trivial to add others.

Using the AdoptOpenJDK AArch64 binary build as there are no prebuilt
binaries for AArch64 on java.net.

Currently there are several Gradle sub-projects that use the image
plugin to build Java images, and each specifies the URL and SHA265 of
the binary JDK to use. Changed the top-level Gradle script to parse
deps.env and select the JDK source based on the OS and architecture
system properties. This gets rid of the duplicated URLs but now each
image { .. } block has a redundant section like the following for each
OS:

        jdk {
            url = project.jdkUrl
            sha256 = project.jdkSha256
        }

So maybe the plugin should be tweaked to move this up into the image {}
block itself?

## Progress

### DIFF
--- a/cli/build.gradle
+++ b/cli/build.gradle
@@ -76,8 +76,8 @@ images {
         launchers = ext.launchers
         bundles = ['zip', 'tar.gz']
         jdk {
-            url = 'https://download.java.net/java/GA/jdk12/GPL/openjdk-12_windows-x64_bin.zip'
-            sha256 = '35a8d018f420fb05fe7c2aa9933122896ca50bd23dbd373e90d8e2f3897c4e92'
+            url = project.jdkUrl
+            sha256 = project.jdkSha256
         }
     }
 
@@ -87,8 +87,8 @@ images {
         man = 'cli/resources/man'
         bundles = ['zip', 'tar.gz']
         jdk {
-            url = 'https://download.java.net/java/GA/jdk12/GPL/openjdk-12_linux-x64_bin.tar.gz'
-            sha256 = 'b43bc15f4934f6d321170419f2c24451486bc848a2179af5e49d10721438dd56'
+            url = project.jdkUrl
+            sha256 = project.jdkSha256
         }
     }
 
@@ -98,8 +98,8 @@ images {
         man = 'cli/resources/man'
         bundles = ['zip', 'tar.gz']
         jdk {
-            url = 'https://download.java.net/java/GA/jdk12/GPL/openjdk-12_osx-x64_bin.tar.gz'
-            sha256 = '52164a04db4d3fdfe128cfc7b868bc4dae52d969f03d53ae9d4239fe783e1a3a'
+            url = project.jdkUrl
+            sha256 = project.jdkSha256
         }
     }
 }
--- a/gradlew
+++ b/gradlew
@@ -82,12 +82,25 @@ extract_zip() {
 
 DIR=$(dirname $0)
 OS=$(uname)
+ARCH=$(uname -m)
 
 . $(dirname "${0}")/deps.env
 case "${OS}" in
     Linux )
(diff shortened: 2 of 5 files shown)


cli/build.gradle
```diff
@@ -98,8 +98,8 @@ images {
         man = 'cli/resources/man'
         bundles = ['zip', 'tar.gz']
         jdk {
-            url = 'https://download.java.net/java/GA/jdk12/GPL/openjdk-12_osx-x64_bin.tar.gz'
-            sha256 = '52164a04db4d3fdfe128cfc7b868bc4dae52d969f03d53ae9d4239fe783e1a3a'
+            url = project.jdkUrl
+            sha256 = project.jdkSha256
```
edvbld: This is unfortunately not how the `jdk` property of the `images` target is meant to be used.

The JDK fully supports cross-platform linking, something we make use of in Skara. This means if you build the `:cli:images` target on a Linux x64 machine then you will actually get fully working Skara distributions for Windows, macOS *and* Linux (all x64), since we are downloading the JDKs for macOS, Windows and Linux. With this patch we only get Linux x64 distributions.

Fortunately the JDK not only supports cross-jlink across operating system, it also supports cross-linking across CPU architectures. So I would prefer to be able to write:

```
linux_x64 {
    modules = ext.modules
    launchers = ext.launchers
    man = 'cli/resources/man'
    bundles = ['zip', 'tar.gz']
    jdk {
        url = 'https://download.java.net/java/GA/jdk12/GPL/openjdk-12_linux-x64_bin.tar.gz'
        sha256 = 'b43bc15f4934f6d321170419f2c24451486bc848a2179af5e49d10721438dd56'
    }
}

linux_aarch64 {
    modules = ext.modules
    launchers = ext.launchers
    man = 'cli/resources/man'
    bundles = ['zip', 'tar.gz']
    jdk {
        url = 'https://github.com/AdoptOpenJDK/openjdk12-binaries/releases/download/jdk-12.0.2%2B10/OpenJDK12U-jdk_aarch64_linux_hotspot_12.0.2_10.tar.gz'
        sha256 = ''855f046afc5a5230ad6da45a5c811194267acd1748f16b648bfe5710703fe8c'
    }
}
```

This way we will always produce Linux AArch64 images on Linux x64 hosts. To get this to work you need to do some minor changes to `ImagesPlugin.java` (and you might have to do some adjustments to `LinkTask.java`). Do you want to take a stab at that? Or do you want me to try and hack something together?

gradlew
```diff
@@ -82,12 +82,25 @@ extract_zip() {
...
+                ;;
+            aarch64 )
+                JDK_URL="${JDK_LINUX_AARCH64_URL}"
+                JDK_SHA256="${JDK_LINUX_AARCH64_SHA256}"
+                ;;
+            *)
+                echo "error: unsupported Linux architecture ${ARCH}"
+                exit 1
+                ;;
+        esac
```
edvbld: This part looks good :+1:

cli/build.gradle
nick-arm: Ah ok, I see that now. I'm happy to have a go at redoing it.

edvbld:
@nick-arm I thought some more about the build today and ended up implementing something a bit more scalable and future-proof than just adding ARM support on Linux. Please have a look PR #237 and see what you think.

If #237 is fine with you, can I go ahead and close this pull request?

Thanks,
Erik

nick-arm:
@edvbld yes I just tried #237 and it works fine modulo the few small comments I left. Thanks!

PR #234, opened 2019-11-05
Support the "neutral" check conclusion

rwestberg:
Hi all,

Please review this minor change that adds support for the "neutral" check conclusion that can be returned by the GitHub check API.

Best regards,
Robin

## Progress

## Approvers
 * Erik Helin ([ehelin](@edvbld) - **Reviewer**) **Note!** Review applies to fe87b6319a4229cba2fb63c2b21e093b730e53b9


forge/src/main/java/org/openjdk/skara/forge/github/GitHubPullRequest.java
```diff
@@ -317,6 +317,7 @@ public ZonedDateTime updatedAt() {
                                         checkBuilder.complete(true, completedAt);
                                         break;
                                     case "failure":
```
edvbld:
```suggestion
                                    case "failure":
                                        // fallthrough
```

PR #237, opened 2019-11-05
Add support for offline and local builds

edvbld:
Hi all,

this patch adds much improved support for both _offline_ and _local_ builds. We are currently downloading both JDK and Gradle distributions for the most common platforms (GNU/Linux x64, macOS x64 and Windows x64), but this scheme does not scale to multiple other operating systems and/or CPU architectures. I have therefore extended the `images` plugin to support a `local` image which uses the JDK that is currently executing Gradle for producing the jlinked image.

I also added support for fully offline builds for those want to supply both a JDK and Gradle themselves and ensure that the build does not download anything.

Thanks,
Erik

## Testing
- Tested multiple variants of builds, including local, offline and default builds

## Progress

## Approvers
 * Robin Westberg ([rwestberg](@rwestberg) - **Reviewer**) **Note!** Review applies to 66a6b56bbab2b8bd7d79f4c31e39c35a94850824

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,14 +39,14 @@ external Git source code hosting providers are available:
 ## Building
 
 [JDK 12](http://jdk.java.net/12/) or later and [Gradle](https://gradle.org/)
-5.2.1 or later is required for building. To build the project on macOS or
-GNU/Linux, just run the following command from the source tree root:
+5.6.2 or later is required for building. To build the project on macOS or
+GNU/Linux x64, just run the following command from the source tree root:
 
 ```bash
 $ sh gradlew
 ```
 
-To build the project on Windows, run the following command from the source tree root:
+To build the project on Windows x64, run the following command from the source tree root:
 
 ```bat
 > gradlew
@@ -54,6 +54,42 @@ To build the project on Windows, run the following command from the source tree 
 
 The extracted jlinked image will end up in the `build` directory in the source
 tree root.
+
+### Other operating systems and CPU architectures
+
+If you want to build on an operating system other than GNU/Linux, macOS or
+Windows _or_ if you want to build on a CPU architecture other than x64, then
+ensure that you have JDK 12 or later installed locally. You can then run the
+following command from the source tree root:
+
+```bash
+$ sh gradlew
+```
+
+The extracted jlinked image will end up in the `build` directory in the source
+tree root.
+
+### Offline builds
+
+If you don't want the build to automatically download any depenendcies, then
+you must ensure that you have installed the following software locally:
+
+- JDK 12 or later
+- Gradle 5.6.2 or later
+
+To create a build then run the command
+
+```bash
+$ gradle offline
+```
+
+_Please note_ that the above command does _not_ make use of `gradlew` to avoid
+downloading Gradle.
+
+The extracted jlinked image will end up in the `build` directory in the source
+tree root.
+
+### Cross-linking
 
 It is also supported to cross-jlink jimages to GNU/Linux, macOS and/or Windows from
 any of the aforementioned operating systems. To build all applicable jimages
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ tree root.
 
 ### Offline builds
 
-If you don't want the build to automatically download any depenendcies, then
+If you don't want the build to automatically download any dependencies, then
 you must ensure that you have installed the following software locally:
 
 - JDK 12 or later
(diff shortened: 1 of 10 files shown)


edvbld:
@nick-arm could you please try this PR out and see if it works for you?

edvbld:
@JornVernee could you give this a spin on Windows and see if it works?

README.md
```diff
@@ -55,6 +55,42 @@ To build the project on Windows, run the following command from the source tree
...
+```bash
+$ sh gradlew
+```
+
+The extracted jlinked image will end up in the `build` directory in the source
+tree root.
+
+### Offline builds
+
+If you don't want the build to automatically download any depenendcies, then
```
rwestberg:
```suggestion
If you don't want the build to automatically download any dependencies, then
```

README.md
```diff
@@ -55,6 +55,42 @@ To build the project on Windows, run the following command from the source tree
...
+you must ensure that you have installed the following software locally:
+
+- JDK 12 or later
+- Gradle 5.6.2 or later
+
+To create a build then run the command
```
rwestberg:
```suggestion
To create a build then run the command:
```

JornVernee:
Works for me with gradle 5.6.4 and JDK 12b33

buildSrc/images/src/main/java/org/openjdk/skara/gradle/images/ImagesPlugin.java
```diff
@@ -29,9 +29,49 @@
...
+        if (p.startsWith("sunos")) {
+            return "solaris";
+        }
+
+        throw new RuntimeException("Unknown operating system: " + System.getProperty("os.name"));
+    }
+
+    private static String getCPU() {
+        var p = System.getProperty("os.arch").toLowerCase();
+        if (p.startsWith("amd64")) {
```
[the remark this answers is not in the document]
edvbld: Thanks, fixed! I also added a check for `"x64"`.

build.gradle
```diff
@@ -104,19 +104,79 @@ reproduce {
...
+        return 'windows'
+    }
+    if (os.startsWith('sunos')) {
+        return 'solaris'
+    }
+    throw new GradleException("Unexpected operating system: " + os)
+}
+
+def getCPU() {
+    def cpu = System.getProperty('os.arch').toLowerCase()
```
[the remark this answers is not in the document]
edvbld: Fixed, also added a check for `"x64"`.

build.gradle
```diff
@@ -104,19 +104,79 @@ reproduce {
...
+    if (cpu.startsWith('sparc')) {
+        return 'sparc'
+    }
+    if (cpu.startsWith('ppc')) {
+        return 'ppc'
+    }
+    if (cpu.startsWith('arm')) {
+        return 'arm'
+    }
+    throw new GradleException("Unexpected operating system: " + cpu)
```
nick-arm: Should be "Unexpected CPU"?

build.gradle
```diff
@@ -104,19 +104,79 @@ reproduce {
...
+    }
+    if (cpu.startsWith('sparc')) {
+        return 'sparc'
+    }
+    if (cpu.startsWith('ppc')) {
+        return 'ppc'
+    }
+    if (cpu.startsWith('arm')) {
+        return 'arm'
+    }
```
nick-arm: 64-bit and 32-bit Arm have different `os.arch` values. So we need to add:

```groovy
if (cpu.startsWith('aarch64')) {
    return 'aarch64'
}
```

buildSrc/images/src/main/java/org/openjdk/skara/gradle/images/ImagesPlugin.java
```diff
@@ -29,9 +29,49 @@
...
+        }
+        if (p.startsWith("sparc")) {
+            return "sparc";
+        }
+        if (p.startsWith("ppc")) {
+            return "ppc";
+        }
+        if (p.startsWith("arm")) {
+            return "arm";
+        }
```
nick-arm: Also need to check for `aarch64` here:

```java
if (p.startsWith("aarch64")) {
    return "aarch64";
}
```

build.gradle
edvbld: Ah, right, thanks. Fixed!

buildSrc/images/src/main/java/org/openjdk/skara/gradle/images/ImagesPlugin.java
edvbld: Thanks, fixed this one as well!

build.gradle
edvbld: Yep, fixed, thanks!

PR #243, opened 2019-11-07
Use tag annotation if present when sending notifications

rwestberg:
Hi all,

Please review this change that updates the notifier to include the tag annotation if it is present when creating a new tag notification email. Also ensure that notifications are sent even if a tag does not conform to the OpenJDK tag parser format.

Best regards,
Robin

## Progress

## Approvers
 * Erik Helin ([ehelin](@edvbld) - **Reviewer**) **Note!** Review applies to fe09eede152c673f917315708b385ecb873e259b

### DIFF
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/JNotifyBot.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/JNotifyBot.java
@@ -170,29 +170,27 @@ class JNotifyBot implements Bot, WorkItem {
                                 .filter(Optional::isPresent)
                                 .map(Optional::get)
                                 .sorted(Comparator.comparingInt(OpenJDKTag::buildNum))
-                                .map(OpenJDKTag::tag);
-        var newNonJdkTags = newTags.stream()
-                                   .filter(tag -> OpenJDKTag.create(tag).isEmpty());
+                                .collect(Collectors.toList());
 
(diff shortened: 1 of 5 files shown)


bots/notify/src/main/java/org/openjdk/skara/bots/notify/JNotifyBot.java
```diff
@@ -170,29 +170,27 @@ private void handleTags(Repository localRepo, UpdateHistory history) throws IOEx
                                 .filter(Optional::isPresent)
                                 .map(Optional::get)
                                 .sorted(Comparator.comparingInt(OpenJDKTag::buildNum))
-                                .map(OpenJDKTag::tag);
+                                .collect(Collectors.toList());
+
         var newNonJdkTags = newTags.stream()
-                                   .filter(tag -> OpenJDKTag.create(tag).isEmpty());
+                                   .filter(tag -> OpenJDKTag.create(tag).isEmpty())
+                                   .collect(Collectors.toList());
```
edvbld: Please move this below the for loop so it is next to the for loop of `newNonJdkTags`

rwestberg: Sure!

kevinrushforth:
I look forwarded to testing this once you've updated the bot with this change.

PR #257, opened 2019-11-18
Rename encoding/src/main/java/org/openjdk/skara/base85/Base85.java

RealCLanger:
## Progress

## Approvers


RealCLanger:
/summary Move Base85.java to the right location (encoding/src/main/java/org/openjdk/skara/encoding)

PR #258, opened 2019-11-19
Query GitHub for current user when using PAT

edvbld:
Hi all,

please review this patch that queries GitHub for the current user when using a PAT as credentials with `GitHubHost.java`. The reason for this is that some credentials manager, in particular Git Credential Manager on Windows, always seems to return "PersonalAccessToken" as the username when calling `git credential fill`. The fix is simple, just query the [/user](https://developer.github.com/v3/users/#get-the-authenticated-user) endoint instead of [/users](https://developer.github.com/v3/users/#get-a-single-user).

Thanks,
Erik

## Testing
- Manual testing with `git pr` and PAT on Linux x64 _with_ a bogus username in credential store.

## Progress

## Approvers


edvbld:
Thanks @RealCLanger for helping out with debugging this!

PR #259, opened 2019-11-19
Fix parsing of editor commandline in git-pr

RealCLanger:
For git editor configurations that aren't a simple call to an editor but have more parameters, the starting of the editor in git pr does not work.
At least, I observed this in my windows environment, where I have the following git config entry:
core.editor="C:\\Program Files\\Notepad++\\notepad++.exe" -multiInst -notabbar -nosession -noPlugin

## Progress

## Approvers

### DIFF
--- a/cli/src/main/java/org/openjdk/skara/cli/GitPr.java
+++ b/cli/src/main/java/org/openjdk/skara/cli/GitPr.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2020, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -40,12 +40,11 @@ import java.nio.file.*;
 import java.util.*;
 import java.util.regex.Pattern;
 import java.util.concurrent.TimeUnit;
-import java.util.function.Supplier;
 import java.util.logging.Level;
+import java.util.regex.Matcher;
 import java.util.stream.Collectors;
 
 public class GitPr {
-    private static final StandardOpenOption APPEND = StandardOpenOption.APPEND;
     private static final Pattern ISSUE_ID_PATTERN = Pattern.compile("([A-Za-z][A-Za-z0-9]+)?-([0-9]+)");
     private static final Pattern ISSUE_MARKDOWN_PATTERN =
         Pattern.compile("^\\[([A-Z]+-[0-9]+)\\]\\(https:\\/\\/bugs.openjdk.java.net\\/browse\\/[A-Z]+-[0-9]+\\): .*$");
@@ -53,13 +52,6 @@ public class GitPr {
     private static void exit(String fmt, Object...args) {
         System.err.println(String.format(fmt, args));
         System.exit(1);
-    }
-
-    private static <T> Supplier<T> die(String fmt, Object... args) {
-        return () -> {
-            exit(fmt, args);
-            return null;
-        };
     }
 
     private static String gitConfig(String key) {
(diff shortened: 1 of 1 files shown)


edvbld:
Hi Christoph,

thanks for contributing! Just two small nits before I will continue to review this:

1. Please describe what you changed in this pull request. You did not provide a description in this PR.
2. One usually do not end PR titles with a `.`, so please remove that from the title of the PR.

Thanks!
Erik

RealCLanger:
> Hi Christoph,
> 
> thanks for contributing! Just two small nits before I will continue to review this:
> 
> 1. Please describe what you changed in this pull request. You did not provide a description in this PR.
> 2. One usually do not end PR titles with a `.`, so please remove that from the title of the PR.
> 
> Thanks!
> Erik

Hi Erik,

thanks for pointing out these points. I think I've addressed them.

I created the pull request via the Skara tool call "git pr create". I'm wondering whether there is any documentation about this tool?
E.g. is there an option to give a pull request description? Also, how is one supposed to refer to a JBS bug (if I created one).
Furthermore, git pr has quite a few suboptions. Is there some documentation on what these all are supposed to do?

Thanks
Christoph

rwestberg:
The cli tools have documentation in man format here: https://github.com/openjdk/skara/tree/master/cli/resources/man/man1 - unfortunately, this documentation isn't complete yet..

Erik knows most about these tools, but I think when you are prompted to enter a description, it should be of the format:

<issue number>: title
<blank line>
Log description

edvbld:
Christopher, as Robin says, we unfortunately only have a man page at the moment (which won't help you very much since you are on Windows). You can get more abbreviated help information by running `git pr -h`. Robin is also right in how you add a longer multiline description to the pull request.

As for the regular expression, could you please add a comments above with a few examples. Also, do you really need the quotation marks around the path to `notepad.exe` for `core.editor`? Won't the following suffice:

```
core.editor = C:\Program Files\Notepad++\notepad++.exe -multiInst -notabbar -nosession -noPlugin
```

Thanks,
Erik

edvbld:
@RealCLanger did you see mine and Robin's comments?

RealCLanger:
Hi,

finally I got back to this.

> As for the regular expression, could you please add a comments above with a few examples.
I've done that.

> Also, do you really need the quotation marks around the path to `notepad.exe` for `core.editor`? Won't the following suffice:
> 
> ```
> core.editor = C:\Program Files\Notepad++\notepad++.exe -multiInst -notabbar -nosession -noPlugin

Probably yes, but that's how I found my git configuration without manual changes that I'm aware of. So it should work. And I think, even if I remove the quotation marks, it still wouldn't work.

Please review :)

edvbld:
Thanks Christopher! I agree, if it is possible to have quotation marks around the binary in `core.editor`, then we should support that. Looking at your patch, what do you think about doing something perhaps a bit simpler than a regular expression:

```java
var parts = editor.split(" ");
if (parts.length > 0) {
    var binary = parts[0];
    // Binary might be wrapped in quotes on Windows
    if (binary.startsWith("\"") && binary.endsWith("\"")) {
        parts[0] = binary.substring(1, binary.length() - 1);
    }
}
var pb = new ProcessBuilder(parts);
```

RealCLanger:
> Thanks Christopher! I agree, if it is possible to have quotation marks around the binary in `core.editor`, then we should support that. Looking at your patch, what do you think about doing something perhaps a bit simpler than a regular expression:
> 
> ```java
> var parts = editor.split(" ");
> if (parts.length > 0) {
>     var binary = parts[0];
>     // Binary might be wrapped in quotes on Windows
>     if (binary.startsWith("\"") && binary.endsWith("\"")) {
>         parts[0] = binary.substring(1, binary.length() - 1);
>     }
> }
> var pb = new ProcessBuilder(parts);
> ```

Hm, I agree that something simpler and more explicit than such a regexp would be desirable. But, as for your suggestion, what if the binary path contains blanks? Then it would be distributed along parts[0...n]. Furthermore, the options for the editor could also be quoted items with blanks in it. So, I guess the regexp would handle all that.

edvbld:
Ah right, my UNIX brain forgot about paths with spaces in them :smile: Yes, seems like we need the regular expression here. Just a quick question on the last part of the regex, `\\s*`, will that turn into groups? I wanted to make sure that we execute the editor _with_ all the arguments the user might have added.

RealCLanger:
> Ah right, my UNIX brain forgot about paths with spaces in them 😄 Yes, seems like we need the regular expression here. Just a quick question on the last part of the regex, `\\s*`, will that turn into groups? I wanted to make sure that we execute the editor _with_ all the arguments the user might have added.

The \\s* in the end will not turn into a group. It'll just lead to accepting trailing whitespace. I've tested the regex again in a small testprogram with several variations in input. One thing I've added is a \\s* in the beginning to accept leading whitespace - just in case. So the regex basically is:
-optional leading whitespace(non-quoted token|quoted token)optional trailing whitespace-
The group is just formed out of the token.

PR is updated, hopefully the final version :)

PR #260, opened 2019-11-20
Add config option for git-sync --pull

edvbld:
Hi all,

please review this small patch that enables configuring `git sync` to always pull the active branch (via the `sync.pull` configuration). If `sync.pull` is set to `always`, for example via `git config sync.pull always` then `git sync` will always try to pull the active branch after a successful sync.

Thanks,
Erik

## Testing
- [x] Manual testing of `git sync` with and without configuration setting

## Progress

## Approvers


edvbld:
Yeah, the macOS test failure is not because of this change. I agree we can integrate this patch and follow up on test failure later.

PR #262, opened 2019-11-21
git-sync should use APIs to find upstream

edvbld:
Hi all,

this patch changes `git-sync` to to use the `Forge` APIs for finding the upstream repository for a personal fork (in case the `upstream` remote is missing). The big part of this change is making `Forge.repository` return `Optional<HostedRepository>` instead of `HostedRepository` (because a forge may not have the requested repository). Robin, what do you think about these changes?

Thanks,
Erik

## Testing
- [x] Manual testing of `git sync`, both with and without an `upstream` remote

## Progress

## Approvers
 * Robin Westberg ([rwestberg](@rwestberg) - **Reviewer**) **Note!** Review applies to 20e4e200121b9fdd6962caf910962f60fd13a71b

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -137,7 +137,9 @@ class CheckRun {
                 var sourceBranch = mergeSourceBranch();
                 if (sourceBranch.isPresent() && sourceRepo.isPresent()) {
                     try {
-                        var mergeSourceRepo = pr.repository().forge().repository(sourceRepo.get()).get();
+                        var mergeSourceRepo = pr.repository().forge().repository(sourceRepo.get()).orElseThrow(() ->
+                                new RuntimeException("Could not find repository " + sourceRepo.get())
+                        );
                         try {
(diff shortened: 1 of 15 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -137,7 +137,7 @@ private boolean checkCommitAuthor(List<Commit> commits) throws IOException {
                 var sourceBranch = mergeSourceBranch();
                 if (sourceBranch.isPresent() && sourceRepo.isPresent()) {
                     try {
-                        var mergeSourceRepo = pr.repository().forge().repository(sourceRepo.get());
+                        var mergeSourceRepo = pr.repository().forge().repository(sourceRepo.get()).get();
```
rwestberg: Perhaps this (and the other similar ones) could be replaced with `.orElseThrow()` instead? A few are written like this already.

edvbld: Yeah, I agree. Although the effect of a failed `.get()` will be similar (a `NoSuchElementException` will be thrown), I agree that `.orElseThrow()` expresses the intent more clearly to the reader.

PR #267, opened 2019-11-24
Ensure Python 3 compatibility for the hg extension

rwestberg:
Hi all,

Please review this change that makes Skara's internally used Mercurial extension work with Python 3, while also retaining Python 2 compatibility. The Mercurial distribution on MacOS (through homebrew) recently switched to using Python 3, so this change will make the CI for Mac pass again.

Best regards,
Robin

## Progress

## Approvers

### DIFF
--- a/vcs/src/main/resources/ext.py
+++ b/vcs/src/main/resources/ext.py
@@ -49,11 +49,14 @@ def ratio(a, b, threshold):
     return ratio
 
 def write(s):
-    sys.stdout.buffer.write(s)
+    if sys.version_info >= (3, 0):
+        sys.stdout.buffer.write(s)
+    else:
+        sys.stdout.write(s)
 
 def writeln(s):
     write(s)
-    sys.stdout.buffer.write(b'\n')
+    write(b'\n')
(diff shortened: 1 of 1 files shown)


edvbld:
Robin what kind of testing have you done on this change? Could you please try to do a fresh conversion of one the smaller repositories, like jtreg, to ensure that we haven't broken anything? The unit tests exercise the Mercurial extension quite a bit, but I really want to ensure we don't mess anything up with regards to the hg -> git conversion.

rwestberg:
🐣 The jtreg repository is actually partially converted in the hgbridge tests, but yes, it makes sense do do something a little more extensive.. I'll kick off a few bigger conversions.

rwestberg:
Converted the jdk and panama repositories, and they ended up with the same commit hashes, so I think it's looking good. This was using using Mercurial with Python 2, but we can probably keep using that for the conversions.

edvbld:
Ok, great! Yeah, I agree we can keep using Mercurial with Python 2 for the conversion.

PR #271, opened 2019-11-26
Save references to materialized branches locally

rwestberg:
Hi all,

Please review this change that ensure that when bots use `Repository.materialize`, they always save a named reference to what was fetched. This ensures that subsequent fetches will present these existing refs, and avoid re-fetching commits that already exist locally.

It also fixes a potential bug where the mlbridge archive reader could have fetched from an incorrect branch.

Best regards,
Robin

## Progress

## Approvers

### DIFF
--- a/bots/hgbridge/src/main/java/org/openjdk/skara/bots/hgbridge/ExporterConfig.java
+++ b/bots/hgbridge/src/main/java/org/openjdk/skara/bots/hgbridge/ExporterConfig.java
@@ -169,7 +169,7 @@ class ExporterConfig {
 
     public Converter resolve(Path scratchPath) throws IOException {
         var localRepo = Repository.materialize(scratchPath, configurationRepo.url(),
-                                               configurationRef + ":hgbridge_config_" + configurationRepo.name());
+                                               "+" + configurationRef + ":hgbridge_config_" + configurationRepo.name());
 
         var replacements = parseMap(localRepo.root(), replacementsFile,
                                     field -> new Hash(field.name()),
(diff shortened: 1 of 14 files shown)


bots/hgbridge/src/main/java/org/openjdk/skara/bots/hgbridge/ExporterConfig.java
```diff
@@ -168,7 +168,8 @@ void sponsors(List<String> sponsors) {
     }
 
     public Converter resolve(Path scratchPath) throws IOException {
-        var localRepo = Repository.materialize(scratchPath, configurationRepo.url(), configurationRef);
+        var localRepo = Repository.materialize(scratchPath, configurationRepo.url(),
+                                               configurationRef + ":hgbridge_config_" + configurationRepo.name());
```
edvbld: Why no `+` here to to imply `--force` to `git fech`?

rwestberg: No reason, good catch!

edvbld: Should we instead change `Repository.materialize` to always prepend the `+` to the refspec? Or do you think that makes it less clear at the call sites?

rwestberg: Not sure, it's something you only want if you have a target ref, and only if you are using git.. And since materialize is implemented as a default method in Repository, it might be a little odd to have that logic there. Perhaps it could be refactored to not be a default method though.

edvbld: Ah right. No, `Repository.materialize` is `static`, not `default` :wink: I think we should leave as it is then.

The only other approach I can think if something like `Repository.materalize(Path path, URI remote, String fromRef, String toRef, boolean force, boolean checkout`), but then the parameter list is becoming really long (and `hg` can't handle `fromRef != toRef` anyhow). So I think we just leave it as it is.

PR #275, opened 2019-11-27
146: git-webrev: Add support for file list argument

edvbld:
Hi all,

please review this pull request that add support for an input argument to `git-webrev`. The argument is a file that itself contains a list of files to include in the webrev. This input file also specifies the order in which the files should appear in the webrev's index.html file.

The main part of the patch is adding support for a list of paths to `ReadOnlyRepository.diff`.

One part of the bug that is not addressed yet is the support for specifying `-` as the input file, `git-webrev` should then read the list of file names (separated by `\n`) from stdin. I plan to address this in a follow-up patch since I wanted to keep this patch as small as possible.

Thanks,
Erik

## Testing
- [x] Added a new unit test
- [x] Manual testing of `git-webrev` with input file

## Progress

## Issue
[SKARA-146](https://bugs.openjdk.java.net/browse/SKARA-146): git-webrev: Add support for file list argument

## Approvers
 * Robin Westberg ([rwestberg](@rwestberg) - **Reviewer**) **Note!** Review applies to e0c3162fd5e8bc596d6632d14d8087537b93506a

### DIFF
--- a/vcs/src/main/resources/ext.py
+++ b/vcs/src/main/resources/ext.py
@@ -169,8 +169,8 @@ def diff_git_raw(ui, repo, rev1, rev2=None, *files, **opts):
     modified, added, removed = [set(l) for l in status[:3]]
 
     files = opts['files']
-    if files != '':
-        wanted = set(files.split(','))
+    if files != b'':
+        wanted = set(files.split(b','))
         modified = modified & wanted
         added = added & wanted
         removed = removed & wanted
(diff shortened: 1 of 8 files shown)


vcs/src/main/resources/ext.py
```diff
@@ -167,6 +167,14 @@ def diff_git_raw(ui, repo, rev1, rev2=None, **opts):
         status = repo.status(ctx1)
 
     modified, added, removed = [set(l) for l in status[:3]]
+
+    files = opts['files']
+    if files != '':
+        wanted = set(files.split(','))
```
rwestberg:
```suggestion
        wanted = set(files.split(b','))
```

This fails on python3

PR #278, opened 2019-11-28
92: GitRepository::currentBranch fails when on a detached HEAD

edvbld:
Hi all,

please review this pull request that makes `git webrev` work when the repository is in a "detached HEAD" state (i.e. there is no active branch). The main part of the change is making `ReadOnlyRepository::currentBranch` return `Optional<Branch>` instead of `Branch`.

Thanks,
Erik

## Testing
- [x] Manually running `git webrev` on a repository with a detached HEAD

## Progress

## Issue
[SKARA-92](https://bugs.openjdk.java.net/browse/SKARA-92): GitRepository::currentBranch fails when on a detached HEAD

## Approvers

### DIFF
--- a/cli/src/main/java/org/openjdk/skara/cli/GitPr.java
+++ b/cli/src/main/java/org/openjdk/skara/cli/GitPr.java
@@ -519,7 +519,11 @@ public class GitPr {
 
                 System.exit(0);
             }
-            var currentBranch = repo.currentBranch();
+            var currentBranch = repo.currentBranch().orElseGet(() -> {
+                    System.err.println("error: the repository is in a detached HEAD state");
+                    System.exit(1);
+                    return null;
+            });
             if (currentBranch.equals(repo.defaultBranch())) {
                 System.err.println("error: you should not create pull requests from the 'master' branch");
                 System.err.println("");
--- a/vcs/src/main/java/org/openjdk/skara/vcs/hg/HgRepository.java
+++ b/vcs/src/main/java/org/openjdk/skara/vcs/hg/HgRepository.java
@@ -599,21 +599,23 @@ public class HgRepository implements Repository {
 
     @Override
     public void rebase(Hash hash, String committerName, String committerEmail) throws IOException {
-        var current = currentBranch().name();
+        var current = currentBranch().orElseThrow(() ->
+                new IOException("No current branch to rebase upon")
+        );
         try (var p = capture("hg", "--config", "extensions.rebase=",
-                             "rebase", "--dest", hash.hex(), "--base", current)) {
+                             "rebase", "--dest", hash.hex(), "--base", current.name())) {
             await(p);
         }
     }
 
     @Override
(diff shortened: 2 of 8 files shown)


cli/src/main/java/org/openjdk/skara/cli/GitPr.java
```diff
@@ -519,7 +519,11 @@ public static void main(String[] args) throws IOException, InterruptedException
 
                 System.exit(0);
             }
-            var currentBranch = repo.currentBranch();
+            var currentBranch = repo.currentBranch().orElseGet(() -> {
+                    System.err.println("error: the repository is in a detached HEAD state");
```
rwestberg:
```suggestion
                    System.err.println("error: the repository is in a detached HEAD state. check out a branch to continue.. or something");
```

edvbld: I thought about this as well, but if you see the "detached HEAD" warning when doing e.g. `git checkout HEAD~` there is just no way we are gonna be able to write all that :smiley:

So I figured that the next best thing was to write something concise that is very easy to search for. If the user does not know what "detached HEAD state" means then he/she will immediately search for it and will most likely stumble upon [this explanation](https://www.git-tower.com/learn/git/faq/detached-head-when-checkout-commit) which is the first hit on both DuckDuckGo and Google for me.

I think that is the best alternative here, what do you say?

vcs/src/main/java/org/openjdk/skara/vcs/hg/HgRepository.java
```diff
@@ -599,21 +599,23 @@ public boolean isAncestor(Hash ancestor, Hash descendant) throws IOException {
 
     @Override
     public void rebase(Hash hash, String committerName, String committerEmail) throws IOException {
-        var current = currentBranch().name();
+        var current = currentBranch().orElseThrow(() ->
+                new IOException("No current branch to rebase upon")
```
rwestberg: Should this mention the detached head then as well perhaps?

cli/src/main/java/org/openjdk/skara/cli/GitPr.java
rwestberg: Sounds good!

vcs/src/main/java/org/openjdk/skara/vcs/hg/HgRepository.java
rwestberg: Scratch that, guess this is hg :)

edvbld: Yes, this is in `HgRepository`, so a "detached HEAD" is not possible. I actually don't think `currenBranch` _can_ fail for `HgRepository`, since the current branch is part of the commit metadata in Mercurial (not just a ref as in git). I added a message anyhow to distinguish this scenario.

PR #285, opened 2019-12-03
Add support for Resolved issue state

rwestberg:
Hi all,

Please review this small change that adds support for setting issues to a Resolved state. Currently only supported by Jira, GitLab and GitHub will treat it as Closed.

Best regards,
Robin

## Progress

## Approvers

### DIFF
--- a/issuetracker/src/main/java/org/openjdk/skara/issuetracker/Issue.java
+++ b/issuetracker/src/main/java/org/openjdk/skara/issuetracker/Issue.java
@@ -104,6 +104,7 @@ public interface Issue {
 
     enum State {
         OPEN,
+        RESOLVED,
         CLOSED
     }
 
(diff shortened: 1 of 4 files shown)


issuetracker/src/main/java/org/openjdk/skara/issuetracker/Issue.java
```diff
@@ -104,6 +104,7 @@
 
     enum State {
         OPEN,
+        RESOLVED,
```
rwestberg: Another option would be to simply map CLOSED to the Jira Resolved state as that's probably what we want to always use..

rwestberg: Jira also supports renaming its states, so it's possible that we'll need to switch to setting State as a string to be interpreted as the particular Issue implementation sees fit.

edvbld: I think using `RESOLVED` makes sense for now :+1:

rwestberg:
Thanks for reviewing!

PR #292, opened 2019-12-04
165: Replace large webrev files with placeholders instead of dropping them

rwestberg:
Hi all,

Please review this change that replaces large files in published webrevs with a placeholder message, instead of just omitting them.

Best regards,
Robin

## Progress

## Issue
[SKARA-165](https://bugs.openjdk.java.net/browse/SKARA-165): Replace large webrev files with placeholders instead of dropping them

## Approvers


edvbld:
This looks good in general, but won't this destroy the navigation arrows for .e.g. the udiff or frames view? Or maybe it just overwrites the content, I don't recall :smile:

rwestberg:
It would indeed destroy navigation, I've updated it a bit to handle .html files a bit nicer. This also ensures that the instructions are displayed with proper line breaks as they will be placed inside `<pre>` tags if possible.

PR #294, opened 2019-12-05
Cache labels in GitHubPullRequest

edvbld:
Hi all,

please review this pull request that introduces a simple label cache for `GitHubPullRequest`. Since a `GitHubPullRequest` instance can't be accessed by multiple threads it is safe to use a very simple cache mechanism. The only calls that can mutate labels on a PR are `addLabel` and `removeLabel`. A user on GitHub could of course add a label, but we make no guarantees that a `GitHubPullRequest` instance reflects what happens on GitHub, code wanting to ensure they have the latest state must create a new `GitHubPullRequest` instance.

Thanks,
Erik

## Progress

## Approvers


edvbld:
Thanks for reviewing Robin! Yes, I agree that there are other parts of `GitHubPullRequest` that it would be worthwhile to cache, but lets tackle that in another PR.

PR #295, opened 2019-12-05
Try a bit harder to generate incremental webrevs

rwestberg:
Hi all,

Please review this change that makes the mailing list bridge bot try a little harder to generate an incremental webrev. If a pull request has been rebased or merged the latest changes from the target branch, it may be possible to exclude these and generate a clean incremental webrev. If so, we do that.

Best regards,
Robin

## Progress

## Approvers


rwestberg:
Thanks for reviewing! Yes, will be interesting to see how often this works in practice!

PR #324, opened 2019-12-12
199: Set fixVersion in an issue when a commit is integrated

rwestberg:
Hi all,

Please review this change that sets fixVersion in a commit's issue after integration, if a version has been configured in the repository's .jcheck/conf file.

Also sets a version for Skara itself!

Best regards,
Robin

## Progress

## Issue
[SKARA-199](https://bugs.openjdk.java.net/browse/SKARA-199): Set fixVersion in an issue when a commit is integrated

## Approvers

### DIFF
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/IssueUpdater.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/IssueUpdater.java
@@ -37,14 +37,20 @@ import java.util.logging.Logger;
 
 public class IssueUpdater implements RepositoryUpdateConsumer, PullRequestUpdateConsumer {
     private final IssueProject issueProject;
+    private final boolean reviewLink;
     private final URI reviewIcon;
+    private final boolean commitLink;
     private final URI commitIcon;
+    private final String fixVersion;
     private final Logger log = Logger.getLogger("org.openjdk.skara.bots.notify");
 
-    IssueUpdater(IssueProject issueProject, URI reviewIcon, URI commitIcon) {
+    IssueUpdater(IssueProject issueProject, boolean reviewLink, URI reviewIcon, boolean commitLink, URI commitIcon, String fixVersion) {
         this.issueProject = issueProject;
+        this.reviewLink = reviewLink;
         this.reviewIcon = reviewIcon;
+        this.commitLink = commitLink;
         this.commitIcon = commitIcon;
+        this.fixVersion = fixVersion;
     }
 
     @Override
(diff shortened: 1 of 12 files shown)


kevinrushforth:
I don't think it is workable to have the configuration stored in the repo. We will have many cases where there will be upstream and downstream repos where commits will be synced from one to the other and you wouldn't want the fix version to be synced. Also, when forking repos or branches you want to set up and configure the new branch/repo and the master branch at the same time of the fork without pushing a commit to the repo. There are also cases (e.g., for sandbox repos) where you might not want the gitupdater feature enabled, and again it would be unworkable for the config to be in the repo.

I think this is better done via a separate database that maps each repo/branch to a fix version as well as enabling the auto update or not.

rwestberg:
I think there may be projects where it would be useful to have the version information stored in the repository - for example, it would be immediately obvious when the version is bumped. But as you say, it may not be the case for all projects. I'll update the change to allow this to be configurable per project: Either a fixed version, read the version from .jcheck/conf, or don't set the version at all.

bots/notify/src/main/java/org/openjdk/skara/bots/notify/IssueUpdater.java
```diff
@@ -59,29 +68,48 @@ public void handleCommits(HostedRepository repository, List<Commit> commits, Bra
...
+                    var linkBuilder = Link.create(repository.webUrl(commit.hash()), "Commit")
+                                          .summary(repository.name() + "/" + commit.hash().abbreviate());
+                    if (commitIcon != null) {
+                        linkBuilder.iconTitle("Commit");
+                        linkBuilder.iconUrl(commitIcon);
+                    }
+                    issue.get().addLink(linkBuilder.build());
+                }
+
+                if (fixVersion != null) {
```
edvbld: How about turning this around a bit into:

```java
if (fixVersion == null) {
   // Try to find fixVersion from .jceck/conf
}
if (fixVersion != null) {
    issue.get().addFixVersion(fixVersion)
}
```
If you don't sent the `fixVersion` via the bot configuration, then it will be `null`, and then we will read it from `.jcheck/conf`. This avoids the need for the special `<repo>` version.

rwestberg: Right, but if you create a fork where you don't want to change the .jcheck/conf file (for merge reasons) you may perhaps not want the fixVersion updated?

edvbld: Good point! Should we then perhaps represent this using two parameters, `boolean shouldUpdateFixVersion` and `String fixVersion`?

PR #329, opened 2019-12-13
202: git-pr create should use a proper Markdown file

edvbld:
Hi all,

please review this patch that updates `git pr create` to show a valid Markdown
file. The instructions have been moved into HTML comments looking like the
following:

```html
<!-- Please enter the pull request message for your changes.                --> 
<!-- The first line will be considered the subject, use a blank line to     --> 
<!-- separate the subject from the body. These HTML comment lines will      --> 
<!-- be removed automatically. An empty message aborts the pull request.    --> 
<!--                                                                        --> 
<!-- Commits to be included from branch 'skara-202':                        --> 
<!-- - 1633065c: skara-202                                                  --> 
<!--   M  cli/src/main/java/org/openjdk/skara/cli/GitPr.java                --> 
<!--                                                                        --> 
<!-- Issue:      https://bugs.openjdk.java.net/browse/SKARA-202             --> 
<!-- Repository: https://github.com/openjdk/skara                           --> 
<!-- Branch:     master                                                     --> 
```

I also took the opportunity to tweak the message a bit. Furthermore I added the
status for each commit so that it is easy to see changed files and also added a
link to an eventual issue. Personally I think the end result turned out quite
nice.

Thanks,
Erik

## Testing
- [x] Manual testing of `git pr create`

## Progress

## Issue
[SKARA-202](https://bugs.openjdk.java.net/browse/SKARA-202): git-pr create should use a proper Markdown file

## Approvers

### DIFF
--- a/cli/src/main/java/org/openjdk/skara/cli/GitPr.java
+++ b/cli/src/main/java/org/openjdk/skara/cli/GitPr.java
@@ -44,6 +44,8 @@ import java.util.logging.Level;
 import java.util.stream.Collectors;
 
 public class GitPr {
+    private static final StandardOpenOption APPEND = StandardOpenOption.APPEND;
+
     private static void exit(String fmt, Object...args) {
         System.err.println(String.format(fmt, args));
         System.exit(1);
@@ -54,6 +56,17 @@ public class GitPr {
             exit(fmt, args);
             return null;
         };
+    }
+
+    private static String rightPad(String s, int length) {
+        return String.format("%-" + length + "s", s);
+    }
+
+    private static void appendPaddedHTMLComment(Path file, String line) throws IOException {
+        var end = " -->";
+        var pad = 79 - end.length();
+        var newLine = "\n";
+        Files.writeString(file, rightPad("<!-- " + line, pad) + end + newLine, StandardOpenOption.APPEND);
     }
 
     private static String format(Issue issue) {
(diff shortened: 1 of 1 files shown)


cli/src/main/java/org/openjdk/skara/cli/GitPr.java
```diff
@@ -729,31 +742,51 @@ public static void main(String[] args) throws IOException, InterruptedException
...
+
             var success = spawnEditor(repo, file);
             if (!success) {
                 System.err.println("error: editor exited with non-zero status code, aborting");
                 System.exit(1);
             }
             var lines = Files.readAllLines(file)
                              .stream()
-                             .filter(l -> !l.startsWith("#"))
+                             .filter(l -> !(l.startsWith("<!--") && l.endsWith("-->")))
```
rwestberg: Perhaps add a .map(String::strip) before this one to prevent whitespace mistakes?

edvbld: Yeah, I was thinking about that, but then I realized that _if_ you want to include an HTML comment in your PR text (like this PR does) then the easiest way to do that is to just insert a space at the end of the sentence :smiley:

Not sure how common it will be to include HTML comments though, this is the first time I've done it, but I think I prefer the "simple" semantics that the line strictly has to start and end with `<!--` and `-->`.

PR #339, opened 2019-12-17
212: Retry notifications for updaters that can check current state

rwestberg:
Hi all,

Please review this change that enables issue notifications to be retried in case transient server errors occur, but will not send out duplicate email notifications.

Best regards,
Robin

## Progress

## Issue
[SKARA-212](https://bugs.openjdk.java.net/browse/SKARA-212): Retry notifications for updaters that can check current state

## Approvers


bots/notify/src/main/java/org/openjdk/skara/bots/notify/PullRequestUpdateConsumer.java
```diff
@@ -28,4 +28,5 @@
 public interface PullRequestUpdateConsumer {
     void handleNewIssue(PullRequest pr, Issue issue);
     void handleRemovedIssue(PullRequest pr, Issue issue);
+    boolean idempotent();
```
edvbld:
```suggestion
    boolean isIdempotent();
```

PR #345, opened 2020-01-08
215: Verify webrev links before considering publication successful

rwestberg:
Hi all,

Please review this change which ensures that webrev links can be successfully reached before returning from the publication step. 

Best regards,
Robin

## Progress

## Issue
[SKARA-215](https://bugs.openjdk.java.net/browse/SKARA-215): Verify webrev links before considering publication successful

## Approvers

### DIFF
--- a/bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/WebrevStorage.java
+++ b/bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/WebrevStorage.java
@@ -30,9 +30,12 @@ import org.openjdk.skara.webrev.Webrev;
 
 import java.io.*;
 import java.net.URI;
+import java.net.http.*;
 import java.nio.file.*;
-import java.util.Comparator;
+import java.time.*;
+import java.util.*;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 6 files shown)


bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/WebrevStorage.java
```diff
@@ -141,6 +145,33 @@ private static void clearDirectory(Path directory) {
...
+                               .build();
+        while (Instant.now().isBefore(end)) {
+            var uncachedUri = uriBuilder.setQuery(Map.of("nocache", UUID.randomUUID().toString())).build();
+            var request = HttpRequest.newBuilder(uncachedUri)
+                                     .timeout(Duration.ofSeconds(30))
+                                     .GET()
+                                     .build();
+            try {
+                var response = client.send(request, HttpResponse.BodyHandlers.ofString());
+                if (response.statusCode() < 400) {
```
edvbld:
```suggestion
                if (respons.statusCode() >= 200 && response.statusCode() < 400) {
```

rwestberg: I don't think < 200's propagate outside of the http client implementation, so should be reasonably safe.

rwestberg:
Thanks for reviewing!

PR #357, opened 2020-01-16
Assign unassigned issues when resolving them

rwestberg:
Hi all,

Please review this change that sets the assignee of unassigned issues when resolving them.

Best regards,
Robin

## Progress

## Approvers
 * Erik Helin ([ehelin](@edvbld) - **Reviewer**) **Note!** Review applies to 151d7b1f85d8c8c019b982dfe7ac73d58ce016f4

### DIFF
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/IssueUpdater.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/IssueUpdater.java
@@ -22,7 +22,9 @@
  */
 package org.openjdk.skara.bots.notify;
 
+import org.openjdk.skara.email.*;
 import org.openjdk.skara.forge.*;
+import org.openjdk.skara.host.HostUser;
 import org.openjdk.skara.issuetracker.Issue;
 import org.openjdk.skara.issuetracker.*;
 import org.openjdk.skara.jcheck.JCheckConfiguration;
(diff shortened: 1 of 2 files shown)


bots/notify/src/main/java/org/openjdk/skara/bots/notify/IssueUpdater.java
```diff
@@ -216,6 +218,20 @@ private Issue findIssue(Issue primary, String fixVersion) {
...
+            return Optional.of(issueProject.issueTracker().user(authorEmail.localPart()));
+        } else {
+            var committerEmail = EmailAddress.from(commit.committer().email());
+            if (!committerEmail.domain().equals("openjdk.org")) {
+                log.severe("Cannot determine issue tracker user name from committer email: " + committerEmail);
+                return Optional.empty();
+            }
+            return Optional.of(issueProject.issueTracker().user(committerEmail.localPart()));
+        }
+    }
```
edvbld: Instead of repeating `Optional.of(issueProject.issueTracker().user(...))`, you might want to code this like:
```
        String username = null;

        var authorEmail = EmailAddress.from(commit.author().email());
        if (authorEmail.domain().equals("openjdk.org")) {
            username = authorEmail.localPart();
        } else {
            var committerEmail = EmailAddress.from(commit.committer().email());
            if (!committerEmail.domain().equals("openjdk.org")) {
                log.severe("Cannot determine issue tracker user name from committer email: " + committerEmail);
            }
            username = committerEmail.localPart();
        }

        return username == null ? Optional.empty() : Optional.of(issueProject.issueTracker().user(username));
```

edvbld: Oh, I should mention that I'm fine your version as well :+1: If you want to use your version with early return, then please remove the `else {` part after the return :smiley:

rwestberg: Fair enough, how about this? :)

PR #364, opened 2020-01-17
217: Enforce check on number of required reviewers

rwestberg:
Hi all,

Please review this change that allows Reviewers of a PR to override the default number of required reviews as configured in the jcheck configuration file.

Best regards,
Robin

## Progress

## Issue
[SKARA-217](https://bugs.openjdk.java.net/browse/SKARA-217): Enforce check on number of required reviewers

## Approvers

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/AdditionalConfiguration.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/AdditionalConfiguration.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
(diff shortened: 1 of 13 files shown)


kevinrushforth:
In addition to allowing a reviewer to increase the required number of reviewers, which is the use case described in the Enhancement request, this will allow a reviewer to reduce the number of required reviewers below the minimum set in `.jcheck/conf`, including setting it to 0, right? Is this what we want?

kevinrushforth:
One more question: If I increase the number of required reviewers to 2, does that mean both have to have the Reviewer role? If so, this won't be nearly as useful to us on the openjfx project. I suspect the same is true of many sub-projects in the JDK itself. For example, the policy in the client team is that fixes in general need 2 reviewers, at least one of which must be a "R" reviewer. OpenJFX has a similar policy (except that only "high-impact" fixes need a second reviewer) in that one one of the reviewers needs to be a "R" reviewer.

rwestberg:
Now that #368 has been implemented, I'll update this to support more complex reviewer requirements. Also, it makes sense not to allow relaxing the requirements below what the jcheck configuration specifies.

nlisker:
> `/reviewers <n> [<role>]` where `<n>` is the additional number of required reviewers.

I fear that this will be confusing. I expect that `n` is the number of reviewers, not the number of additional reviewers.

kevinrushforth:
I agree with @nlisker. I believe that both the `jdk` and `openjfx` projects will set 1 reviewer in the .jcheck/conf file (which must be someone with a Reviewer role), with the common use case to increase that being to bump the required reviewers to 2 or maybe 3 where the additional reviewers don't have a required role. If I want to add a requirement for 2 reviewers, I would expect to be able to type:

    /reviewers 2

And the bot would figure out that it means one extra reviewer with no required role in the project. Anything else would be a "nice to have", but I wouldn't want to make it more confusing to do the simple thing.

kevinrushforth:
Thanks for making the requested update. To check that it matches my understanding, if the default policy as set in `.jcheck/conf` is 1 reviewer with a role of `Reviewer`, then I expect the following:

`/reviewers 0` -- error
`/reviewers 1` -- reset to default (1 reviewer with role of `Reviewer` in this example)
`/reviewers 2` -- num reviewers is now two, one of which must have the role of `Reviewer` and the other must have the role of `Author` (or higher)

Is my understanding correct? It seems so in looking at your most recent change, but wanted to double-check.

rwestberg:
Yep, that's correct!

nlisker:
Looks good. After this is integrated, when will it take effect? Reviewers need to be notified.

rwestberg:
Will probably deploy it later this week.

PR #368, opened 2020-01-20
234: Allow more granular jcheck reviewer configuration

edvbld:
Hi all,

please review this patch that allows for more granular reviewers configuration
with jcheck. With this patch a project can now specify exactly how many
reviewers are need for each role, for example:

```
[checks "reviewer"]
reviewer = 1
committer = 1
author = 1
```

The above configuration would require three reviewers in total, out of which one
must be reviewer, one must be committer and one must be author. This granularity
of configuration is needed to fully implement
[PR 364](https://git.openjdk.java.net/skara/pulls/364).

I made sure that the patch is backwards compatible, we still support the old
configuration syntax:

```
[checks "reviewer"]
minimum = 1
role = reviewer
```

The above will simply be translated to the following configuration:

```
[checks "reviewer"]
reviewer = 1
```

Thanks,
Erik

## Testing
- [x] Added a number of additional unit tests

### DIFF
--- a/jcheck/src/test/java/org/openjdk/skara/jcheck/ReviewersCheckTests.java
+++ b/jcheck/src/test/java/org/openjdk/skara/jcheck/ReviewersCheckTests.java
@@ -57,6 +57,9 @@ class ReviewersCheckTests {
         "  <person name=\"qux\">",
         "    <full-name>Qux</full-name>",
         "  </person>",
+        "  <person name=\"contributor\">",
+        "    <full-name>Contributor</full-name>",
+        "  </person>",
         "  <group name=\"test\">",
         "    <full-name>Test</full-name>",
         "    <person ref=\"foo\" role=\"lead\" />",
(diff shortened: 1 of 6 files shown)


jcheck/src/test/java/org/openjdk/skara/jcheck/ReviewersCheckTests.java
```diff
@@ -281,4 +300,87 @@ void ignoredReviewersShouldBeExcluded() throws IOException {
...
+    void relaxedRoleShouldPass() throws IOException {
+        var commit = commit(List.of("bar", "qux"));
+        var check = new ReviewersCheck(census(), utils);
+        var issues = toList(check.check(commit, message(commit), conf(0, 1, 1)));
+
+        assertEquals(0, issues.size());
+    }
+
+    @Test
+    void relazedRoleAndMissingRoleShouldFail() throws IOException {
```
rwestberg:
```suggestion
    void relaxedRoleAndMissingRoleShouldFail() throws IOException {
```

PR #377, opened 2020-01-23
Make it possible to only allow PRs against a select set of target branches

rwestberg:
Hi all,

Please review this change that makes it possible to only allow PRs to target a select set of branches.

Best regards,
Robin

## Progress

## Approvers
 * Erik Helin ([ehelin](@edvbld) - **Reviewer**) **Note!** Review applies to 7b261d73abbb0367ddc413e52e6f110da4facb43

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -43,8 +43,6 @@ class CheckRun {
     private final List<Review> activeReviews;
     private final Set<String> labels;
     private final CensusInstance censusInstance;
-    private final Map<String, String> blockingLabels;
-    private final IssueProject issueProject;
 
     private final Logger log = Logger.getLogger("org.openjdk.skara.bots.pr");
     private final String progressMarker = "<!-- Anything below this marker will be automatically updated, please do not edit manually! -->";
@@ -54,7 +52,7 @@ class CheckRun {
 
     private CheckRun(CheckWorkItem workItem, PullRequest pr, PullRequestInstance prInstance, List<Comment> comments,
                      List<Review> allReviews, List<Review> activeReviews, Set<String> labels,
-                     CensusInstance censusInstance, Map<String, String> blockingLabels, IssueProject issueProject) {
+                     CensusInstance censusInstance) {
         this.workItem = workItem;
         this.pr = pr;
         this.prInstance = prInstance;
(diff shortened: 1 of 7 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -64,17 +62,19 @@ private CheckRun(CheckWorkItem workItem, PullRequest pr, PullRequestInstance prI
...
-                        List<Review> allReviews, List<Review> activeReviews, Set<String> labels, CensusInstance censusInstance, Map<String, String> blockingLabels,
-                        IssueProject issueProject) {
-        var run = new CheckRun(workItem, pr, prInstance, comments, allReviews, activeReviews, labels, censusInstance, blockingLabels, issueProject);
+                        List<Review> allReviews, List<Review> activeReviews, Set<String> labels, CensusInstance censusInstance) {
+        var run = new CheckRun(workItem, pr, prInstance, comments, allReviews, activeReviews, labels, censusInstance);
         run.checkStatus();
     }
 
+    private boolean checkTargetBranch() {
+        var matcher = workItem.bot.allowedTargetBranches().matcher(pr.targetRef());
```
edvbld: Should we use a pattern or should we simply just enumerate the protected branches? It seems potentially overkill to support a regex, it seems like an anti-pattern to have a repository with so many protected branches that they are cumbersome to enumerate, so you need a regex :smiley:

rwestberg:
How about enumerating the valid targets in the error message?

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -64,17 +62,29 @@ private CheckRun(CheckWorkItem workItem, PullRequest pr, PullRequestInstance prI
...
     static void execute(CheckWorkItem workItem, PullRequest pr, PullRequestInstance prInstance, List<Comment> comments,
-                        List<Review> allReviews, List<Review> activeReviews, Set<String> labels, CensusInstance censusInstance, Map<String, String> blockingLabels,
-                        IssueProject issueProject) {
-        var run = new CheckRun(workItem, pr, prInstance, comments, allReviews, activeReviews, labels, censusInstance, blockingLabels, issueProject);
+                        List<Review> allReviews, List<Review> activeReviews, Set<String> labels, CensusInstance censusInstance) {
+        var run = new CheckRun(workItem, pr, prInstance, comments, allReviews, activeReviews, labels, censusInstance);
         run.checkStatus();
     }
 
+    private boolean checkTargetBranch() {
```
edvbld:
```suggestion
    private boolean isTargetBranchAllowed() {
```

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -114,6 +124,14 @@ private boolean checkCommitAuthor(List<Commit> commits) throws IOException {
     private List<String> botSpecificChecks() throws IOException {
         var ret = new ArrayList<String>();
 
+        if (!checkTargetBranch()) {
```
edvbld:
```suggestion
        if (!isTargetBranchAllowed()) {
```

rwestberg: Sounds good, will change!

PR #379, opened 2020-01-24
245: CLI tools do not consider git's http.proxy configuration

edvbld:
Hi all,

please review this patch that makes the CLI tools respect git's
[http.proxy](https://git-scm.com/docs/git-config#Documentation/git-config.txt-httpproxy)
setting. The behavior with this patch is if `http_proxy`, `https_proxy` and/or
`no_proxy` is *not* set, then we try to set the proxy based on git's
`http.proxy` config (i.e. `http_proxy` and friends have precedence).

Thanks,
Erik

## Testing
- [x] Manual testing of CLI tools behind http(s) proxy

## Progress

## Issue
[SKARA-245](https://bugs.openjdk.java.net/browse/SKARA-245): CLI tools do not consider git's http.proxy configuration

## Approvers


rwestberg:
Should the precedence order perhaps be changed to match how git treats the configuration?

PR #381, opened 2020-01-24
243: Merge bot should support merging with certain frequency

edvbld:
Hi all,

please review this patch that enables the merge bot to periodically sync
branches with a certain frequency. This is implemented by extending the
configuration for a merge specification:

```
"merge": {
    "target": "github:openjdk/panama",
    "specs": [
        {
	    "from": "github:openjdk/jdk:master",
	    "to": "master",
	    "frequency": {
	        "interval": "weekly",
		"weekday": "Friday",
		"hour": "11"
	    }
	}
    ]
}
```

The above configuration would merge the `master` branch from the
[jdk](https://github.com/openjdk/jdk) repository to the `master` branch in the
[panama](https://github.com/openjdk/panama) repository every Friday at 11:00.
The merge will only be done once, i.e. it won't be retried until the next Friday
if the merge fails (and result in a merge conflict PR).

There is one limitation with the patch and that is that the merge bot only
"remember" previous merges in memory. This is problematic if we redeploy the
merge bot during 11:00 - 11:59 on a Friday and the merge already has occurred,
the merge bot will then do a second merge. This is known limitation and one I
think we can live with, I don't foresee it to become a problem in practice.

Thanks,
Erik

## Testing
- [x] Added a bunch of new unit tests

## Progress

## Issue
[SKARA-243](https://bugs.openjdk.java.net/browse/SKARA-243): Merge bot should support merging with certain frequency

## Approvers

### DIFF
--- a/bots/merge/src/main/java/org/openjdk/skara/bots/merge/MergeBotFactory.java
+++ b/bots/merge/src/main/java/org/openjdk/skara/bots/merge/MergeBotFactory.java
@@ -27,6 +27,8 @@ import org.openjdk.skara.vcs.Branch;
 
 import java.io.*;
 import java.nio.file.Files;
+import java.time.DayOfWeek;
+import java.time.Month;
 import java.util.*;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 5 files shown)


bots/merge/src/main/java/org/openjdk/skara/bots/merge/MergeBotFactory.java
```diff
@@ -38,6 +40,96 @@ public String name() {
...
+                return DayOfWeek.SATURDAY;
+            case "sunday":
+                return DayOfWeek.SUNDAY;
+            default:
+                throw new IllegalArgumentException("Unknown weekday: " + s);
+        }
+    }
+
+    private static Month toMonth(String s) {
+        switch (s.toLowerCase()) {
```
rwestberg: Would perhaps be a little less verbose to use Month.from with an appropriate formatter, but can keep it as-is if you prefer. :)

edvbld: Yeah, I agree, but since I've already written the verbose version, lets just keep it for now :smiley:

PR #382, opened 2020-01-24
245: Fix proxy protocol setting, add logging

rwestberg:
Hi all,

Please review this follow-up change that fixes the proxy type selection, and also adds some logging. In addition, if proxy settings are picked up from the environment, nonProxyHosts are now always set. This avoids a problem when nonProxyHosts has a default value.

Best regards,
Robin

## Progress

## Issue
[SKARA-245](https://bugs.openjdk.java.net/browse/SKARA-245): CLI tools do not consider git's http.proxy configuration

## Approvers
 * Erik Helin ([ehelin](@edvbld) - **Reviewer**) **Note!** Review applies to bdae54cc12c069edac864bdaa60e7628be1c5106

### DIFF
--- a/proxy/src/main/java/org/openjdk/skara/proxy/HttpProxy.java
+++ b/proxy/src/main/java/org/openjdk/skara/proxy/HttpProxy.java
@@ -23,21 +23,27 @@
 package org.openjdk.skara.proxy;
 
 import java.io.IOException;
-import java.net.URI;
-import java.net.URISyntaxException;
+import java.net.*;
 import java.nio.charset.StandardCharsets;
-import java.util.Arrays;
-import java.util.List;
+import java.util.*;
+import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
 public class HttpProxy {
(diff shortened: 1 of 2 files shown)


proxy/src/main/java/org/openjdk/skara/proxy/HttpProxy.java
```diff
@@ -49,49 +55,50 @@ public static void setup() {
...
-                    var uri = new URI(output);
-                    setProxyHostAndPortBasedOn(uri);
-                } else {
-                    for (var scheme : List.of("http", "https")) {
-                        var uri = new URI(scheme + "://" + output);
-                        setProxyHostAndPortBasedOn(uri);
-                    }
+            if (res == 0 && !output.isEmpty()) {
+                if (!output.startsWith("http://") && !output.startsWith("https://")) {
+                    output = "http://" + output;
```
edvbld: Add a comment that the protocol being prepended doesn't matter?

rwestberg: Sure, sounds reasonable.

PR #397, opened 2020-02-05
Use very defensive programming in CSRCommand

edvbld:
Hi all,

please review this follow-up patch to 5632c91342c703d2197aad402492c3679762638e
that uses even more defensive programming to avoid `null` pointers.

Thanks,
Erik

## Progress

## Approvers


edvbld:
> This should do it!
:crossed_fingers:

PR #403, opened 2020-02-06
Add toString methods to all WorkItems

edvbld:
Hi all,

please review this small patch that adds proper `toString` implementations to
all classes implementing `WorkItem` (to get better logging).

Thanks,
Erik

## Progress

## Approvers
 * Robin Westberg ([rwestberg](@rwestberg) - **Reviewer**) **Note!** Review applies to 1f8ffb736f135a60cfdc24a8aea6e22d556fda81

### DIFF
--- a/bots/mirror/src/main/java/org/openjdk/skara/bots/mirror/MirrorBot.java
+++ b/bots/mirror/src/main/java/org/openjdk/skara/bots/mirror/MirrorBot.java
@@ -85,7 +85,7 @@ class MirrorBot implements Bot, WorkItem {
 
     @Override
     public String toString() {
-        return "MirrorBot@" + from.name() + "-> " + to.name();
+        return "MirrorBot@" + from.name() + "->" + to.name();
     }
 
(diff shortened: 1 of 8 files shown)


bots/mirror/src/main/java/org/openjdk/skara/bots/mirror/MirrorBot.java
```diff
@@ -85,7 +85,7 @@ public void run(Path scratchPath) {
 
     @Override
     public String toString() {
-        return "MirrorBot@(" + from.name() + "-> " + to.name() + ")";
+        return "MirrorBot@" + from.name() + "-> " + to.name();
```
rwestberg:
```suggestion
        return "MirrorBot@" + from.name() + " -> " + to.name();
```

rwestberg: Or the other way around 🙂

PR #430, opened 2020-02-12
test-bot: better logging for IOException

edvbld:
Hi all,

please review this small patch that updates the logging for the test bot when it
receives an `IOException`.

Testing:

Thanks,
Erik

## Progress

## Approvers
 * Robin Westberg ([rwestberg](@rwestberg) - **Reviewer**) **Note!** Review applies to c90268cd9106d148f705d0fc4ac726237404599c

### DIFF
--- a/bots/tester/src/main/java/org/openjdk/skara/bots/tester/TestBot.java
+++ b/bots/tester/src/main/java/org/openjdk/skara/bots/tester/TestBot.java
@@ -97,7 +97,7 @@ public class TestBot implements Bot {
                     jobs = ci.jobsFor(pr);
                 } catch (IOException e) {
                     log.info("Could not retrieve test jobs for PR: " + desc);
-                    log.info(e.getMessage());
+                    log.throwing("TestBot", "getPeriodicItems", e);
                 }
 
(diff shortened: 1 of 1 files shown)


bots/tester/src/main/java/org/openjdk/skara/bots/tester/TestBot.java
```diff
@@ -88,38 +90,43 @@
...
-                            if (!states.containsKey(job.id())) {
-                                shouldUpdate = true;
-                                states.put(job.id(), new Observation(job.state(), job.state()));
-                            } else {
-                                var observed = states.get(job.id());
+                    log.info("Getting test jobs for " + desc);
+                    jobs = ci.jobsFor(pr);
+                } catch (IOException e) {
+                    log.info("Could not retrieve test jobs for PR: " + desc);
+                    log.info(e.getMessage());
```
rwestberg: Perhaps use log.throwing here?

edvbld: Ah, good point, wasn't aware of that one :+1:

PR #439, opened 2020-02-14
279: Optionally only update project-specific issues

rwestberg:
Hi all,

Please review this change that allows the issue notifier to only update issues when the commit originates from a matching PR. This can be used for projects that merge commits from/to different branches, but only want to be notified about the original commit resulting from a PR integration.

Best regards,
Robin

## Progress

## Issue
[SKARA-279](https://bugs.openjdk.java.net/browse/SKARA-279): Optionally only update project-specific issues 

## Approvers

### DIFF
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/IssueUpdater.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/IssueUpdater.java
@@ -257,14 +257,14 @@ public class IssueUpdater implements RepositoryUpdateConsumer, PullRequestUpdate
 
                 String requestedVersion = null;
                 if (prOnly) {
-                    var pullRequests = issue.links().stream()
+                    var pullRequestCount = issue.links().stream()
                                             .filter(link -> link.title().orElse("notitle").equals("Review"))
                                             .filter(link -> link.summary().orElse("nosummary").startsWith(repository.name() + "/"))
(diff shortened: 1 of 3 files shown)


bots/notify/src/main/java/org/openjdk/skara/bots/notify/IssueUpdater.java
```diff
@@ -252,25 +255,40 @@ public void handleCommits(HostedRepository repository, Repository localRepositor
...
-                                var version = parsed.general().version();
-                                requestedVersion = version.orElse(null);
+                if (prOnly) {
+                    var pullRequests = issue.links().stream()
+                                            .filter(link -> link.title().orElse("notitle").equals("Review"))
+                                            .filter(link -> link.summary().orElse("nosummary").startsWith(repository.name() + "/"))
+                                            .map(link -> link.summary().orElseThrow().substring(repository.name().length() + 1))
+                                            .map(repository::pullRequest)
+                                            .filter(pr -> pr.targetRef().equals(branch.name()))
+                                            .collect(Collectors.toList());
```
edvbld: Use [Stream.count](https://docs.oracle.com/en/java/javase/13/docs/api/java.base/java/util/stream/Stream.html#count()) here instead of materializing the list? You would have to change the following `if` to `if (count == 0L) {`.

rwestberg: Good point, will fix!

PR #444, opened 2020-02-17
git-webrev: create links for bug ids in commit messages

edvbld:
Hi all,

please review this patch that creates links to
[JBS](https://bugs.openjdk.java.net) for issue id:s in commit messages. This
patch also creates a link to the upstream revision if possible. Finally I also
updated some of the titles to match `webrev.ksh`.

The reason for not using the types `IssueProject` and `PullRequest` is that
`webrev.ksh` has traditionally never required a network connection. The bots
however are free to do so and can therefore provide better implementations for
`issueLinker` and `commitLinker`. This work will be done in a follow-up patch.

Testing:
- Manual testing of `git-webrev`

Thanks,
Erik

## Progress

## Reviewers

### DIFF
--- a/cli/src/main/java/org/openjdk/skara/cli/GitWebrev.java
+++ b/cli/src/main/java/org/openjdk/skara/cli/GitWebrev.java
@@ -83,6 +83,10 @@ public class GitWebrev {
         }
     }
 
+    private static boolean isDigit(char c) {
+        return Character.isDigit(c);
+    }
+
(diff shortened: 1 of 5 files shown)


edvbld:
> Just wanted to note that webrev.ksh actually does attempt to fetch the issue title from JBS:

Ah. right, now I remember. We may or may not want to do that, `git-webrev` is _really_ fast in offline mode :smiley:

PR #451, opened 2020-02-18
csr bot: use pull request cache

edvbld:
Hi all,

please review this patch that makes the CSR bot use the pull request cache. The
CSR bot now has two phases, the first phase simply checks if any updated has
gotten the "csr" label. The addition of a "csr" label can only happen by
comments being added to the PR, so the `PullRequestCache` will return these PRs
are requiring an update.

The second phase then loops through all pull requests that have the CSR label
and check if the CSR request has been resolved.

Testing:

Thanks,
Erik

## Progress

## Reviewers
 * Robin Westberg ([rwestberg](@rwestberg) - **Reviewer**) ⚠️ Review applies to 623db4bfad3eb716b870448ac334ddfe19beba9b

### DIFF
--- a/bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
+++ b/bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
@@ -73,6 +73,8 @@ class CSRBot implements Bot, WorkItem {
             log.info("Checking CSR label for " + describe(pr) + "...");
             if (pr.labels().contains(CSR_LABEL)) {
                 hasCSRLabel.add(pr.id());
+            } else {
+                hasCSRLabel.remove(pr.id());
             }
         }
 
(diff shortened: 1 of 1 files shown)


bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
```diff
@@ -58,32 +64,46 @@ private String describe(PullRequest pr) {
...
-                    log.info("No issue found in title for " + describe(pr));
-                    continue;
-                }
-                var jbsIssue = project.issue(issue.get().id());
-                if (jbsIssue.isEmpty()) {
-                    log.info("No issue found in JBS for " + describe(pr));
-                    continue;
-                }
+                hasCSRLabel.add(pr.id());
+            }
```
rwestberg: Perhaps check if the CSR_LABEL is gone but the PR is in hasCSRLabel, in case it was manually removed?

edvbld: Good idea, will fix!

edvbld: Fixed in 38429c07fbd20d66ffc6d4eda0f0ddee9c7397a0

PR #454, opened 2020-02-18
git-pr: support git pr create --draft

edvbld:
Hi all,

please review this patch adds support for `--draft` switch to `git pr create`.

Testing:
- Manual testing of `git pr create`

Thanks,
Erik

---------
### Progress

### Reviewers
 * Robin Westberg ([rwestberg](@rwestberg) - **Reviewer**) ⚠️ Review applies to 4d86006d9d114e24a52b2cca015b0889ec99ab25

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/454/head:pull/454`
`$ git checkout pull/454`


edvbld:
Just pushed a new commit that also enables the command `git pr set --no-draft` in order to support moving a PR in draft state to a non-draft state. The GitHub API for doing so was only available in the GraphQL API, so I had to add a new `RestRequest` to `GitHubHost`. It makes sense for the GraphQL `RestRequest` to be a field in `GitHubHost` since all GraphQL requests use the same [endpoint](https://api.github.com/graphql).

forge/src/main/java/org/openjdk/skara/forge/PullRequest.java
```diff
@@ -135,4 +135,5 @@
      * @return
      */
     boolean isDraft();
+    void makeNotDraft();
```
rwestberg: While symmetrical, perhaps `makeFinal` or `makeReady` would look nicer.. But if so `isDraft` should perhaps be `isFinal` as well.

edvbld: Yeah, I sort of agree. However both platforms (GitHub and GitLab) does not have a word for the "not draft" state. So to make it very clear I think I prefer the current verbose (and slightly weird) naming to be make the code clearer.

PR #510, opened 2020-03-16
Make the notifier retry logic more granular

rwestberg:
Hi all,

Please review this change that makes the notifier retry logic more granular. Instead of marking an entire notifier as either retriable or not, allow a notifier to throw a NonRetriableException if a non-retriable operation fails.

Best regards,
Robin

---------
### Progress

### Reviewers
 * Erik Helin ([ehelin](@edvbld) - **Reviewer**) ⚠️ Review applies to db69a722261f814564991b17b219f37db97b2c1e

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/510/head:pull/510`
`$ git checkout pull/510`

### DIFF
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/NonRetriableException.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/NonRetriableException.java
@@ -23,13 +23,13 @@
 package org.openjdk.skara.bots.notify;
 
 public class NonRetriableException extends Exception {
-    private final RuntimeException cause;
+    private final Throwable cause;
 
-    public NonRetriableException(RuntimeException cause) {
+    public NonRetriableException(Throwable cause) {
         this.cause = cause;
     }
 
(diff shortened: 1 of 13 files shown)


bots/notify/src/main/java/org/openjdk/skara/bots/notify/NonRetriableException.java
```diff
@@ -0,0 +1,35 @@
...
+    private final RuntimeException cause;
+
+    public NonRetriableException(RuntimeException cause) {
+        this.cause = cause;
+    }
+
+    public RuntimeException cause() {
+        return cause;
+    }
+}
```
edvbld:
```suggestion
public class NonRetriableException extends Exception {
    private final Throwable cause;

    public NonRetriableException(Throwable cause) {
        this.cause = cause;
    }

    public Throwable cause() {
        return cause;
    }
}
```

PR #517, opened 2020-03-18
316: git-webrev does not work without origin remote

edvbld:
Hi all,

please review this patch that updates `git-webrev` to work with repositories
with slightly more unusual configurations by default. `git-webrev` will now by
default work with many more repositories that for example have multiple remotes,
repositories missing a remote named "origin", etc. For the most complicated
scenarios the user can now also specify `--remote` to pick the remote that
should be used for calculating outgoing changes.

Testing:
- Manual testing of `git-webrev` on Linux x64 with some more exotic repository
  configurations

Thanks,
Erik

---------
### Progress

### Issue
 * [SKARA-316](https://bugs.openjdk.java.net/browse/SKARA-316): git-webrev does not work without origin remote

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/517/head:pull/517`
`$ git checkout pull/517`

### DIFF
--- a/cli/src/main/java/org/openjdk/skara/cli/GitWebrev.java
+++ b/cli/src/main/java/org/openjdk/skara/cli/GitWebrev.java
@@ -121,6 +121,11 @@ public class GitWebrev {
                   .describe("CR#")
                   .helptext("Include link to CR (aka bugid) in the main page")
                   .optional(),
+            Option.shortcut("")
+                  .fullname("remote")
+                  .describe("NAME")
+                  .helptext("Use remote to calculate outgoing changes")
+                  .optional(),
             Switch.shortcut("b")
                   .fullname("")
                   .helptext("Do not ignore changes in whitespace (always true)")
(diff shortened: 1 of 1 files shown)


cli/src/main/java/org/openjdk/skara/cli/GitWebrev.java
```diff
@@ -158,7 +163,7 @@ private static void generate(String[] args) throws IOException {
         }
 
         var cwd = Paths.get("").toAbsolutePath();
-        var repository = ReadOnlyRepository.get(cwd);
+        var repository = Repository.get(cwd);
```
rwestberg: What non-read-only method did you need? Can't spot it. :)

edvbld: To figure out the outgoing commits I need to find the closest merge-base with regards to all upstream branches. Since it is not certain that all upstream branches exists locally, I need to fetch the upstream branch to `FETCH_HEAD`. This is not a destructive operation for the local repository, it only mutates `FETCH_HEAD` which you shouldn't rely on anyhow staying the same.

The `fetch` method is only available on `Repository`, so I had to use `Repository` instead of `ReadOnlyRepository`.

PR #518, opened 2020-03-19
jcheck: add "problemlists" check

iignatev:
Hi all,

This is more of a request for comments than an actual PR.

The idea behind the patch is to ensure that whenever someone fixes a bug, its ID is not used in any of the problem lists. As some of you might know this is not a hypothetical issue, people forget to clean problem lists quite often, which in some cases results in poorly tested patches. Although it's true that even if we ensure that all commits are correct, there still will be ways to have "fixed" bugs in problem lists, e.g. closing as WNF, duplicate, etc, it's nevertheless believed that the majority (or the most annoying) "violations" are caused by commits.

The proposed check parses all problem lists specified by `dirs` (bar-separated list of test suites) and `pattern` (regexp pattern for files withing the `dirs`) and verifies that none of the issues fixed by a patch is listed in any of problem lists. The original / python-time [patch](http://cr.openjdk.java.net/~iignatyev//jcheck/problem-list/webrev.00/index.html) used a list glob-pattern, but unfortunately, Java SE doesn't have a good(fast) API to find all files matching a glob, and I didn't want to bring 3rd party dependency just for this use-case.

Thanks,
-- Igor

---------
### Progress

### Reviewers
 * Erik Helin ([ehelin](@edvbld) - **Reviewer**) ⚠️ Review applies to 24310cd9b0b5716bc69340cb2150b1cf3098c399

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/518/head:pull/518`
`$ git checkout pull/518`

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2020, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
@@ -215,4 +215,11 @@ class PullRequestCheckIssueVisitor implements IssueVisitor {
     public void visit(BinaryIssue issue) {
         log.fine("ignored: binary file");
     }
+
+    @Override
+    public void visit(ProblemListsIssue issue) {
+        failedChecks.add(issue.check().getClass());
+        messages.add(issue.issue() + " is used in problem lists: " + issue.files());
+        readyForReview = false;
+    }
 }
--- a/cli/src/main/java/org/openjdk/skara/cli/JCheckCLIVisitor.java
+++ b/cli/src/main/java/org/openjdk/skara/cli/JCheckCLIVisitor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2020, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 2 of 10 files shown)


edvbld:
Hi Igor,

thanks for contributing :tada:

We are in general very positive to bring in additional "jchecks" to upstream since they can easily be configured per repository. The different OpenJDK projects usually choose the checks they want enabled/disabled themselves.

I think the check you are proposing makes a lot of sense, I've run into this scenario myself in the past. I had a brief look at the code, will do a more in-depth review later, but at a first glance it looks really good :+1: One thing that stands out: could you please add one or more unit tests for your new check? You can use the unit tests for other checks as an inspiration, please see the tests in [jcheck/src/test/java/org/openjdk/skara/jcheck](https://github.com/openjdk/skara/tree/master/jcheck/src/test/java/org/openjdk/skara/jcheck).

Thanks,
Erik

iignatev:
Hi Erik,

Thanks for your prompt response. One of the reasons I haven't written any tests for this check is that I 1st wanted to get opinions on whether such a check will be accepted to skara/jcheck as my previous experience with jcheck.py showed that it's near to impossible to extend it with any new checks. Another reason is that this check reads the content of files in a repo, which AFAICT none of the existing tests do. Anyhow, I'll work on developing tests.

Cheers,
-- Igor

edvbld:
Ah ok, I get it. Skara's jcheck is very extensible, it is perfectly fine for a repository to start requiring more checks after a given commit by updating .jcheck/conf. It is also fine for a repository to only utilize a subset of the available checks (this is very common in practice).

Yes and no, the other checks do read at least .jcheck/conf :smile: What you need to be careful with is _which_ revision you read the file contents from. You probably want to use `ReadOnlyRepository::lines` instead of reading a file directly from disk (since a user might jcheck an earlier commit than the one currently being checked out). It should be fairly easy writing a unit test showcasing this scenario :+1:

Thanks,
Erik

iignatev:
Hi Erik,

thanks for the hint, I've updated the check to use `ReadOnlyRepository` to read files' content and list files in a directory. I've also added unit tests to cover different scenarios and configuration combinations.

there is one thing which I would like to discuss before we (hopefully) get it integrated -- the default value for `dirs`. as `ProblemListsCheck` will most probably be used only by "jdk-based" project, it might make sense to use values which reflects jdk layout (`test/jdk|test/langtools|test/hotspot/jtreg|test/nashorn|test/jaxp`) so most projects will be able to use the default value, on the other hand, I don't want to bake in the current directory layout, so I defaulted `dirs` to `test` which thought isn't good for any of the existing OpenJDK projects. so what do you think? 

BTW, should I update copyright years in all the touched files?

Thanks,
-- Igor

edvbld:
Great work on the patch and on the tests Igor, this is a really solid contribution! :raised_hands: 

> there is one thing which I would like to discuss before we (hopefully) get it integrated -- the default value for `dirs`. as `ProblemListsCheck` will most probably be used only by "jdk-based" project, it might make sense to use values which reflects jdk layout (`test/jdk|test/langtools|test/hotspot/jtreg|test/nashorn|test/jaxp`) so most projects will be able to use the default value, on the other hand, I don't want to bake in the current directory layout, so I defaulted `dirs` to `test` which thought isn't good for any of the existing OpenJDK projects. so what do you think?

I think defaulting to `test` is fine, I prefer that OpenJDK projects have to specify the directories containing `ProblemList.txt` files in their `.jcheck/conf`. That makes it clear which files that will actually be considered.

> BTW, should I update copyright years in all the touched files?

Yes please :pray:

iignatev:
Erik, thanks for reviewing this! I've updated copyrights in (hopefully) all the files. and yeah, I'm totally fine with the updated PR's title.

PR #519, opened 2020-03-19
304: Squash additional commits into merge commits for merge PRs

rwestberg:
Hi all,

Please review this change that squashes additional commits made on top of the first merge commit, to allow updating the PR.

Best regards,
Robin

---------
### Progress

### Issue
 * [SKARA-304](https://bugs.openjdk.java.net/browse/SKARA-304): Squash additional commits into merge commits for merge PRs

### Reviewers
 * Erik Helin ([ehelin](@edvbld) - **Reviewer**) ⚠️ Review applies to 978e08f79023e6cc1bd07171fdd2e321c11efddc

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/519/head:pull/519`
`$ git checkout pull/519`

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -164,8 +164,15 @@ class CheckRun {
             if (commits.size() < 2) {
                 ret.add("A Merge PR must contain at least two commits that are not already present in the target.");
             } else {
-                if (!commits.get(0).isMerge()) {
-                    ret.add("The top commit must be a merge commit.");
+                // Find the last merge commit - the very last commit is not eligible (as the merge needs a parent)
+                int mergeCommitIndex = commits.size();
+                for (int i = 0; i < commits.size() - 1; ++i) {
+                    if (commits.get(i).isMerge()) {
+                        mergeCommitIndex = i;
+                    }
+                }
+                if (mergeCommitIndex >= commits.size() - 1) {
+                    ret.add("A Merge PR must contain a merge commit.");
                 }
 
                 var source = mergeSource();
@@ -176,7 +183,7 @@ class CheckRun {
                         );
                         try {
                             var sourceHash = prInstance.localRepo().fetch(mergeSourceRepo.url(), source.get().branchName);
-                            if (!prInstance.localRepo().isAncestor(commits.get(1).hash(), sourceHash)) {
+                            if (!prInstance.localRepo().isAncestor(commits.get(mergeCommitIndex + 1).hash(), sourceHash)) {
                                 ret.add("The merge contains commits that are not ancestors of the source");
                             }
                         } catch (IOException e) {
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -172,7 +172,7 @@ class CheckRun {
                     }
                 }
                 if (mergeCommitIndex >= commits.size() - 1) {
-                    ret.add("A Merge PR must contain a merge commit.");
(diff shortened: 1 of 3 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -164,8 +164,15 @@ private MergeSource(String repositoryName, String branchName) {
...
-                if (!commits.get(0).isMerge()) {
-                    ret.add("The top commit must be a merge commit.");
+                // Find the last merge commit - the very last commit is not eligible (as the merge needs a parent)
+                int mergeCommitIndex = commits.size();
+                for (int i = 0; i < commits.size() - 1; ++i) {
+                    if (commits.get(i).isMerge()) {
+                        mergeCommitIndex = i;
+                    }
+                }
+                if (mergeCommitIndex >= commits.size() - 1) {
```
edvbld: What if you do a "Merge PR" with a single trivial merge just to merge to branches (that have identical commits). Too narrow of a use case to think about?

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -164,8 +164,15 @@ private MergeSource(String repositoryName, String branchName) {
...
+                    ret.add("A Merge PR must contain a merge commit.");
```
edvbld: Related to my previous comment, should the message be "A Merge PR must contain at least one merge commit and one non-merge commit"?

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestInstance.java
```diff
@@ -133,16 +133,31 @@ private Hash commitSquashed(List<Review> activeReviews, Namespace namespace, Str
...
         var contributor = namespace.get(pr.author().id());
         if (contributor == null) {
             throw new RuntimeException("Merges can only be performed by Committers");
         }
 
         var author = new Author(contributor.fullName().orElseThrow(), contributor.username() + "@" + censusDomain);
-
         var commitMessage = commitMessage(activeReviews, namespace, true);
+
+        localRepo.checkout(commits.get(mergeCommitIndex).hash(), true);
```
edvbld: Check that `mergeCommitIndex < commits.size()` before and otherwise throw an exception?

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
rwestberg: Not sure if that is something we would need to support, not even sure how to create such a merge commit. :)

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
rwestberg: Sure, can update that.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestInstance.java
rwestberg: That should never happen (tm), the actual merge depends on jcheck passing first so we can assume that the PR is well-formed.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
edvbld: Ok, then forget about this comment :smile:

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestInstance.java
edvbld: Ok, sounds good!

PR #525, opened 2020-03-23
jcheck: add symlink check

edvbld:
Hi all,

please review this patch that adds a check for symbolic links to JCheck.

Testing:
- Added two new unit tests

Thanks,
Erik

---------
### Progress

### Reviewers
 * Robin Westberg ([rwestberg](@rwestberg) - **Reviewer**) ⚠️ Review applies to 69b8feddbc80898d6327b1aa3aeba9140ad6b8bd

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/525/head:pull/525`
`$ git checkout pull/525`

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
@@ -208,7 +208,7 @@ class PullRequestCheckIssueVisitor implements IssueVisitor {
 
     @Override
     public void visit(SymlinkIssue issue) {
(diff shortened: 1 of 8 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
```diff
@@ -206,6 +206,13 @@ public void visit(ExecutableIssue issue) {
         readyForReview = false;
     }
 
+    @Override
+    public void visit(SymlinkIssue issue) {
+        messages.add(String.format("Symbolic linkes are not allowed (file: %s)", issue.path()));
```
rwestberg:
```suggestion
        messages.add(String.format("Symbolic links are not allowed (file: %s)", issue.path()));
```

PR #536, opened 2020-03-24
mirror-bot: support mirroring of selected branches

edvbld:
Hi all,

please review this patch that enables the mirror bot to mirror only selected
branches (instead of all as is done today). This is useful for projects who
might have a `master` that is a replica of `jdk:master`, but don't want tags
(nor potentially other branches) from the
[jdk](https://git.openjdk.java.net/jdk) repository.

Testing:
- Added a new unit test

Thanks,
Erik

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/536/head:pull/536`
`$ git checkout pull/536`


kevinrushforth:
Nice. Would it be possible to have an option to also sync reachable tags, using `git push --follow-tags` in the case where only a single branch is mirrored?

edvbld:
@kevinrushforth yes, I thought about adding that as the next feature :smiley:

PR #553, opened 2020-03-31
343: Use `ssh -G` for parsing ssh configuration files

edvbld:
Hi all,

please review this patch that uses `ssh -G` instead of relying on my (somewhat
buggy) `ssh_config` parser. Resolving hosts defined in a `ssh_config` file is
actually fairly complex and my parser only handled a subset of scenarios. One
drawback is that we now rely on `ssh` being present on `PATH`, but that must be
the case if a user has a SSH path for a remote.

Testing:
- Manual testing using `git-webrev` with various paths

Thanks,
Erik

---------
### Progress

### Issue
 * [SKARA-343](https://bugs.openjdk.java.net/browse/SKARA-343): Use `ssh -G` for parsing ssh configuration files

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/553/head:pull/553`
`$ git checkout pull/553`


pliden:
Thanks for fixing! :)

PR #555, opened 2020-03-31
344: PR should warn when source branch is an upstream branch

edvbld:
Hi all,

please review this patch that warns contributors that creates a poll request
from a branch that is already present in the upstream repository. This is rarely
what you want to do and therefore we should issue a warning.

Testing:

Thanks,
Erik

---------
### Progress

### Issue
 * [SKARA-344](https://bugs.openjdk.java.net/browse/SKARA-344): PR should warn when source branch is an upstream branch

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/555/head:pull/555`
`$ git checkout pull/555`


edvbld:
Thanks @rwestberg for reviewing! Based on a offline discussion I changed `HostedBranch` from being an `interface` to a `class`.

PR #585, opened 2020-04-16
cli: add git-proxy tool

edvbld:
Hi all,

please review this patch that adds a small new tool: `git-proxy`. `git-proxy`
can be used to set up the `HTTPS_PROXY` and `HTTP_PROXY` environment
variables based on either the `http.proxy` configuration setting or the `-c
http.proxy=` argument. Finally `git-proxy` will also set up the
`GIT_SSH_COMMAND` environment variable to use the `CONNECT` method for tunneling
SSH over an HTTP proxy.

Testing:
- manual testing of `git-proxy`

Thanks,
Erik

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/585/head:pull/585`
`$ git checkout pull/585`

### DIFF
--- a/cli/src/main/java/org/openjdk/skara/cli/GitProxy.java
+++ b/cli/src/main/java/org/openjdk/skara/cli/GitProxy.java
@@ -1,0 +1,134 @@
+/*
+ * Copyright (c) 2019, 2020, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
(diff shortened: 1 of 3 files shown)


cli/src/main/java/org/openjdk/skara/cli/GitProxy.java
```diff
@@ -0,0 +1,134 @@
...
+            if (proxy == null) {
+                proxy = System.getenv(key.toUpperCase());
+            }
+            if (proxy != null) {
+                return proxy;
+            }
+        }
+        return "";
+    }
+
```
pliden, line 58: It looks like both these methods are unused?

edvbld: Oops, yes :smile: They are remnants from a prototype, before I realized I could just utilize the `HttpProxy` class.

PR #636, opened 2020-06-03
json patches introduced to the OpenJDK copy in JDK-8246435

magicus:
This PR contains the proposed changes to the json system made in JDK-8246435. If the same changes are made in Skara, tthe two code bases are kept from diverging.

---------
### Progress

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/636/head:pull/636`
`$ git checkout pull/636`

### DIFF
--- a/json/src/main/java/org/openjdk/skara/json/JSONArray.java
+++ b/json/src/main/java/org/openjdk/skara/json/JSONArray.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2020, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -36,6 +36,24 @@ public class JSONArray implements JSONValue, Iterable<JSONValue> {
         this.values = new ArrayList<JSONValue>(array.length);
         for (var v : array) {
             values.add(v);
+        }
+    }
+
+    private void append(JSONValue value) {
+        if (value instanceof JSONArray) {
+            for (var v : value.asArray()) {
+                append(v);
+            }
+        } else {
+            this.values.add(value);
+        }
+    }
+
+    public JSONArray(JSONValue value, JSONValue... values) {
+        this.values = new ArrayList<JSONValue>(values.length + 1);
+        append(value);
+        for (var v : values) {
+            append(v);
         }
     }
 
--- a/json/src/main/java/org/openjdk/skara/json/JSONBoolean.java
+++ b/json/src/main/java/org/openjdk/skara/json/JSONBoolean.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2020, Oracle and/or its affiliates. All rights reserved.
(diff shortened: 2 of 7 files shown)


json/src/main/java/org/openjdk/skara/json/JSONArray.java
```diff
@@ -39,6 +39,24 @@ public JSONArray(JSONValue[] array) {
         }
     }
 
+    private void append(JSONValue value) {
```
edvbld, line 42: Could you rename this to `concat` and explicitly take a `JSONArray` instead as parameter? I'm fine with having `append`, but `append` should be just `append` (that is, append a single item to the array).

json/src/main/java/org/openjdk/skara/json/JSONArray.java
```diff
@@ -39,6 +39,24 @@ public JSONArray(JSONValue[] array) {
...
+        if (value instanceof JSONArray) {
+            for (var v : value.asArray()) {
+                append(v);
+            }
+        } else {
+            this.values.add(value);
+        }
+    }
+
+    public JSONArray(JSONValue value, JSONValue... values) {
```
edvbld, line 52: Related to the comment about `concat` vs `append`, I would prefer that the constructor does `append` and not `concat`. One could imagine a `static` version of `concat` as well, e.g. `public static JSONArray concat(JSONArray... arrays)`.

json/src/main/java/org/openjdk/skara/json/JSONArray.java
magicus: To me there's no deeper semantic difference between "append" and "concat". To be precise, this method do a flattening append/concat, and should probably be named something like addAndFlatten to be completely unambiguous. But that kind of naming did not seem to fit in with the rest of the design here.

If you insist that "concat" means "flatten while adding" but "append" does not, sure, I can rename it.

json/src/main/java/org/openjdk/skara/json/JSONArray.java
magicus: I don't care about a non-flattening constructor. You can add one if you need one. But I can change it to a static factory method instead, if you want. It's probably clearer, given that we can agree on a suitable name. (I'll accept "concat" if you insist, but I'm not persuaded that it's really better...)

json/src/main/java/org/openjdk/skara/json/JSONArray.java
magicus: Ah, sorry, now I see what you mean. You want a concat(JSONArray a) instead. Well, it might be good to have, but that does not really solve my problem, does it? :-)

edvbld: Well, if you have `append(JSONValue v)` and `concat(JSONArray a)` then code using such an API can easily have the code:

```java
void appendOrConcat(JSONArray a, JSONValue v) {
    if (v.isArray()) {
        a.concat(v.asArray());
    } else {
        a.append(v);
    }
}
```

I'm fine with having `appendOrConcat` in `JSONArray` as well, but then I want it named `appendOrConcat`, not `append`. I agree that `append` is shorter, but it is too misleading of a name IMHO for a method that really does `appendOrConcat`.

magicus: I still don't get it. Your example code would not recursively inline arrays; it will only work for a single nested layer. I can't see how that would be helpful to amything but the most specific situations. What is it you want to achieve?

magicus:
Yeah, yeah... Stop whining, bot.

PR #639, opened 2020-06-04
forge: support commit comments

edvbld:
Hi all,

please review this patch that adds the possibility to add a comment on a particular commit to `HostedRepository` via the method `addCommitComment`. To start out with I've only added support for adding a general comment about a whole commit, not a comment for a particular line in a commit. 

The comments for a commit can also be listed with the method `commitComments`.

Testing:
- Not that much, later patches will build upon the APIs introduced in this patch

Thanks,
Erik

---------
### Progress

### Reviewers
 * Robin Westberg ([rwestberg](@rwestberg) - **Reviewer**) ⚠️ Review applies to b5a30420b23c5afb010d165fa99cf417c2d15ab0

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/639/head:pull/639`
`$ git checkout pull/639`

### DIFF
--- a/forge/src/main/java/org/openjdk/skara/forge/CommitComment.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/CommitComment.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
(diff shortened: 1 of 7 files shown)


forge/src/main/java/org/openjdk/skara/forge/CommitComment.java
```diff
@@ -0,0 +1,103 @@
+/*
+ * Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
```
rwestberg, line 2: 2020!

forge/src/main/java/org/openjdk/skara/forge/CommitComment.java
```diff
@@ -0,0 +1,103 @@
...
+     * Returns the relative path of the file.
+     */
+    public Optional<Path> path() {
+        return Optional.ofNullable(path);
+    }
+
+    /**
+     * Returns the line index in the diff.
+     */
+    public Optional<Integer> position() {
```
rwestberg, line 65: Perhaps drop this one? The other API's only expose a line number

forge/src/main/java/org/openjdk/skara/forge/CommitComment.java
edvbld: Thanks, will fix!

forge/src/main/java/org/openjdk/skara/forge/CommitComment.java
edvbld: Ok, yeah, sounds good :+1:

PR #641, opened 2020-06-05
notify: store resulting commit in PullRequestState

edvbld:
Hi all,

please review this patch that the resulting (integrated) commit from a PR in the `PullRequestState`. This data will be used in future patches for more easier handling of notifications. Note that all data in the `Storage` might not have the `"commit"` field. If that is the case, then the code will lazily populate the `"commit"` field.

Testing:

Thanks,
Erik

---------
### Progress

### Reviewers
 * Robin Westberg ([rwestberg](@rwestberg) - **Reviewer**) ⚠️ Review applies to 4f70c74e07d13cf56ea33af3f9855fab0eb2c9d4

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/641/head:pull/641`
`$ git checkout pull/641`

### DIFF
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/PullRequestWorkItem.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/PullRequestWorkItem.java
@@ -24,8 +24,10 @@ package org.openjdk.skara.bots.notify;
 
 import org.openjdk.skara.bot.WorkItem;
 import org.openjdk.skara.forge.PullRequest;
+import org.openjdk.skara.host.HostUser;
 import org.openjdk.skara.json.*;
 import org.openjdk.skara.storage.StorageBuilder;
+import org.openjdk.skara.vcs.Hash;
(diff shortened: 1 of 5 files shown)


bots/notify/src/main/java/org/openjdk/skara/bots/notify/PullRequestWorkItem.java
```diff
@@ -130,31 +176,32 @@ public void run(Path scratchPath) {
                 .materialize(historyPath);
 
         var issues = parseIssues();
-        var prIssues = new PullRequestState(pr, issues);
-        var current = storage.current();
-        if (current.contains(prIssues)) {
+        var commit = resultingCommitHashFor(pr, pr.repository().forge().currentUser());
```
rwestberg, line 179: The notifier bot may not be running with the same credentials as the commit bot, so this is possibly the wrong user to look for. The other bots (mlbridge, pr) takes a username from the configuration to use for matching comment poster ids.

PR #648, opened 2020-06-10
Allow a WorkItem to return follow-up items

rwestberg:
Hi all,

Please review this change that allows a WorkItem to return follow-up items to schedule after itself has run. Use this to run another CheckWorkItem after a CommandWorkItem successfully has executed commands.

Best regards,
Robin

---------
### Progress

### Reviewers
 * Erik Helin ([ehelin](@edvbld) - **Reviewer**) ⚠️ Review applies to d76eda50fd10c3b2a0145ce7ed767c6a6b6e10bc

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/648/head:pull/648`
`$ git checkout pull/648`

### DIFF
--- a/bot/src/main/java/org/openjdk/skara/bot/WorkItem.java
+++ b/bot/src/main/java/org/openjdk/skara/bot/WorkItem.java
@@ -34,8 +34,10 @@ public interface WorkItem {
     boolean concurrentWith(WorkItem other);
 
     /**
-     * Execute the appropriate tasks with the provided scratch folder.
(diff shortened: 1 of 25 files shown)


bot/src/main/java/org/openjdk/skara/bot/WorkItem.java
```diff
@@ -37,7 +37,7 @@
      * Execute the appropriate tasks with the provided scratch folder.
      * @param scratchPath
```
edvbld, line 38: Maybe want to update the doc?

PR #660, opened 2020-06-12
Update the isReady condition for the notifier

rwestberg:
Hi all,

Please review this change that updates the isReady condition for the notifier. Instead of having this as a configurable option, just use the well-known values.

Best regards,
Robin

---------
### Progress

### Reviewers
 * Erik Helin ([ehelin](@edvbld) - **Reviewer**) ⚠️ Review applies to 8db8ec4273118533d2618191701552e9883dcc48

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/660/head:pull/660`
`$ git checkout pull/660`

### DIFF
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/NotifyBot.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/NotifyBot.java
@@ -66,7 +66,7 @@ public class NotifyBot implements Bot, Emitter {
         return new NotifyBotBuilder();
     }
 
-    private boolean isReady(PullRequest pr) {
+    private boolean isOfInterest(PullRequest pr) {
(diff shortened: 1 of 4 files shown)


bots/notify/src/main/java/org/openjdk/skara/bots/notify/NotifyBot.java
```diff
@@ -70,11 +68,9 @@ public static NotifyBotBuilder newBuilder() {
 
     private boolean isReady(PullRequest pr) {
```
edvbld, line 69: Perhaps rename `isOfInterest`?

rwestberg: Sure, sounds reasonable

PR #665, opened 2020-06-16
411: Warn when a PR issue is not of a primary type

edvbld:
Hi all,

please review this patch that makes it possible the issue types that should be allowed to be referenced from a PR for a repository.

Testing:
- [x] Added a new unit test

Thanks,
Erik

---------
### Progress

### Issues
 * [SKARA-411](https://bugs.openjdk.java.net/browse/SKARA-411): Warn when a PR issue is not of a primary type
 * [SKARA-415](https://bugs.openjdk.java.net/browse/SKARA-415): A repository should be configurable to only accept backports

### Reviewers
 * Robin Westberg ([rwestberg](@rwestberg) - **Reviewer**) ⚠️ Review applies to 580d6cde3d82d851f67a5a91b9ab82c98bef7fb5

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/665/head:pull/665`
`$ git checkout pull/665`

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestBotFactory.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestBotFactory.java
@@ -118,6 +118,13 @@ public class PullRequestBotFactory implements BotFactory {
             if (repo.value().contains("ignorestale")) {
                 botBuilder.ignoreStaleReviews(repo.value().get("ignorestale").asBoolean());
             }
+            if (repo.value().contains("issuetypes")) {
+                var types = repo.value().get("issuetypes").asArray()
+                                                          .stream()
+                                                          .map(JSONValue::asString)
+                                                          .collect(Collectors.toSet());
+                botBuilder.allowedIssueTypes(types);
+            }
             if (repo.value().contains("targetbranches")) {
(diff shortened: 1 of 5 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestBotFactory.java
```diff
@@ -118,6 +118,13 @@ public String name() {
             if (repo.value().contains("ignorestale")) {
                 botBuilder.ignoreStaleReviews(repo.value().get("ignorestale").asBoolean());
             }
+            if (repo.value().contains("issuetypes")) {
+                var types = repo.value().get("issuetypes").asArray()
```
rwestberg, line 122: You just do .stream directly here, right? At least I tend to do that.. :)

edvbld: Right, yes, will fix!

kevinrushforth:
> 415: A repository should be configurable to only accept backports

Actually, it's quite the opposite: we want to ensure that a backport record is *never* used as the bug ID. This PR makes the list of issue types configurable, which is good.

PR #674, opened 2020-06-18
webrev: add logging

edvbld:
Hi all,

please review this small patch that add some additional logging to `Webrev` and dereferences `Optional`s a bit more carefully in `ModifiedFileView`.

Testing:

Thanks,
Erik

---------
### Progress

### Reviewers
 * Jorn Vernee ([jvernee](@JornVernee) - **Reviewer**) ⚠️ Review applies to df7bb77cdf08cb4c2f23999455ebcb0b0dda5636

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/674/head:pull/674`
`$ git checkout pull/674`

### DIFF
--- a/webrev/src/main/java/org/openjdk/skara/webrev/ModifiedFileView.java
+++ b/webrev/src/main/java/org/openjdk/skara/webrev/ModifiedFileView.java
@@ -50,20 +50,25 @@ class ModifiedFileView implements FileView {
         this.formatter = formatter;
         if (patch.isTextual()) {
             binaryContent = null;
-            oldContent = repo.lines(patch.source().path().get(), base).orElseThrow(() -> {
-                throw new IllegalArgumentException("Could not get content for file " +
-                                                   patch.source().path().get() +
-                                                   " at revision " + base);
-            });
+            var sourcePath = patch.source().path().orElseThrow(() ->
+                new IllegalArgumentException("Could not get source path for file with hash " +
+                                                   patch.source().hash() + " with target path" +
+                                                   patch.target().path().get())
+            );
+
+            oldContent = repo.lines(sourcePath, base).orElseThrow(() ->
+                new IllegalArgumentException("Could not get content for file " +
+                                                   sourcePath + " at revision " + base)
+            );
             if (head == null) {
                 var path = repo.root().resolve(patch.target().path().get());
                 if (patch.target().type().get().isVCSLink()) {
                     var tip = repo.head();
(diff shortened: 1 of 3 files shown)


webrev/src/main/java/org/openjdk/skara/webrev/ModifiedFileView.java
```diff
@@ -50,20 +50,25 @@ public ModifiedFileView(ReadOnlyRepository repo, Hash base, Hash head, List<Comm
...
             binaryContent = null;
-            oldContent = repo.lines(patch.source().path().get(), base).orElseThrow(() -> {
-                throw new IllegalArgumentException("Could not get content for file " +
-                                                   patch.source().path().get() +
-                                                   " at revision " + base);
-            });
+            var sourcePath = patch.source().path().orElseThrow(() ->
+                new IllegalArgumentException("Could not get source path for file with hash " +
+                                                   patch.source().hash() + " with target path" +
+                                                   patch.target().path().get())
```
JornVernee, line 56: Is this `get()` safe though? Maybe you want to extract the target path into a variable as well, since the `patch.target().path().get()` is used a few times below as well? (It would reduce the number of `get()` calls`, so less scrutiny needed).

webrev/src/main/java/org/openjdk/skara/webrev/Webrev.java
```diff
@@ -198,6 +202,11 @@ private void generate(Diff diff, Hash tailEnd, Hash head) throws IOException {
             }
 
             var headHash = head == null ? repository.head() : head;
+            var filesDesc = files.isEmpty() ? "" :
+                            " for files " +
+                            String.join(", ", files.stream().map(Path::toString).collect(Collectors.toList()));
```
JornVernee, line 207: You could also use [`Collectors.joining`](https://docs.oracle.com/en/java/javase/14/docs/api/java.base/java/util/stream/Collectors.html#joining(java.lang.CharSequence)) here
```suggestion
                            files.stream().map(Path::toString).collect(Collectors.joining(", "));
```

webrev/src/main/java/org/openjdk/skara/webrev/ModifiedFileView.java
edvbld: Yeah, `path.target().path().get()` _should_ be safe. If the patch does not have `source` path nor have a `target` path, then something is seriously wrong :smile:

PR #707, opened 2020-07-29
SKARA-461: Do a git version check

JornVernee:
Hi,

This PR adds a git version check to the `git skara` CLI. This will print a warning if the git version is not a known working version. Since outdated versions can lead to IOExceptions due to e.g. unsupported git CLI options, hopefully the warning message will help to inform users about their use of an unsupported git version, and a potential way to solve the problem i.e. upgrading to a more recent version.

Currently at least git 2.22 is required since we use the `--combined-all-paths` option of git log (see GitCommits), which was added in 2.22.

Thanks,
Jorn

---------
### Progress

### Issue
 * [SKARA-461](https://bugs.openjdk.java.net/browse/SKARA-461): Do a git version check

### Reviewers
 * Kevin Rushforth ([kcr](@kevinrushforth) - no project role)

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/707/head:pull/707`
`$ git checkout pull/707`

### DIFF
--- a/cli/src/main/java/org/openjdk/skara/cli/GitSkara.java
+++ b/cli/src/main/java/org/openjdk/skara/cli/GitSkara.java
@@ -24,6 +24,7 @@ package org.openjdk.skara.cli;
 
 import org.openjdk.skara.args.Main;
 import org.openjdk.skara.vcs.Repository;
+import org.openjdk.skara.vcs.git.GitVersion;
 import org.openjdk.skara.vcs.openjdk.CommitMessageParsers;
 import org.openjdk.skara.version.Version;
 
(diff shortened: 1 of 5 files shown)


kevinrushforth:
How hard would it be to actually support older versions? Requiring 2.22 might be a barrier for some users. I just checked and 2 of the 4 machines I use, including my main development platform, have versions older than that. I'm not a heavy user of the Skara CLI tools, but am concerned about those who are.

kevinrushforth:
Btw, I think this PR is fine unless / until the Skara CLI tools support older versions.

JornVernee:
@kevinrushforth Thanks for taking a look.

My intent was not to say the CLI tools should require 2.22 or some other version, but the current list of known working versions is adjusted to the fact that the current implementation requires version 2.22 to run successfully (otherwise e.g. https://bugs.openjdk.java.net/browse/SKARA-454 can occur).

I think the intent was to support older versions as well, since a lot of package managers provide older versions of git, but currently using a version older than 2.22 causes problems. I've added the other versions to the code as well, but commented them out for now.

The required version should probably be adjusted downwards once lesser versions have been confirmed to work.

PR #736, opened 2020-08-24
vcs: improve git->hg converter

edvbld:
Hi all,

please review this patch that significantly improves the `GitToHgConverter`. The new converter handles all OpenJDK Git repositories and also handles additional Git tags. I also improved the testing by quite a bit, the verification is much more strict and I added a bunch of more tests. I also added some `@Disabled` tests (since they require an internet connection), but they are still useful for running locally when working on the converter.

Testing:
- [x] Added a number of new unit tests

Thanks,
Erik

---------
### Progress

### Reviewers
 * Robin Westberg ([rwestberg](@rwestberg) - **Reviewer**) ⚠️ Review applies to b957495aae1fb6989f568b000b4b7bc96226e735

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/736/head:pull/736`
`$ git checkout pull/736`

### DIFF
--- a/vcs/src/test/java/org/openjdk/skara/vcs/openjdk/converter/GitToHgConverterTests.java
+++ b/vcs/src/test/java/org/openjdk/skara/vcs/openjdk/converter/GitToHgConverterTests.java
@@ -25,86 +25,86 @@ package org.openjdk.skara.vcs.openjdk.converter;
 import org.openjdk.skara.test.TemporaryDirectory;
 import org.openjdk.skara.vcs.*;
 import org.openjdk.skara.vcs.openjdk.convert.GitToHgConverter;
+import org.openjdk.skara.vcs.openjdk.convert.Mark;
 
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Disabled;
 
 import java.io.IOException;
 import java.nio.file.Files;
+import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
 import java.util.*;
+import java.net.URI;
+import java.util.stream.Collectors;
+import java.time.ZonedDateTime;
 
 import static org.junit.jupiter.api.Assertions.*;
 
 class GitToHgConverterTests {
-    void assertCommitEquals(Commit gitCommit, Commit hgCommit) {
+    void assertCommitEquals(ReadOnlyRepository gitRepo, Commit gitCommit, ReadOnlyRepository hgRepo, Commit hgCommit) throws IOException {
+        System.out.println("git commit: " + gitCommit.hash() + ", hg commit: " + hgCommit.hash());
         assertEquals(gitCommit.authored(), hgCommit.authored());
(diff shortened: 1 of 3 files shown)


vcs/src/test/java/org/openjdk/skara/vcs/openjdk/converter/GitToHgConverterTests.java
```diff
@@ -25,86 +25,86 @@
...
+import java.net.URI;
+import java.util.stream.Collectors;
+import java.time.ZonedDateTime;
 
 import static org.junit.jupiter.api.Assertions.*;
 
 class GitToHgConverterTests {
-    void assertCommitEquals(Commit gitCommit, Commit hgCommit) {
+    void assertCommitEquals(ReadOnlyRepository gitRepo, Commit gitCommit, ReadOnlyRepository hgRepo, Commit hgCommit) throws IOException {
+        System.out.println("git commit: " + gitCommit.hash() + ", hg commit: " + hgCommit.hash());
```
rwestberg, line 46: Debugging leftover?

vcs/src/test/java/org/openjdk/skara/vcs/openjdk/converter/GitToHgConverterTests.java
```diff
@@ -400,16 +400,121 @@ void convertMergeCommitWithP1Diff() throws IOException {
...
+
+            var annotated = hgRepo.annotate(new Tag("1.0"));
+            assertTrue(annotated.isPresent());
+            assertEquals("foo", annotated.get().author().name());
+            assertEquals(tagDate, annotated.get().date());
+            assertEquals("Added tag 1.0", annotated.get().message());
+        }
+    }
+
+    @Disabled("Depends on internet connection")
```
rwestberg, line 473: There are similar tests for the bridge bot that are skipped if there is no connection:

https://github.com/openjdk/skara/blob/269027247e5c5778560d3031476dfd0ece4e57d3/bots/hgbridge/src/test/java/org/openjdk/skara/bots/hgbridge/BridgeBotTests.java#L131

but not sure if something similar would be appropriate here as well?

rwestberg (approved):
Looks good, just some minor comments!

vcs/src/test/java/org/openjdk/skara/vcs/openjdk/converter/GitToHgConverterTests.java
edvbld: Thanks, removed

vcs/src/test/java/org/openjdk/skara/vcs/openjdk/converter/GitToHgConverterTests.java
edvbld: Good idea, but these takes also takes a bit too long to run in addition to requiring an internet connection, so I would prefer to leave them out for now. I might look into conditionally enable them depending on an environment variable so that we can have them running in CI.

PR #737, opened 2020-08-24
cli: skara debug import-git should always save marks

edvbld:
Hi all,

please review this patch that makes `git skara debug import-git` always save the marks that were created as part of a conversion, even though the conversion might have failed. This allows for incremental git imports.

Testing:

Thanks,
Erik


rwestberg (approved):
Looks good!

PR #748, opened 2020-08-27
Updates to config/mailinglist/rules/jdk.json related to compiler-dev and kulla-dev.

lahodaj:
…and kulla-dev.

I tried to keep the sensible stuff, sending to multiple lists where in a real doubt. Some other javac/javadoc maintainers may disagree with some of the choices, though.

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/748/head:pull/748`
`$ git checkout pull/748`


config/mailinglist/rules/jdk.json
```diff
@@ -113,83 +113,24 @@
...
-            "src/hotspot/share/classfile/classFileParser.hpp",
-            "src/hotspot/share/code/dependencies.cpp",
-            "src/hotspot/share/interpreter/bytecode.cpp",
-            "src/hotspot/share/interpreter/bytecodeTracer.cpp",
-            "src/hotspot/share/interpreter/linkResolver.hpp",
-            "src/hotspot/share/interpreter/templateTable.cpp",
-            "src/hotspot/share/interpreter/templateTable.hpp",
-            "src/hotspot/share/runtime/reflection.hpp",
-            "src/java.base/share/classes/java/lang/",
             "src/java.base/share/classes/jdk/internal/PreviewFeature.java",
```
lahodaj, line 123: I think this shows the question on maintenance from the jdk-dev list - the (internal) PreviewFeature may be moved to a different (internal) place sometime in the foreseeable future. Who will be updating this list?

rwestberg: When in doubt, I think it's better to err on the side of caution, so perhaps it should not be included in the list at all. But will expand on this in the jdk-dev thread.

rwestberg (approved):
Thanks for the update!

PR #754, opened 2020-08-30
webrev: not all commits changes files

edvbld:
Hi all,

please review this small patch that updates webrev's JSON generator to handle commits that don't change any files. This can happen for example with trivial merges (since we use the `-c` flag to `git log`).

Thanks,
Erik

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/754/head:pull/754`
`$ git checkout pull/754`


rwestberg (approved):
Looks good!

PR #760, opened 2020-09-01
storage: support nested paths in file storage

edvbld:
Hi all,

please review this path that makes `FileStorage` with files in nested directories.

Testing:
- Manual testing on Linux x64

Thanks,
Erik

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/760/head:pull/760`
`$ git checkout pull/760`

### DIFF
--- a/storage/src/main/java/org/openjdk/skara/storage/FileStorage.java
+++ b/storage/src/main/java/org/openjdk/skara/storage/FileStorage.java
@@ -64,6 +64,10 @@ class FileStorage<T> implements Storage<T> {
             return;
         }
         try {
+            Files.createDirectories(file.getParent());
+            if (!Files.exists(file)) {
+                Files.createFile(file);
+            }
             Files.writeString(file, updated, StandardCharsets.UTF_8);
         } catch (IOException e) {
             throw new UncheckedIOException(e);


storage/src/main/java/org/openjdk/skara/storage/FileStorage.java
```diff
@@ -64,6 +64,10 @@ public void put(Collection<T> items) {
             return;
         }
         try {
+            Files.createDirectories(file.getParent());
+            if (!Files.exists(file)) {
```
rwestberg, line 68: Doesn't hurt, but Files.writeString should create it if it doesn't exist as well I think.

rwestberg (approved):
Looks good!

storage/src/main/java/org/openjdk/skara/storage/FileStorage.java
edvbld: It didn't for nested paths, so the `Files.createDirectories` is definitely needed. The `if (!Files.exists)` could potentially be left out, but otoh it doesn't hurt to be explicit here.

PR #764, opened 2020-09-03
webrev: do not use follow for ranges with merges

edvbld:
Hi all,

please review this patch that makes JSON webrev use plain `git log` if the range of commits include a merge commits. This is needed because `git log -c --follow` can be extremely slow for commit ranges with a lot of merge commits.

Testing:
- `make test` on Linux x64
- Manual verification of performance

Thanks,
Erik

---------
### Progress

### Reviewers
 * [Robin Westberg](https://openjdk.java.net/census#rwestberg] (@rwestberg - **Reviewer**)

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/764/head:pull/764`
`$ git checkout pull/764`


rwestberg (approved):
Looks good!

PR #803, opened 2020-09-07
checkout: check status before converting

edvbld:
Hi all,

please review this patch that makes the checkout bot checks if the `HEAD` commit on the branch in the `from` repository already has been converted. If it has, then there is no need to run the converter.

Testing:

Thanks,
Erik

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/803/head:pull/803`
`$ git checkout pull/803`


rwestberg (approved):
Looks good!

PR #805, opened 2020-09-08
jcheck: remove unused code

edvbld:
Hi all,

please review this patch that removes some very old code in `JCheck` that has never been used and just clutters the `JCheck` API. We have other ways of ensuring whether a certain commit should or should not be allowed to be pushed, the code that this patch removes is not used.

This is a fully backwards compatible change with both older and newer .jcheck/conf files.

Testing:
- [x] `make images` passes on Linux x64

Thanks,
Erik

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/805/head:pull/805`
`$ git checkout pull/805`


rwestberg (approved):
Looks good!

PR #806, opened 2020-09-08
vcs: do not hard-code VCS default branch names

edvbld:
Hi all,

please review this patch that ensures that we don't hard-code the name of the default branches for various version control system throughout the Skara source code. Skara has from the beginning been very strict about abstracting away VCS implementation details, but the name of the default branches has managed to find its way into multiple corners of the source code.

Testing:
- [x] `make images` passes on Linux x64

Thanks,
Erik

---------
### Progress

### Reviewers
 * [Robin Westberg](https://openjdk.java.net/census#rwestberg) (@rwestberg - **Reviewer**) ⚠️ Review applies to 6b2063aa75e52f89bf3a96d56d19222cec505de4

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/806/head:pull/806`
`$ git checkout pull/806`


rwestberg (approved):
Looks good, thanks for fixing!

PR #820, opened 2020-09-14
673: PR author should be able to use /reviewers command

edvbld:
Hi all,

please review this patch that allows pull request author's to issue the `/reviewers` command (sometimes the PR author knows ahead of time that it is a complicated to PR to review).

Testing:
- [x] Added one additional unit test

Thanks,
Erik

---------
### Progress

### Issue
 * [SKARA-673](https://bugs.openjdk.java.net/browse/SKARA-673): PR author should be able to use /reviewers command

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/820/head:pull/820`
`$ git checkout pull/820`

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CommandInvocation.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CommandInvocation.java
@@ -11,15 +11,13 @@ class CommandInvocation {
     private final CommandHandler handler;
     private final String name;
     private final String args;
-    private final Comment comment;
 
-    CommandInvocation(String id, HostUser user, CommandHandler handler, String name, String args, Comment comment) {
+    CommandInvocation(String id, HostUser user, CommandHandler handler, String name, String args) {
         this.id = id;
         this.user = user;
         this.handler = handler;
         this.name = name;
         this.args = args != null ? args.strip() : "";
-        this.comment = comment;
     }
 
     String id() {
@@ -40,17 +38,5 @@ class CommandInvocation {
 
     String args() {
         return args;
-    }
-
-    Optional<Comment> comment() {
-        return Optional.ofNullable(comment);
-    }
-
-    boolean isInBody() {
-        return comment == null;
-    }
-
-    boolean isInComment() {
-        return comment != null;
     }
 }
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CommandInvocation.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CommandInvocation.java
@@ -1,6 +1,7 @@
 package org.openjdk.skara.bots.pr;
 
 import org.openjdk.skara.host.HostUser;
+import org.openjdk.skara.issuetracker.Comment;
 
 import java.util.Optional;
 
@@ -10,13 +11,15 @@ class CommandInvocation {
     private final CommandHandler handler;
     private final String name;
     private final String args;
+    private final Comment comment;
 
-    CommandInvocation(String id, HostUser user, CommandHandler handler, String name, String args) {
+    CommandInvocation(String id, HostUser user, CommandHandler handler, String name, String args, Comment comment) {
         this.id = id;
         this.user = user;
         this.handler = handler;
         this.name = name;
         this.args = args != null ? args.strip() : "";
+        this.comment = comment;
     }
 
     String id() {
@@ -37,5 +40,17 @@ class CommandInvocation {
 
     String args() {
         return args;
+    }
+
+    Optional<Comment> comment() {
+        return Optional.ofNullable(comment);
+    }
+
+    boolean isInBody() {
+        return comment == null;
+    }
+
+    boolean isInComment() {
+        return comment != null;
     }
 }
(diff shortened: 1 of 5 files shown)


kevinrushforth:
I presume that the author cannot reduce the number of reviewers once it has been increased, unless they are also a "R"eviewer?

bots/pr/src/main/java/org/openjdk/skara/bots/pr/ReviewersCommand.java
```diff
@@ -49,8 +49,8 @@ private void showHelp(PrintWriter reply) {
 
     @Override
     public void handle(PullRequestBot bot, PullRequest pr, CensusInstance censusInstance, Path scratchPath, CommandInvocation command, List<Comment> allComments, PrintWriter reply) {
-        if (!censusInstance.isReviewer(command.user())) {
-            reply.println("Only [Reviewers](https://openjdk.java.net/bylaws#reviewer) are allowed to change the number of required reviewers.");
+        if (!pr.author().equals(command.user()) && !censusInstance.isReviewer(command.user())) {
```
kevinrushforth, line 52: This would suggest that my assumption is wrong, and that an author can reduce the number of reviewers required even if they are not a Reviewer. I think this might not be what we want.

prrace: I think it is fine to let them increase it.

kevinrushforth: Agreed. It's the case of an author reducing it after a Reviewer has increased it that I think should be disallowed.

edvbld: @kevinrushforth and @prrace - I agree, see changes in commit 347594f

bots/pr/src/main/java/org/openjdk/skara/bots/pr/ReviewersTracker.java
```diff
@@ -99,9 +99,14 @@ String role() {
     }
 
     static Optional<AdditionalRequiredReviewers> additionalRequiredReviewers(HostUser botUser, List<Comment> comments) {
+        return additionalRequiredReviewers(botUser, comments, null);
+    }
+
+    static Optional<AdditionalRequiredReviewers> additionalRequiredReviewers(HostUser botUser, List<Comment> comments, Comment exclude) {
```
rwestberg, line 105: The exclude comment parameter probably isn't the one you are looking for, as the reply marker isn't in the command comment but the reply.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CommandInvocation.java
```diff
@@ -10,13 +11,15 @@
     private final CommandHandler handler;
     private final String name;
     private final String args;
+    private final Comment comment;
 
-    CommandInvocation(String id, HostUser user, CommandHandler handler, String name, String args) {
+    CommandInvocation(String id, HostUser user, CommandHandler handler, String name, String args, Comment comment) {
```
rwestberg, line 16: Think you can drop this parameter if you drop the exclude param!

rwestberg (commented):
Looks good, but you can probably simplify it a bit since the "exclude" parameter isn't really used. :)

bots/pr/src/main/java/org/openjdk/skara/bots/pr/ReviewersTracker.java
edvbld: Thanks, fixed in 7b17768

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CommandInvocation.java
edvbld: Yep, fixed in 7b17768

bots/pr/src/main/java/org/openjdk/skara/bots/pr/ReviewersCommand.java
```diff
@@ -83,6 +108,21 @@ public void handle(PullRequestBot bot, PullRequest pr, CensusInstance censusInst
             role = roleMappings.get(splitArgs[1].toLowerCase());
         }
 
+        if (pr.author().equals(command.user()) && !censusInstance.isReviewer(command.user())) {
```
kevinrushforth, line 111: Maybe simplify this to just `! isReviewer`? Including the test for `is author` doesn't really add value (removing it seems to better reflect the intent and might be safer / more future proof).

kevinrushforth (approved):
Looks good with one minor suggestion (feel free to ignore it).

bots/pr/src/main/java/org/openjdk/skara/bots/pr/ReviewersCommand.java
edvbld: No, we don't want people who are _not_ the PR author and who are _not_ Reviewers to be able to increase the required number of reviewers. Allowing that makes it possible to harass pull request by e.g. registering a spam account and just drop a bunch `/reviewers 9` in a bunch of PRs.

rwestberg (approved):
Looks good!

bots/pr/src/main/java/org/openjdk/skara/bots/pr/ReviewersCommand.java
kevinrushforth: I agree, but that isn't related to this block (your concern is a good one and is handled by the previous block). By adding the `author &&` to `!reviewer` it makes this less restrictive not more. In any case, it doesn't matter...at least not given the prior test.

PR #828, opened 2020-09-17
692

rwestberg:
Update the integration ready message slightly in response to feedback.

Best regards,
Robin


edvbld (approved):
Looks good, just one more test to fix!

PR #830, opened 2020-09-17
649: SKARA will not build/install without termcap

edvbld:
Hi all,

please review this small patch that makes us fall back to `TERM=dumb` if `TERM` is not specified when building on Linux.

Testing:
- [x] `make` works on Linux x64 with and without `TERM` defined

Thanks,
Erik

---------
### Progress

### Issue
 * [SKARA-649](https://bugs.openjdk.java.net/browse/SKARA-649): SKARA will not build/install without termcap

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/830/head:pull/830`
`$ git checkout pull/830`

### DIFF
--- a/gradlew
+++ b/gradlew
@@ -188,6 +188,7 @@ if [ "${OS}" = "Linux" ]; then
     export LC_ALL=en_US.UTF-8
     export LANG=en_US.UTF-8
     export LANGUAGE=en_US.UTF-8
+    export TERM=${TERM:-dumb}
 fi
 
 exec "${GRADLE_LAUNCHER}" "$@"


rwestberg (approved):
Looks good!

gradlew
```diff
@@ -188,6 +188,7 @@ if [ "${OS}" = "Linux" ]; then
     export LC_ALL=en_US.UTF-8
     export LANG=en_US.UTF-8
     export LANGUAGE=en_US.UTF-8
+    export TERM=${TERM:-dumb}
```
phedlin, line 191: TERM=dumb will work as a workaround (dropping the _fancy_ progress bar) but TERM not being set is not the proper condition for lack of **termcap** support (e.g. when **terminfo** is used and TERM set accordingly).

edvbld: @phedlin I'm not sure that there is a way to determine whether a process uses `termcap` or `terminfo`, do you know of a reliable way? I think the best we can do here is to add some documentation to `README.md`, instructing the user to run `sh gradlew --console=plain` as a workaround. Or do you have any other suggestions?

edvbld: @phedlin I might have a way to default to `--console=plain` while still allowing Skara developers to opt-in to a more fancy console. Can you please try building with `sh gradlew --console=plain` and see if it works on your system?

phedlin:
```
JAVA_HOME=~/local/jdk14-import/build/linux-aarch64/images/jdk sh ./gradlew --console=plain

FAILURE: Build failed with an exception.

* What went wrong:
Could not open terminal for stdout: could not get termcap entry

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org
```

PR #848, opened 2020-09-25
update-sync labels should be applied to all-bar-one records

robm-openjdk:
This modification fixes the situation where a backport record may be a member
of two distinct streams and has its hgupdate-sync label removed as a result.

E.g. with backport records for 14, 15, 14.0.1, 15.0.1 no label would be
applied to 15. The current logic gives precedence to the fact that it is the
first release in the 15-updates stream and ignores the fact that it is not
the first release in the feature stream. The label added by the feature stream
is removed when evaluating the 15-updates stream. This should not be the case.

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/848/head:pull/848`
`$ git checkout pull/848`

### DIFF
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/Backports.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/Backports.java
@@ -313,13 +313,12 @@ public class Backports {
         allIssues.addAll(related);
 
         for (var streamIssues : groupByReleaseStream(allIssues)) {
-            // First entry should not have the label
-            var first = streamIssues.get(0);
-            if (first.labels().contains(label)) {
-                first.removeLabel(label);
-            }
-
-            // But all the following ones should
+            // The first issue may have the label if it was part of another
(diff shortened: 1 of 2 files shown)


robm-openjdk:
@kevinrushforth - it'd be great if you could take a look at this when you're able.

rwestberg (approved):
Looks good to me, thanks for fixing! Perhaps @kevinrushforth would like to take a look as well though.

kevinrushforth (approved):
The fix looks good to me. One question about the test: is the `assertNotLabeled` method really needed? It looks like `assertLabeled` will fail unless the set of issues with the labels is exactly as expected (i.e., it will fail if it finds an issue that should have the sync label but doesn't or if if find an issue that should not have the sync label but does), unless I'm missing something.

robm-openjdk:
I think I like the lack of ambiguity in the name. I vote to leave it in, but if Robin objects I can remove it.

rwestberg:
I don't mind keeping it if you think it makes the test more readable!

PR #854, opened 2020-09-29
forge: add search method

edvbld:
Hi all,

please review this patch that adds the `Forge.search` method to allow searching for a commit with a particular hash on a forge.

Thanks,
Erik

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/854/head:pull/854`
`$ git checkout pull/854`

### DIFF
--- a/forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabHost.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabHost.java
@@ -26,12 +26,15 @@ import org.openjdk.skara.forge.*;
 import org.openjdk.skara.host.*;
 import org.openjdk.skara.json.*;
 import org.openjdk.skara.network.*;
+import org.openjdk.skara.vcs.*;
 
 import java.io.IOException;
 import java.net.*;
 import java.nio.charset.StandardCharsets;
 import java.util.*;
+import java.util.stream.Collectors;
 import java.util.logging.Logger;
+import java.time.ZonedDateTime;
 
 public class GitLabHost implements Forge {
     private final String name;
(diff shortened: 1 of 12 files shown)


forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabHost.java
```diff
@@ -210,4 +216,44 @@ public boolean isMemberOf(String groupId, HostUser user) {
...
+    public Optional<HostedCommitMetadata> search(Hash hash) {
+        var hex = hash.hex();
+        for (var group : groups) {
+            var projects = request.get("groups/" + group + "/projects")
+                                  .execute()
+                                  .stream()
+                                  .map(o -> o.get("id").asString())
+                                  .collect(Collectors.toList());
+            for (var project : projects) {
+                var c = request.get("projects/" + project + "/repository/commits/" + hex)
```
rwestberg, line 246: Could we use this endpoint instead? https://docs.gitlab.com/ce/api/commits.html#get-a-single-commit

rwestberg (commented):
Looks good, a bit worried that auto-pagination of thousands of commits will take a while though.. :)

forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabHost.java
edvbld: That is the endpoint the code is using :smile_cat: Did you perhaps miss the `+ hex` part?

edvbld:
> Looks good, a bit worried that auto-pagination of thousands of commits will take a while though.. :)

Since we are limiting the GitHub search to selected organizations, we should for the https://github.com/openjdk organization get at most `O(#repos)` results.

forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabHost.java
rwestberg: Well then, carry on!

rwestberg (approved):

PR #862, opened 2020-10-06
Add support for marking PR body content with an expiration time

rwestberg:
For information in the PR body that is calculated using data from external sources (like Jira, another personal fork, etc.) we need to perform a periodical refresh, as our normal way of detecting updates cannot be used (the updatedAt field of a pull request).

---------
### Progress

### Testing

| Build / test | ⏳ (1/1 running) | ⏳ (1/1 running) | ⏳ (1/1 running) |

### Reviewers
 * [Erik Helin](https://openjdk.java.net/census#ehelin) (@edvbld - **Reviewer**) ⚠️ Review applies to cd19ac11a56a9d040719a5194386545620d0c4af

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/862/head:pull/862`
`$ git checkout pull/862`

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
@@ -68,7 +68,7 @@ class CheckWorkItem extends PullRequestWorkItem {
         try {
             var hasExpired = ExpirationTracker.hasExpired(body);
             if (hasExpired) {
-                return Base64.getUrlEncoder().encodeToString(UUID.randomUUID().toString().getBytes(StandardCharsets.UTF_8));
+                return String.valueOf(Math.random());
             }
 
             var approverString = reviews.stream()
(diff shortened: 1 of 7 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -66,6 +66,11 @@ private String encodeReviewer(HostUser reviewer, CensusInstance censusInstance)
     String getMetadata(CensusInstance censusInstance, String title, String body, List<Comment> comments,
                        List<Review> reviews, Set<String> labels, boolean isDraft) {
         try {
+            var hasExpired = ExpirationTracker.hasExpired(body);
+            if (hasExpired) {
+                return Base64.getUrlEncoder().encodeToString(UUID.randomUUID().toString().getBytes(StandardCharsets.UTF_8));
```
edvbld, line 71:
```suggestion
                return String.valueOf(Math.random());
```

edvbld (approved):
Looks good, just a small suggestion to simplify returning a random string :smile:

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
rwestberg: Guess that makes sense. :)

PR #885, opened 2020-10-15
pr: detect clean backports

edvbld:
Hi all,

please review this patch that adds support to `CheckRun` for identifying "clean" backport pull requests. A backport pull request is "clean" if the diff of the pull request is the same as the diff for the original _modulo_ exact lines being changed. That is, the same backport pull request must contain the same _changes_ as the original commit, but the changes don't have to be applied to the exact same lines. The result is equivalent to running `patch` and passing a fuzzing factor via `--fuzz` that is the same as the number of context lines.

A clean backport also doesn't require reviews. Note though that non-Committers still need a sponsor. I plan to make this configurable in a follow-up patch.

Testing:
- [x] Added a bunch of tests

Thanks,
Erik

---------
### Progress

### Testing

| Build / test | ❌ (1/1 failed) | ✔️ (1/1 passed) | ✔️ (1/1 passed) |

**Failed test task**
- [Linux x64](https://github.com/edvbld/skara/runs/1263795015)

### Reviewers
 * [Robin Westberg](https://openjdk.java.net/census#rwestberg) (@rwestberg - **Reviewer**) ⚠️ Review applies to c33ec6feeaaddaefe24d8643f509cc75960e4e7e

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/885/head:pull/885`
`$ git checkout pull/885`

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -959,7 +959,7 @@ class CheckRun {
             Hash localHash;
             try {
                 // Do not pass eventual original commit even for backports since it will cause
-                // the reviwer check to be ignored.
+                // the reviewer check to be ignored.
                 localHash = checkablePullRequest.commit(commitHash, censusInstance.namespace(), censusDomain, null, null);
             } catch (CommitFailure e) {
                 additionalErrors = List.of(e.getMessage());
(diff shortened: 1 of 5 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -846,6 +958,8 @@ private void checkStatus() {
             List<String> additionalErrors = List.of();
             Hash localHash;
             try {
+                // Do not pass eventual original commit even for backports since it will cause
+                // the reviwer check to be ignored.
```
rwestberg, line 962:
```suggestion
                // the reviewer check to be ignored.
```

rwestberg (approved):
Looks good!

PR #929, opened 2020-10-29
Add tabfiles to jcheck configuration

magicus:
The tabfiles setting is just like the files setting, but it lists files in which tabs are allowed.

(This is to support makefiles)

---------
### Progress

### Reviewers
 * [Erik Joelsson](https://openjdk.java.net/census#erikj) (@erikj79 - **Reviewer**) ⚠️ Review applies to 4aaa659f33923f13afe96ab50a75a67d27eafd70

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/929/head:pull/929`
`$ git checkout pull/929`

### DIFF
--- a/jcheck/src/main/java/org/openjdk/skara/jcheck/WhitespaceCheck.java
+++ b/jcheck/src/main/java/org/openjdk/skara/jcheck/WhitespaceCheck.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2020, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -26,11 +26,8 @@ import org.openjdk.skara.census.Census;
 import org.openjdk.skara.vcs.Commit;
 import org.openjdk.skara.vcs.openjdk.CommitMessage;
 
-import java.io.IOException;
-import java.nio.file.Path;
 import java.util.*;
 import java.util.regex.Pattern;
-import java.util.stream.Stream;
 import java.util.logging.Logger;
 
 public class WhitespaceCheck extends CommitCheck {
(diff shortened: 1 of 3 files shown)


edvbld:
How about naming the configuration `ignore-tabs` instead? To make it a bit more clear what it does :)

jcheck/src/main/java/org/openjdk/skara/jcheck/WhitespaceCheck.java
```diff
@@ -56,7 +54,8 @@
                             var row = hunk.target().range().start() + i;
                             var tabIndex = line.indexOf('\t');
                             var crIndex = line.indexOf('\r');
-                            if (tabIndex >= 0 || crIndex >= 0 || line.endsWith(" ")) {
+                            var ignoreTab = tabPattern.matcher(path.toString()).matches();
+                            if ((tabIndex >= 0 && !ignoreTab) || crIndex >= 0 || line.endsWith(" ")) {
```
erikj79, line 58: This won't catch if a tabFile has a trailing tab.

magicus: I'm not sure I understand. Do you want jcheck to complain about trailing tabs, even if the file type is on tabfiles/ignore-tabs?

magicus:
@edvbld 
> How about naming the configuration `ignore-tabs` instead? To make it a bit more clear what it does :)

Sure. I checked for "prior art" on multiword keywords, but did not find any such instances. :) This will mean that the corresponding Java variable needs a different name, since "-" is not valid in Java identifiers.

(And if we're going this way, personally I think `allow-tabs` would be even better)

erikj79 (approved):

rwestberg:
Looks good in general, but would be nice with a test to see that it still flags trailing tabs (which I think it is meant to do still, right?)

magicus:
@rwestberg @edvbld I have now added a test that verifies that trailing tabs are always rejected, and that leading tabs are accepted iff ignore-tabs is true for that file.

Ok to integrate now?

edvbld (approved):
Looks good!

PR #935, opened 2020-11-01
Replace explicit type with <>

amCap1712:
Hi!
Analyzing the skara codebase in Intellij IDEA reveals that there several opportunities for minor improvements in the codebase. One such improvement is to use the `<>` instead of explicit types. Hence, I propose this patch.

There are other inspections and improvements possible as well. In case such changes are welcome, I would be happy to contribute patches for those as well. Also, in that case should those changes be a part of this PR or should there be a separate PR for each ?
Thanks.
Regards,
Kartik

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/935/head:pull/935`
`$ git checkout pull/935`


edvbld (approved):
Looks good!

PR #937, opened 2020-11-02
Switch to using GitLab approvals instead of award_emojis

rwestberg:
Later versions of GitLab allows Merge Request approvals even in the CE version, so we can start to use that instead of the award-emoji workaround used previously.

---------
### Progress

### Testing

| Build / test | ✔️ (1/1 passed) | ✔️ (1/1 passed) | ✔️ (1/1 passed) |

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/937/head:pull/937`
`$ git checkout pull/937`


magicus:
So this will remove the ability to review using emojis? This sounds like there need to be a "flag day" for switching review methods. (I'm not saying this is bad, just verifying that I understand the implications correctly.)

kevinrushforth:
Will this be a hard switch, or will you continue to also honor the "thumbs-up" emoji?

rwestberg:
This will be a hard switch, the thumbs-up button has an unfortunate drawback, which is that GitLab does not update the merge request's "updated_at" field when it is used. This in turn means that we have had to disable several optimizations to reduce network traffic -> we have to use a much lower polling rate for GitLab -> command replies can take up to a minute instead of a few seconds.

magicus:
@rwestberg Sounds like an acceptable rationale for disabling the old behavior. Just make sure to communicate clearly at which point review method changes.

PR #954, opened 2020-11-17
SKARA-758: Add /check pull request command

amCap1712:
Hi!
Kindly review the patch to allow manual execution of jcheck. I think it might be worthwhile to restrict this command to some specific people but I am not sure who that should be (maybe restricting to the author is enough but a reviewer might want to execute it as well in case its urgent). 

I wanted to test this locally by setting up the skara bots on my own repository but was unable to figure out how to do so due to lack of documentation.
Thanks.
Regards,
Kartik

---------
### Progress

### Testing

|     | Linux x64 | Windows x64 |
| --- | ----- | ----- |
| Build / test | ✔️ (1/1 passed) | ✔️ (1/1 passed) |

### Issue
 * [SKARA-758](https://bugs.openjdk.java.net/browse/SKARA-758): Add /check pull request command

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/954/head:pull/954`
`$ git checkout pull/954`

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckCommand.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckCommand.java
@@ -1,0 +1,22 @@
+package org.openjdk.skara.bots.pr;
+
+import org.openjdk.skara.forge.PullRequest;
+import org.openjdk.skara.issuetracker.Comment;
+
+import java.io.PrintWriter;
+import java.nio.file.Path;
+import java.time.Instant;
+import java.util.List;
+
+public class CheckCommand implements CommandHandler {
+    @Override
+    public String description() {
+        return "run Jcheck again";
+    }
+
+    @Override
+    public void handle(PullRequestBot bot, PullRequest pr, CensusInstance censusInstance, Path scratchPath, CommandInvocation command, List<Comment> allComments, PrintWriter reply) {
+        bot.scheduleRecheckAt(pr, Instant.now());
+    }
+
+}
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckCommand.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckCommand.java
@@ -1,11 +1,15 @@
 package org.openjdk.skara.bots.pr;
 
+import org.openjdk.skara.forge.HostedRepositoryPool;
 import org.openjdk.skara.forge.PullRequest;
+import org.openjdk.skara.forge.PullRequestUtils;
 import org.openjdk.skara.issuetracker.Comment;
 
+import java.io.IOException;
 import java.io.PrintWriter;
+import java.io.UncheckedIOException;
 import java.nio.file.Path;
-import java.time.Instant;
+import java.util.HashSet;
 import java.util.List;
 
 public class CheckCommand implements CommandHandler {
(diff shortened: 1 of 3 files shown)


edvbld:
Hi @amCap1712, thanks for working on this!

Please consider adding a unit test for the `/check` command, see e.g. [SummaryTests.java](https://github.com/openjdk/skara/blob/master/bots/pr/src/test/java/org/openjdk/skara/bots/pr/SummaryTests.java) for an example of a unit test for a pull request command.

Thanks,
Erik

amCap1712:
Hi @edvbld,
Thanks for the review. I have added a unit test as you asked. Let me know if any other changes are required.
Regards,
Kartik.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckCommand.java
```diff
@@ -0,0 +1,41 @@
...
+public class CheckCommand implements CommandHandler {
+    @Override
+    public String description() {
+        return "run Jcheck again";
+    }
+
+    @Override
+    public void handle(PullRequestBot bot, PullRequest pr, CensusInstance censusInstance, Path scratchPath, CommandInvocation command, List<Comment> allComments, PrintWriter reply) {
+        try {
+            var workItem = new CheckWorkItem(bot, pr, e -> {});
```
rwestberg, line 24: Don't think you actually have to do the check here, `CommandWorkItem` will schedule an additional check after every command anyway. But to force the check to run you'll need to ensure that the "metadata" of the existing check becomes invalid. If you update `metadataComments` to add a special reply marker from the `/check` command, and ensure that the check command outputs it, I think thinks should work.

bots/pr/src/test/java/org/openjdk/skara/bots/pr/CheckCommandTests.java
```diff
@@ -0,0 +1,47 @@
...
+            var localRepo = CheckableRepository.init(localRepoFolder, author.repositoryType());
+            var masterHash = localRepo.resolve("master").orElseThrow();
+            assertFalse(CheckableRepository.hasBeenEdited(localRepo));
+            localRepo.push(masterHash, author.url(), "master", true);
+
+            // Make a change with a corresponding PR
+            var editHash = CheckableRepository.appendAndCommit(localRepo);
+            localRepo.push(editHash, author.url(), "edit", true);
+            var pr = credentials.createPullRequest(author, "master", "edit", "This is a pull request");
+            pr.addComment("/check");
```
rwestberg, line 39: Looks like a good start, but it's possible that this test is a bit too simple. Would it still pass if you removed this line (as checking will occur anyway)?

rwestberg (changes requested):

rwestberg:
> I wanted to test this locally by setting up the skara bots on my own repository but was unable to figure out how to do so due to lack of documentation.

Yeah, sorry about that, it has not really been a priority as no-one else actually run them in "production". It can be a bit complicated to configure them properly for running stand-alone, but if you want to run a test against GitHub instead of the mock implementation that's a bit easier. If you want to try that I'll be happy to assist with the required configuration.

amCap1712:
@rwestberg Yes, please. I plan to keep contributing in the future and having such a setup will help me to explore and understand the code much more quickly and in a better way. Hence, I think setting it up will be worth it. Thanks again for the help.

rwestberg:
Sure, let's start with the required setup for running tests against GitHub (or GitLab) then! You can pass `-Pcredentials=/path/to/credentials.json` where the file contains something like this:

```json
{
    "type": "github",
    "host": "https://github.com",
    "project": "rwestberg/skara-integration",
    "namespace": "github.com",
    "apps": [
        {
            "key": "certificates/sbot.private-key.pkcs8.pem",
            "id": "12837",
            "installation": "195957"
        },
        {
            "key": "certificates/sbot2.private-key.pkcs8.pem",
            "id": "18848",
            "installation": "379295"
        },
        {
            "key": "certificates/sbot3.private-key.pkcs8.pem",
            "id": "19394",
            "installation": "397926"
        }
    ]
}
```

Most tests require two or three different accounts in order to simulate various interactions, so that's why there are a list of them. Each account credential corresponds to a GitHub app that you can create on your account. The `id` can be seen after creating the app, the `installation` value can be found after the app is installed on your account. Finally, the key is the private key you associate with the GitHub app. The path is relative to the .json file. Note that it has to be in PEM format, you'll have to convert the file you get from GitHub.

amCap1712:
I have created the 3 Github Apps and generated the .json file as you instructed. How should I proceed ?

rwestberg:
Now you can use the credentials file when running the tests through Gradle, probably something like this (I usually run from IntelliJ):

`sh gradlew --tests *testCheckRun* -Pcredentials=/path/to/credentials.json`

PR #957, opened 2020-11-18
Always recheck in-progress checks

rwestberg:
If we end up wanting to run jcheck for a PR that has it marked as in progress, always rerun it. The only way this can happen is if the running check was interrupted (such as the process was killed), checks will never run concurrently for the same PR.

---------
### Progress

### Testing

| Build / test | ✔️ (1/1 passed) | ✔️ (1/1 passed) | ⏳ (1/1 running) |

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/957/head:pull/957`
`$ git checkout pull/957`


edvbld (approved):
Looks good!

PR #959, opened 2020-11-19
Handle trailing blank lines in commit message

rwestberg:
This change ensures that a commit message ending with a blank line does not generate an exception.

### DIFF
--- a/vcs/src/main/java/org/openjdk/skara/vcs/openjdk/CommitMessageParsers.java
+++ b/vcs/src/main/java/org/openjdk/skara/vcs/openjdk/CommitMessageParsers.java
@@ -116,6 +116,9 @@ public class CommitMessageParsers {
             Hash original = null;
             while (i < lines.size() && lines.get(i).equals("")) {
                 i++;
+                if (i >= lines.size()) {
+                    break;
+                }
 
                 if (lines.get(i).startsWith("Co-authored-by:") ||
                     lines.get(i).startsWith("Reviewed-by:") ||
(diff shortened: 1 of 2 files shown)


vcs/src/main/java/org/openjdk/skara/vcs/openjdk/CommitMessageParsers.java
```diff
@@ -116,6 +116,9 @@ public CommitMessage parse(List<String> lines) {
             Hash original = null;
             while (i < lines.size() && lines.get(i).equals("")) {
                 i++;
+                if (i >= lines.size()) {
+                    break;
+                }
```
JornVernee, line 121: I guess I'm a little confused as to how this fixes the problem :/ Is it because `firstDelimiter` is false after the first line and then it adds a blank?

Does it make sense to move the `i++` to the end of the while loop, and then increment it one time separately before the loop, so that the while loop condition can check the `i >= lines.size()` condition, and no separate `if` is needed?

rwestberg: I agree that the code is a bit unintuitive, have had to look at it a few times.. :) But no, incrementing should only be done if we're at a blank line, so can't do it unconditionally outside of the loop..

I pushed a different version that is perhaps a bit more clear.

JornVernee (approved):
Okay, thanks for explaining. I think the new version looks a little better. (but it does need a test change?)

rwestberg:
Well, the test is new so I don't think it matters - a commit message like this could never be generated by the bots, but a similar one was pushed directly to a repo and made the notifier fail.. So perhaps this is more correct than the first version. :)

PR #963, opened 2020-11-26
Add newline after closed issue warning

JornVernee:
Adds a missing newline after the 'issue is not open' warning, which is currently missing, and leads to the list not being formatted correctly. See e.g. https://github.com/openjdk/jdk/pull/1444

image

Any testing needed for this?

---------
### Progress

### Reviewers
 * [Robin Westberg](https://openjdk.java.net/census#rwestberg) (@rwestberg - **Reviewer**) ⚠️ Review applies to 00e70cda7833f59d8c237a42fb1185e040a973fc

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/963/head:pull/963`
`$ git checkout pull/963`

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -482,9 +482,8 @@ class CheckRun {
                             }
                             if (iss.get().state() != org.openjdk.skara.issuetracker.Issue.State.OPEN) {
                                 if (!pr.labels().contains("backport")) {
-                                    progressBody.append(" ⚠️ Issue is not open.\n");
+                                    progressBody.append(" ⚠️ Issue is not open.");
                                 }
-                                continue;
                             }
                             if (properties.containsKey("issuetype") && !primaryTypes.contains(properties.get("issuetype").asString())) {
                                 progressBody.append(" ⚠️ Unexpected issue type `").append(properties.get("issuetype").asString()).append("`.");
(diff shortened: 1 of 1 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -482,7 +482,7 @@ private String getStatusMessage(List<Comment> comments, List<Review> reviews, Pu
                             }
                             if (iss.get().state() != org.openjdk.skara.issuetracker.Issue.State.OPEN) {
                                 if (!pr.labels().contains("backport")) {
-                                    progressBody.append(" ⚠️ Issue is not open.");
+                                    progressBody.append(" ⚠️ Issue is not open.\n");
                                 }
                                 continue;
```
rwestberg, line 487: I know I wrote this, but another solution would be to drop the `continue` here, I don't remember why it's there. :)

rwestberg (approved):
Looks good, I guess a test would be nice, but since there isn't one at the moment there's no need to remedy that right now unless you really want to.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
JornVernee: Ok, will do that than. (seems a little cleaner to me). I think that will only change whether the `if` below is evaluated. Maybe you wanted only one warning to be printed at a time?

rwestberg: Yeah I guess, but then it was a bit inconsistent with the rest of the warnings anyway... So this looks good to me!

PR #971, opened 2020-12-10
Use headers and other info from the most recent response

rwestberg:
When a 304 response is received from a GET request, we retrieve the original request from the cache and return it instead. However, we should only reuse the cached body as other headers (such as links to further result pages) may have been updated.

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/971/head:pull/971`
`$ git checkout pull/971`


edvbld (approved):
Looks good!

PR #972, opened 2020-12-14
The latest request may not contain the full set of headers

rwestberg:
When returning a cached result as a response to a 304, use the combined set of headers.

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/972/head:pull/972`
`$ git checkout pull/972`


edvbld (approved):
Looks good!

PR #985, opened 2021-01-15
749: If an integration fails at a late stage, it will both say that it succeeded and failed

rwestberg:
Wait with outputting the success message until after the final push command during `/integrate` or `/sponsor` has actually completed.

---------
### Progress

### Issue
 * [SKARA-749](https://bugs.openjdk.java.net/browse/SKARA-749): If an integration fails at a late stage, it will both say that it succeeded and failed

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/985/head:pull/985`
`$ git checkout pull/985`


edvbld (approved):
Looks good!

PR #995, opened 2021-01-29
checkout: sort marks before converting

edvbld:
Hi all,

please review this small patch that ensures that we sort the existing marks before passing them to the converter.

Thanks,
Erik

---------
### Progress

### Reviewers
 * [Jorn Vernee](https://openjdk.java.net/census#jvernee) (@JornVernee - **Reviewer**)

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/995/head:pull/995`
`$ git checkout pull/995`

### DIFF
--- a/bots/checkout/src/main/java/org/openjdk/skara/bots/checkout/CheckoutBot.java
+++ b/bots/checkout/src/main/java/org/openjdk/skara/bots/checkout/CheckoutBot.java
@@ -135,6 +135,7 @@ public class CheckoutBot implements Bot, WorkItem {
                     var toRepo = Repository.get(to).orElseThrow(() ->
                         new IllegalStateException("Repository vanished from " + to));
                     var existing = new ArrayList<>(marks.current());
+                    Collections.sort(existing);
                     log.info("Found " + existing.size() + " existing marks");
                     converter.convert(fromRepo, toRepo, existing);
                     hasConverted = true;


rwestberg (approved):
Looks good!

bots/checkout/src/main/java/org/openjdk/skara/bots/checkout/CheckoutBot.java
```diff
@@ -135,6 +135,7 @@ public String toString() {
                     var toRepo = Repository.get(to).orElseThrow(() ->
                         new IllegalStateException("Repository vanished from " + to));
                     var existing = new ArrayList<>(marks.current());
+                    Collections.sort(existing);
```
JornVernee, line 138: If the `convert` method relies on the marks being sorted, maybe it would be more robust to move the call to sort into there?

JornVernee (approved):
One comment, but up to you whether to address

bots/checkout/src/main/java/org/openjdk/skara/bots/checkout/CheckoutBot.java
edvbld: Yeah, I agree. Problem is that the `convert` method takes a `List<Mark>` (which might be an `UnmodifiableList`), so I can't just pass it to `Collections.sort` :disappointed: I would have to make a copy of the passed list and then sort _that_ list, and that just seems a bit annoying (give that we can have around 100,000 marks).

PR #997, opened 2021-02-02
868: Skara should ignore .gitconfig and /etc/gitconfig

magicus:
Skara is already ignoring .hgrc for mercurial, but the corresponding solution for git was never put in place.

I also fixed a problem where an error condition would cause an infinite loop, which I ran into during testing.

---------
### Progress

### Issue
 * [SKARA-868](https://bugs.openjdk.java.net/browse/SKARA-868): Skara should ignore .gitconfig and /etc/gitconfig

### Reviewers
 * [Erik Helin](https://openjdk.java.net/census#ehelin) (@edvbld - **Reviewer**) ⚠️ Review applies to 96e826123adb9dcadd11ca971dc238815e6485db

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/997/head:pull/997`
`$ git checkout pull/997`

### DIFF
--- a/process/src/main/java/org/openjdk/skara/process/Process.java
+++ b/process/src/main/java/org/openjdk/skara/process/Process.java
@@ -67,7 +67,7 @@ public class Process {
             return this;
         }
 
-        public Description environMap(Map<String, String> keyValueMap) {
+        public Description environ(Map<String, String> keyValueMap) {
             if (keyValueMap != null) {
                 getCurrentProcessBuilderSetup().environment.putAll(keyValueMap);
             }
--- a/process/src/main/java/org/openjdk/skara/process/Process.java
+++ b/process/src/main/java/org/openjdk/skara/process/Process.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2021, Oracle and/or its affiliates. All rights reserved.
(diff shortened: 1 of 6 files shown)


edvbld:
@magicus thanks for contributing! What do you think about adding a unit test to `RepositoryTests.java`?

magicus:
@edvbld What exactly do you propose to test in the unit test? This is really solving a problem with the environment -- the tests fail when I run as my user due to conflicting modifications in my `~/.gitconfig`. I can't really modify the users personal gitconfig file in a unit test. The only thing I could really think if is verifying that `git config -l` is empty. Hm, maybe that's a good idea...

edvbld:
@magicus can't you run a test that asserts that `user.name` is empty? Almost everyone configures `user.name` (Git sort-of requires it for commits). In a CI environment the test will of course be trivially correct since we never setup `user.name`, but the test will be accurate when run on developer's personal machines.

magicus:
If `user.name` is set, then `git config -l` will not be empty. :)

process/src/main/java/org/openjdk/skara/process/Process.java
```diff
@@ -67,6 +67,13 @@ public Description environ(String key, String value) {
             return this;
         }
 
+        public Description environMap(Map<String, String> keyValueMap) {
```
edvbld, line 70: I would have just overloaded `environ`, as in `public Description environ(Map<String, String> env)`

edvbld (approved):
Looks good, just a minor comment inline.

PR #1004, opened 2021-02-10
build: use included build instead of buildSrc

edvbld:
Hi all,

please review this small patch that makes the build use "included builds" for Skara's Gradle plugins instead of the "buildSrc" feature of Gradle. The "buildSrc" feature doesn't play nice with Gradle's caching mechanism - any change to a file in the `buildSrc` directory invalidates the build cache. The solution is to just include the plugins using the `includeBuild` feature (the Gradle team is even thinking of making `buildSrc` an implicitly included build, see https://github.com/gradle/gradle/issues/2531).

Testing:
- [x] `make images` on Linux x64

Thanks,
Erik

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/1004/head:pull/1004`
`$ git checkout pull/1004`


rwestberg (approved):
Good stuff!

build.gradle
rybak, line 1: I've stumbled upon an outdated reference to `buildSrc` in the root `build.gradle`:

```gradle
        // Force Gradle to load the JUnit Platform Launcher from the module-path, as
        // configured in buildSrc/.../ModulePlugin.java -- see SKARA-69 for details.
        testRuntimeOnly 'org.junit.platform:junit-platform-launcher:1.6.2'
```

https://github.com/openjdk/skara/pull/1004/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7L49-R62

PR #1009, opened 2021-02-12
build: run build and tests in parallel

edvbld:
Hi all,

please review this patch that enables builds and tests to happen in parallel. I also tweaked the build a little:

- the Gradle daemon now gets an `Xmx` of `2048 Mb`
- Gradle will run up to 100 tests in the same JVM before forking a new one
- the JVM for compiling will be forked, reducing heap pressure on the Gradle daemon thread
- enabled the file system watching feature for faster rebuilds

The results on my AMD Ryzen 5 3600 is that `sh gradlew clean && sh gradlew --info test` went from ~10 minutes to ~2 minutes :partying_face: :rocket: 

Testing:
- [x] `make images` on Linux x64

Thanks,
Erik

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/1009/head:pull/1009`
`$ git checkout pull/1009`


rwestberg (approved):
Looks good!

PR #1011, opened 2021-02-12
574: git-info should use info from commit notifications

magicus:
git-info should use info from commit notifications to get the link to the review, not JBS (since not all PR have issues)

This required refactoring to move the hosted repo logic from cli/pr/Utils.java to where it can be accessed by all CLI tools. After some deliberating, I left a bunch of trivial wrappers in cli/pr/Utils.java -- otherwise this change would have touched *all* the files in cli/pr.

Testing: ad hoc testing of "git info" on some arbitrary Skara commits.

---------
### Progress

### Issue
 * [SKARA-574](https://bugs.openjdk.java.net/browse/SKARA-574): git-info should use info from commit notifications

### Reviewers
 * [Erik Helin](https://openjdk.java.net/census#ehelin) (@edvbld - **Reviewer**) ⚠️ Review applies to 50d827563e47142b011e9cee801e598112f5c2c3

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/1011/head:pull/1011`
`$ git checkout pull/1011`

### DIFF
--- a/forge/src/main/java/org/openjdk/skara/forge/HostedRepository.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/HostedRepository.java
@@ -28,6 +28,7 @@ import org.openjdk.skara.vcs.*;
 import java.net.URI;
 import java.time.ZonedDateTime;
 import java.util.*;
+import java.util.regex.Pattern;
 
 public interface HostedRepository {
     Forge forge();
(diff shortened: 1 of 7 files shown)


forge/src/main/java/org/openjdk/skara/forge/HostedRepository.java
```diff
@@ -82,4 +83,29 @@ default PullRequest createPullRequest(HostedRepository target,
...
+        }
+
+        /** The review comment looks like this:
+         * <!-- COMMIT COMMENT NOTIFICATION -->
+         * ### Review
+         *
+         * - [openjdk/skara/123](https://git.openjdk.java.net/skara/pull/123)
+         */
+
+        var pattern = Pattern.compile("### Review[^]]*]\\((.*)\\)");
```
edvbld, line 103: Do you need to use `Pattern.DOTALL` here? As in `Pattern.compile("### Review[^]]*]\\((.*)\\)", Pattern.DOTALL);` ? I'm thinking about the `.*` handling `\n`.

magicus: As you can see from the test, this works just fine. :-)

The `DOTALL` flag only applies to the dot `.`. I'm searching for "anything but ]", `[^]]` which also includes line terminators. And the URL (that I match with `.*`) should of course not have line breaks in it.

edvbld (approved):
Looks good modulo some whitespace issues found by jcheck :robot:

PR #1025, opened 2021-02-19
cli: rework git-backport

edvbld:
Hi all,

please review this patch that reworks how `git-backport` functions. `git-backport` is now much smaller (and more focused) than it used to be and now composes much better with other Skara CLI tools like `git-pr-create`. `git-backport` now simply fetches a commit from a remote repository and cherry-picks the fetched commit on top of the current branch (without committing). `git-backport` then finally makes a commit with the proper backport commit message (a commit message of the form `Backport <HASH>`).

The tool `git-backport` is now meant to be used in combination with `git-pr-create`, `git-pr-set` and `git-pr-integrate`. The following example shows how a commit can be backported, a pull request created, marking the pull request as clean and then finally integrating the pull request:

```
$ git checkout -b backport-5a01c3d68ac2
$ git backport --from=openjdk/jdk 5a01c3d68ac22b7ee6f0746405a9bdef43281cb7
$ git pr create
$ sleep 15 # give the bots some time to work
$ git pr set --clean
$ sleep 15 # allow the bots to do some more work
$ git pr integrate
$ git checkout -
```

All (or parts) of the above can of course be wrapped into a [git alias](https://git-scm.com/book/en/v2/Git-Basics-Git-Aliases) for those that don't like to type. The following show an example alias defined in `~/.gitconfig`:

```
[alias]
bp = "!f() { \
        git checkout -b backport-$1 && \
        git backport --from=openjdk/jdk $1 && \
        git pr create; \
      }; f"
```

A user can then simply run `git bp 5a01c3d68ac22b7ee6f0746405a9bdef43281cb7`.

The `--from` parameter to `git-backport` is configurable, so a user who most likely often backports from [openjdk/jdk](https://github.com/openjdk/jdk) can run the following to never have to set `--from` on the command-line:

```
$ git config --global backport.from https://github.com/openjdk/jdk
```

Note that `git-backport` only requires a personal access token (PAT) in the case when the `--from` option contains a simple repository name, for example `jdk`. In all other cases there is no need for a PAT, making the command accessible for those who are not using PATs. 

Testing:

Thanks,
Erik

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/1025/head:pull/1025`
`$ git checkout pull/1025`

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ and [Git](https://git-scm.com/) hashes
 - git-skara - learn about and update the Skara CLI tools
 - git-trees - run a git command in a tree of repositories
 - git-publish - publishes a local branch to a remote repository
+- git-backport - backports a commit from another repository onto the current branch
 
 There are also CLI tools available for importing OpenJDK
 [Mercurial](https://mercurial-scm.org/) repositories into
--- a/cli/src/main/java/org/openjdk/skara/cli/GitBackport.java
+++ b/cli/src/main/java/org/openjdk/skara/cli/GitBackport.java
@@ -31,7 +31,7 @@ import org.openjdk.skara.version.Version;
 import org.openjdk.skara.proxy.HttpProxy;
 
 import java.io.IOException;
-import java.net.URI;
+import java.net.*;
 import java.nio.file.*;
 import java.util.*;
 import java.util.function.Supplier;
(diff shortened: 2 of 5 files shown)


rwestberg (approved):
Looks good!

RealCLanger:
Cool. Looking forward to this being integrated. When will the /backport comment be enabled?

PR #1032, opened 2021-02-23
forge: fix GitLabRepository.recentCommitComments

edvbld:
Hi all,

please review this patch that re-works `GitLab.recentCommitComments` into its hopefully final shape. This patch fixes two problems:

- multiple commits can have the same commit message title (i.e. first line of commit message)
- GitLab sometimes returns a trimmed `target_title` for commit comment events

The first problem is fixed by always starting out with fetching the entire history (once) and then always return a set of candidate hashes for a given commit title. We can then fetch the commit comments for each of the candidate hashes to figure which hash the comment was made on. The solution to the second problem then piggy-backs on the first solution: if a hash with a given title isn't found, and the title ends with `...`, then go through all commit message titles and check if any of them has the given title as prefix.

The drawback is that it can take a couple of minutes to populate the initial commit message title to hash mapping, but since 874a89ceb0d5fd742db75b6f9744b56e200cfe32 this will run in a separate `WorkItem`, so the degraded performance won't affect other `WorkItem`s during startup. Once the map is populated there will only be one additional REST request for most calls to `recentCommitComments`.

Testing:
- [x] Manual testing against GitLab on Linux x64 with `git skara debug commit-comments`

Thanks,
Erik

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/1032/head:pull/1032`
`$ git checkout pull/1032`

### DIFF
--- a/forge/src/main/java/org/openjdk/skara/forge/CommitComment.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/CommitComment.java
@@ -29,11 +29,9 @@ import org.openjdk.skara.vcs.Hash;
 import java.nio.file.Path;
 import java.time.ZonedDateTime;
 import java.util.*;
-import java.util.function.Supplier;
 
 public class CommitComment extends Comment {
     private Hash commit;
-    private final Supplier<Hash> commitSupplier;
     private final Path path;
     private final int line;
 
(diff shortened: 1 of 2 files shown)


rwestberg (approved):
Looks good!

PR #1041, opened 2021-02-28
pr: update all local branches for commit comments

edvbld:
Hi all,

please review this patch that makes `CommitCommentsWorkItem` properly update all local branches prior to calling `ReadOnlyRepository.commitMetdata`. `ReadOnlyRepository.commitMetadata` is essentially `rev-list —all` for Git and all local refs need to be updated for this to work. In order to make as few networks call as possible I manually merge each local branch with its remote counterpart (instead of doing `pull`).

Testing:

Thanks,
Erik


bots/pr/src/main/java/org/openjdk/skara/bots/pr/CommitCommentsWorkItem.java
```diff
@@ -67,12 +67,14 @@ private boolean isAncestor(ReadOnlyRepository repo, Hash ancestor, Hash descenda
             var hostedRepositoryPool = new HostedRepositoryPool(seedPath);
             var localRepoPath = scratchPath.resolve("pr").resolve("commit-comments").resolve(bot.repo().name());
             var localRepo = hostedRepositoryPool.materialize(bot.repo(), localRepoPath);
+            localRepo.fetchAllRemotes(false);
```
rwestberg, line 70: Not sure this is needed? The materialize call should have done a fetchAll as well..

rwestberg (approved):
Looks good!

PR #1042, opened 2021-03-01
Read resolvedInBuild JBS property properly

rwestberg:
The resolvedInBuild property was not read correctly from JBS, leading to incorrect issue labels.

---------
### Progress

### Reviewers
 * [Magnus Ihse Bursie](https://openjdk.java.net/census#ihse) (@magicus - Committer)

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/1042/head:pull/1042`
`$ git checkout pull/1042`

### DIFF
--- a/jbs/src/main/java/org/openjdk/skara/jbs/Backports.java
+++ b/jbs/src/main/java/org/openjdk/skara/jbs/Backports.java
@@ -73,8 +73,8 @@ public class Backports {
             log.warning("Issue " + issue.id() + " has multiple valid fixVersions - ignoring");
             return Optional.empty();
         }
-        if (issue.properties().containsKey("customfield_10006") && issue.properties().get("customfield_10006").isObject()) {
-            return JdkVersion.parse(versionString.get(0), issue.properties().get("customfield_10006").get("value").asString());
+        if (issue.properties().containsKey("customfield_10006")) {
+            return JdkVersion.parse(versionString.get(0), issue.properties().get("customfield_10006").asString());
         } else {
             return JdkVersion.parse(versionString.get(0));
         }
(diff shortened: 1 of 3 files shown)


jbs/src/main/java/org/openjdk/skara/jbs/Backports.java
```diff
@@ -73,8 +73,8 @@ private static boolean isNonScratchVersion(String version) {
             log.warning("Issue " + issue.id() + " has multiple valid fixVersions - ignoring");
             return Optional.empty();
         }
-        if (issue.properties().containsKey("customfield_10006") && issue.properties().get("customfield_10006").isObject()) {
-            return JdkVersion.parse(versionString.get(0), issue.properties().get("customfield_10006").get("value").asString());
+        if (issue.properties().containsKey("customfield_10006")) {
+            return JdkVersion.parse(versionString.get(0), issue.properties().get("customfield_10006").asString());
```
edvbld, line 77: Would you mind creating a variable naming the custom field? For example:

```java
var resolvedInBuild = "customfield_10006";
if (issue.properties().containsKey(resolvedInBuild)) {
```

edvbld (commented):
Looks good, I just have a minor comment inline!

magicus (approved):
LGTM (but I agree with Erik's suggestion).

jbs/src/main/java/org/openjdk/skara/jbs/Backports.java
rwestberg: This exist in quite a few places already, and properly fixing it would require a larger refactoring. So I think we should do that separately.

edvbld: Ok, sounds good!

edvbld (approved):

PR #1044, opened 2021-03-09
Improve GitHub review comment line number parsing

rwestberg:
When a comment is made on the "previous" version of a change, the line number may not match the reported hash. This should improve things. Tested manually on https://github.com/openjdk/jdk/pull/2843/files .


magicus (approved):
LGTM.

PR #1045, opened 2021-03-09
636

rwestberg:
It's not enough to use \ to escape the @ character, so we can use this: @nomention instead to prevent unwanted mentions.


magicus (approved):
LGTM

edvbld (approved):
Haha, nice workaround with the HTML comment :laughing: Looks good, thanks for fixing this :partying_face:

PR #1047, opened 2021-03-10
forge: add methods addCollaborator and canPush

edvbld:
Hi all,

please review this patch that adds two methods to `HostedRepository`:

- `addCollabarator`
- `canPush`

These methods can be used to check and grant access for a user to a particular hosted repository.

Thanks,
Erik


rwestberg (approved):
Looks good!

PR #1049, opened 2021-03-11
913: Skara should honor .gitconfig and /etc/gitconfig

magicus:
In SKARA-868, the fix to ignore .gitconfig and /etc/gitconfig was applied universally, which is causing trouble for users. The intention from the filer was to just disable this when running the Skara tests. I still think we should have a way to ignore gitconfig when running tests, but we need to keep honoring the git configuration at runtime.

---------
### Progress

### Issue
 * [SKARA-913](https://bugs.openjdk.java.net/browse/SKARA-913): Skara should honor .gitconfig and /etc/gitconfig

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/1049/head:pull/1049`
`$ git checkout pull/1049`

### DIFF
--- a/vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
+++ b/vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
@@ -50,8 +50,8 @@ public class GitRepository implements Repository {
     private Path cachedRoot = null;
     private static final Hash EMPTY_TREE = new Hash("4b825dc642cb6eb9a060e54bf8d69288fbee4904");
 
-    public static void ignoreUserConfiguration(boolean ignore) {
-        currentEnv = ignore ? NO_CONFIG_ENV : Collections.emptyMap();
+    public static void ignoreConfiguration() {
+        currentEnv = NO_CONFIG_ENV;
     }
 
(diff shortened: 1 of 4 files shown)


vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
```diff
@@ -38,17 +38,22 @@
...
             "GIT_CONFIG_NOSYSTEM", "true"
     );
 
+    public static Map<String, String> currentEnv = Collections.emptyMap();
     private final Path dir;
     private final Logger log = Logger.getLogger("org.openjdk.skara.vcs.git");
     private Path cachedRoot = null;
     private static final Hash EMPTY_TREE = new Hash("4b825dc642cb6eb9a060e54bf8d69288fbee4904");
 
+    public static void ignoreUserConfiguration(boolean ignore) {
```
edvbld, line 53: I would probably have named the method `ignoreConfiguration` and skipped the `boolean ignore` parameter. It doesn't seem like we ever pass `false` to `ignoreUserConfiguration`. If we need it in then in the future then we can add a method like `enableConfiguration`.

edvbld (approved):
Looks good!

PR #1050, opened 2021-03-11
912: PR title correction logic can get stuck

magicus:
The PR title correct logic introduced with SKARA-572 can end up trying to correct the title forever. An observed case is where the JBS title ends with a trailing space. When trying to update the PR title to include this trailing space, it will be stripped by GitLab and no update will actually be done. This will then get retried again and again. 

The fix is to strip the JBS title before doing any comparisons.

---------
### Progress

### Issue
 * [SKARA-912](https://bugs.openjdk.java.net/browse/SKARA-912): PR title correction logic can get stuck

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/1050/head:pull/1050`
`$ git checkout pull/1050`

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
@@ -170,11 +170,11 @@ class CheckWorkItem extends PullRequestWorkItem {
 
             var issue = project.issue(id);
             if (issue.isPresent()) {
-                var issueTitle = issue.get().title();
+                var issueTitle = issue.get().title().strip();
                 if (title.isEmpty()) {
                     // If the title is in the form of "[project-]<bugid>" only
                     // we add the title from JBS
-                    var newPrTitle = id + ": " + issue.get().title();
+                    var newPrTitle = id + ": " + issueTitle;
                     pr.setTitle(newPrTitle);
                     return true;
                 } else {
@@ -184,9 +184,9 @@ class CheckWorkItem extends PullRequestWorkItem {
                         title = title.substring(0, title.length() - 1);
                     }
                     if (issueTitle.startsWith(title) && issueTitle.length() > title.length()) {
-                        var newPrTitle = id + ": " + issue.get().title();
+                        var newPrTitle = id + ": " + issueTitle;
                         pr.setTitle(newPrTitle);
-                        var remainingTitle = issue.get().title().substring(title.length());
+                        var remainingTitle = issueTitle.substring(title.length());
                         if (pr.body().startsWith(ELLIPSIS + remainingTitle + "\n\n")) {
                             // Remove remaning title, plus decorations
                             var newPrBody = pr.body().substring(remainingTitle.length() + 3);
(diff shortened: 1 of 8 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -170,11 +170,11 @@ private boolean updateTitle() {
 
             var issue = project.issue(id);
             if (issue.isPresent()) {
-                var issueTitle = issue.get().title();
+                var issueTitle = issue.get().title().strip();
```
rwestberg, line 173: The title() method is called in a lot of places, perhaps it would be better to perform the stripping closer to the source?

magicus: I'm not sure what you mean by that? We did use `issue.get().title()` in lots of places, but I replaces them with `issueTitle `to only have a single strip. That's as close to the source I can see possible.

bots/pr/src/test/java/org/openjdk/skara/bots/pr/CheckTests.java
```diff
@@ -932,6 +933,21 @@ void issueTitleCutOff(TestInfo testInfo) throws IOException {
...
+            var editHash2 = CheckableRepository.appendAndCommit(localRepo);
+            localRepo.push(editHash2, author.url(), "edit", true);
+
+            var prCutOff2 =  credentials.createPullRequest(author, "master", "edit", issue2.id() + ": My second issue ending in space", List.of(), false);
+
+            // Check the status
+            TestBotRunner.runPeriodicItems(checkBot);
+
+            // The PR title should contain the issue title without trailing space
+            assertEquals("TEST-1: My second issue ending in space", prCutOff2.title());
```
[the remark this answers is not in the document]
magicus: Yeah, my bad... i commented out the test before while developing to speed up the re-run time. Didn't realize this changed the behavior when bringing it back. Will fix.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
rwestberg: I was thinking of usages outside of this particular function, perhaps like https://github.com/openjdk/skara/blob/dbef34b8c9785ec185a70365324063a02b1ec010/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java#L256 - we probably don't want trailing tabs or spaces there either.. So instead of searching for more of these, my suggestion was to add a call to strip before returning from the title() method to avoid similar problems.

magicus: Seems a bit more risky since this can have effects on lots of other code. But sure, I fix it. I did try to look through the existing calls to title() and at least make a quick check that it seemed okay to strip the title.

rwestberg: Thanks! Yeah if this change breaks anything, I think it's better that we find out now rather than worry about it. :)

rwestberg (approved):
Looks good!

PR #1053, opened 2021-03-12
Improve the Download message in PR description

magicus:
Both @simonis and I have struggled of finding the proper way to update a checked-out PR. Let's update the help text with Volker's findings of the best way to do this!

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/1053/head:pull/1053`
`$ git checkout pull/1053`

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2021, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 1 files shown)


edvbld:
You mean when you are updating your local copy of someone else's pull request? Because otherwise you just pull from your personal fork, right?

magicus:
If you follow the instructions in the Download section to get a local copy of the PR, this works just fine. However, if you want to update your local copy when the PR has changed with new code, this turned out to be a non-trivial exercise in gitology for both me and Volker (and many more, I presume). I think this is understood from the context of the "Download" part.

magicus:
See this thread https://mail.openjdk.java.net/pipermail/skara-dev/2021-March/004322.html

edvbld:
Ok, I see what you want to do. My below instructions are not appropriate to add to the pull request body (due to the complexity of the commands), but the most ergonomic way to set this up is to just add a remote and add an additional fetch command:

1. `$ git remote add upstream https://git.openjdk.java.net/skara`
2. `$ git config --add remote.upstream.fetch '+refs/pull/*/head:refs/remotes/upstream/pull/*`

Then whenever you want to checkout a pull request locally you can just run

```
$ git fetch upstream # in case you haven't fetched in a while
$ git checkout pull/1053
```

The above will ensure that the local branch `pull/1053` tracks `refs/remotes/upstream/pull/1053` which in turn is updated from `refs/pull/1053/head` whenever you run `git fetch upstream` :smile: Again, a little bit _too_ long to add this information to pull request body, so I think we just go with your changes in this PR. Could potentially be something for an FAQ or the developer's guide?

edvbld (approved):
Looks good

simonis:
Thanks @magicus  for updating the download message and thanks @edvbld for shedding some light on how to properly define an upstream for pull requests :) If I understand you correct, with that setup a `git fetch upstream` will fetch *all* pull requests from upstream. Is that correct?

edvbld:
@simonis yep, that is correct. As usual when you fetch refs from a remote they will only becomes refs under `refs/remote/<remote-name>/<branch-name>`, so you won't see all the refs when running e.g. `git branch` (you would have to run `git branch -a`).

simonis:
@edvbld thanks for the confirmation. I'll definitely put that line into my `.git/config` (if only to not forget the syntax :)

PR #1056, opened 2021-03-15
pr: add /tag commit command

edvbld:
Hi all,

please review this patch that adds the `/tag` commit command. As the name implies, the command is used to create a tag with the given name referring to the commit the command is applied to. The command will check that the tag isn't already used and that it conforms to an eventual `.jcheck/conf` configuration.

Thanks,
Erik

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/1056/head:pull/1056`
`$ git checkout pull/1056`


rwestberg (approved):
Looks good!

kevinrushforth:
Will a project be able to configure the role / review requirements for adding a tag? For example, I can imagine that some projects might want to allow a Reviewer to add a tag without needing review approval from another Reviewer.

kevinrushforth:
> So my question is basically the opposite: are there projects where tagging *should* require a review procedure?

Probably not.

PR #1057, opened 2021-03-15
Add support for using ssh cloning with GitLab

rwestberg:
Add support for configuring a GitLab forge to use ssh for cloning repositories using a private ssh key.

---------
### Progress

### Reviewers
 * [Magnus Ihse Bursie](https://openjdk.java.net/census#ihse) (@magicus - Committer)

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/1057/head:pull/1057`
`$ git checkout pull/1057`

### DIFF
--- a/forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabForgeFactory.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabForgeFactory.java
@@ -27,9 +27,13 @@ import org.openjdk.skara.host.Credential;
 import org.openjdk.skara.json.JSONObject;
 import org.openjdk.skara.json.JSONValue;
 
+import java.io.*;
 import java.net.URI;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.*;
 import java.util.HashSet;
 import java.util.Set;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 public class GitLabForgeFactory implements ForgeFactory {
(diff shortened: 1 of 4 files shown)


forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabForgeFactory.java
```diff
@@ -43,6 +47,33 @@ public String name() {
...
+        var filtered = existingMatcher.replaceAll("");
+        var result = "Match host=" + userHost + "\n" +
+                "  Hostname " + hostName + "\n" +
+                "  PreferredAuthentications publickey\n" +
+                "  StrictHostKeyChecking no\n" +
+                "  IdentityFile " + sshKey + "\n" +
+                "\n";
+
+        try {
+            Files.writeString(cfgFile, result + filtered.strip() + "\n", StandardCharsets.UTF_8);
```
magicus, line 71: Can this result in us writing to a users .ssh/config when running Skara CLI tools?

forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabForgeFactory.java
```diff
@@ -43,6 +47,33 @@ public String name() {
...
+        try {
+            existing = Files.readString(cfgFile, StandardCharsets.UTF_8);
+        } catch (IOException ignored) {
+        }
+
+        var userHost = userName + "." + hostName;
+        var existingBlock = Pattern.compile("^Match host=" + Pattern.quote(userHost) + "(?:\\R[ \\t]+.*)+", Pattern.MULTILINE);
+        var existingMatcher = existingBlock.matcher(existing);
+        var filtered = existingMatcher.replaceAll("");
+        var result = "Match host=" + userHost + "\n" +
```
magicus, line 63: I don't think this is valid .ssh/config syntax. Afaik it should be `Match <keyword> <value>`, that is, no `=`. I see no tests in this PR. Are you sure this works?

forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabForgeFactory.java
rwestberg: No, CLI tools never pass a configuration object when creating a Forge.

magicus: Good.

forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabForgeFactory.java
rwestberg: Yes, tested it manually and seems to work fine. And no, there are no automated tests that exercise GitLab or GitHub APIs, except for running them with a special credential configuration..

magicus: I still think this does not follow the documented ssh config file syntax (https://man7.org/linux/man-pages/man5/ssh_config.5.html). And even if that works on your machine right now, it can break on a future update of ssh.

rwestberg: > Configuration options may be separated by whitespace or optional whitespace and exactly one ‘=’

Still, works without it as well, so can certainly change it..

magicus (approved):
LGTM

edvbld (approved):
Looks good!

PR #1058, opened 2021-03-16
vcs: add ReadOnlyRepository.commitMetadataFor

edvbld:
Hi all,

please review this patch that adds the method `ReadOnlyRepository.commitMetadataFor`. I opted to *not* implement the Mercurial version due to the way `HgRepository` tries to mimic Git branches using bookmarks. The Skara code that uses `HgRepository` has never needed to use branches, so the the current bookmark implementation is more of a proof-of-concept. The implementation of the `ReadOnlyRepository.commitMetadataFor` for `HgRepository` would look something like the following:

```java
@Override
public List<CommitMetadata> commitMetadataFor(List<Branch> branches) throws IOException {
    var cmd = new ArrayList<String>();
    cmd.addAll(List.of("hg", "log", "--template", HgCommitMetadata.TEMPLATE));
    for (var branch : branches) {
        cmd.add("--rev");
        cmd.add("0:" + branch.name());
    }
    return readMetadata(cmd);
}
```

The above will get messy however if `branches` contains a proper branch (e.g. `default`) and not just a bookmark. This can be solved, but I don't think it is worth it right now. If we ever need to write a tool interacting with Mercurial repositories that needs `commitMetadataFor` then we can implement the method then.

Please also note that I had to name the method `commitMetadataFor` instead of just overloading `commitMetadata` since we already have `commitMetadata(List<Path>)`.

Thanks,
Erik

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/1058/head:pull/1058`
`$ git checkout pull/1058`

### DIFF
--- a/jcheck/src/test/java/org/openjdk/skara/jcheck/TestRepository.java
+++ b/jcheck/src/test/java/org/openjdk/skara/jcheck/TestRepository.java
@@ -185,6 +185,10 @@ class TestRepository implements ReadOnlyRepository {
         return List.of();
     }
 
+    public List<CommitMetadata> commitMetadataFor(List<Branch> branches) throws IOException {
+        return List.of();
+    }
+
     public Path root() throws IOException {
         return null;
     }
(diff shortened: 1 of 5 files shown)


rwestberg (approved):
Sounds reasonable!

PR #1061, opened 2021-03-16
vcs: make RepositoryTests deterministic

edvbld:
Hi all,

please review this small patch to make the test `RepositoryTests.testCommitMetadataWithBranchesWithGit` deterministic. The problem is that different versions of `git` can traverse the commit graph differently and the resulting list from the call to `commitMetadataFor` might therefore contain commits in an unknown order. The test should therefore not assume that the commits are ordered in a particular way, but instead ensure that the list contains only the expected commits.

Thanks,
Erik

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/1061/head:pull/1061`
`$ git checkout pull/1061`


rwestberg (approved):
Looks good!

PR #1066, opened 2021-03-16
Really fix so tests ignore .gitconfig

magicus:
It turned out that [SKARA-913](https://bugs.openjdk.java.net/browse/SKARA-913) which fixed unwanted side effects of [SKARA-868](https://bugs.openjdk.java.net/browse/SKARA-868), in fact broke Skara testing completely the way SKARA-868 was supposed to fix. In SKARA-913 only a few of the Skara tests were properly configured to really ignore the local configuration; something I discovered lately when running all tests on a machine with a .gitconfig that foils several git tests from running.

Since there is no really single point of entry to say to the Repository logic "I am now running a test", I have created a wrapper TestableRepository.init(), which calls Repository.init() and makes sure ignoreConfiguration is on.

At least this solves the problem for me. There might be other tests hiding where other strange configurations provoke a failure; I know of no systematic way to find all these, so I'll guess we have to fix that when it is discovered.

### DIFF
--- a/bots/checkout/src/test/java/org/openjdk/skara/bots/checkout/CheckoutBotTests.java
+++ b/bots/checkout/src/test/java/org/openjdk/skara/bots/checkout/CheckoutBotTests.java
@@ -22,11 +22,8 @@
  */
 package org.openjdk.skara.bots.checkout;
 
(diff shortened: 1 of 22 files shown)


edvbld (approved):
Looks good!

PR #1076, opened 2021-03-18
cli: make it possible to configure forge name

edvbld:
Hi all,

please review this patch that makes it possible to configure the forge name for a repository using `git config forge.name`, for example `git config forge.name github`. This can be useful for speeding up the forge selection logic in `Forge.from` so that the code doesn't have to query multiple candidate forges.

Testing:

Thanks,
Erik

---------
### Progress

### Reviewers

### Download
To checkout this PR locally:
`$ git fetch https://git.openjdk.java.net/skara pull/1076/head:pull/1076`
`$ git checkout pull/1076`

To update a local copy of the PR:
`$ git checkout pull/1076`
`$ git pull https://git.openjdk.java.net/skara pull/1076/head`

### DIFF
--- a/cli/src/main/java/org/openjdk/skara/cli/ForgeUtils.java
+++ b/cli/src/main/java/org/openjdk/skara/cli/ForgeUtils.java
@@ -41,6 +41,19 @@ public class ForgeUtils {
         System.exit(1);
     }
 
+    private static void gitConfig(String key, String value) {
+        try {
+            var pb = new ProcessBuilder("git", "config", key, value);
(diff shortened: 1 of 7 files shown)


rwestberg:
Looks good in general, but how about also storing this configuration automatically as well when auto-detection completes successfully (if it's not configured yet)? That way everyone could benefit from it on subsequent runs without having to configure it manually.

edvbld:
> Looks good in general, but how about also storing this configuration automatically as well when auto-detection completes successfully (if it's not configured yet)?

Good idea! See the latest commit for an implementation :smile:

rwestberg (approved):
Looks good!

PR #1104, opened 2021-03-31
pr: do not fetch tags for backport command

edvbld:
Hi all,

please review this patch that makes the `/backport` commit command *not* fetch tags. The `/backport` command doesn't need the tags and different openjdk repositories can have slightly different hashes for the annotated tag object itself (not the commit the tag is referring to).

I also updated the help text from the bot to instruct the user to fetch with `--no-tags`.

Thanks,
Erik


rwestberg (approved):
Looks good!

PR #1110, opened 2021-04-08
Backport security level should be inherited from main issue

erikj79:
This is a change Robin prepared, but we need to deploy it ASAP to match the new behavior in the backport plugin after the Jira upgrade. The security level of a backport should simply be based on the security level of the primary bug.

### DIFF
--- a/bots/notify/src/test/java/org/openjdk/skara/bots/notify/issue/IssueNotifierTests.java
+++ b/bots/notify/src/test/java/org/openjdk/skara/bots/notify/issue/IssueNotifierTests.java
@@ -1028,6 +1028,7 @@ public class IssueNotifierTests {
                                                  Map.of("issuetype", JSON.of("Enhancement"),
                                                         "customfield_10008", JSON.of("java.io")
                                                  ));
+            var level = issue.properties().get("security");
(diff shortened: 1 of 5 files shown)


bots/notify/src/test/java/org/openjdk/skara/bots/notify/issue/IssueNotifierTests.java
```diff
@@ -1061,6 +1062,11 @@ void testIssueBackport(TestInfo testInfo) throws IOException {
 
             // Labels should not
             assertEquals(0, backport.labelNames().size());
+
+            // If the parent issue has a security level (can be configured when running a test manually) it should be propagated
+            if (level != null) {
+                assertEquals(level.asString(), backport.properties().get("security").asString());
+            }
```
kevinrushforth, line 1069: Maybe also worth also checking that if level is null, the returned property is empty? This could be done as a follow-up given the need to get this in right away.

kevinrushforth (approved):
Looks good to me (although I am not a Skara project Reviewer).

erikj79:
The tests failed because they were run while JBS was down. There are apparently tests needing to touch JBS in there. I have rerun locally after JBS came up and it's all green.

PR #1120, opened 2021-04-15
SKARA-971: Stop clumping up logs in logstash

erikj79:
I would like to change how we send logs to logstash so each log message is stored as an individual document. To make this more practical, we need to add some more meta data to each message so we can properly filter them. I suggest adding the following:

instance_start_time: An ISO_INSTANT time stamp string indicating when this JVM instance was started.
work_id: A running counter generated ID for each work item, rest request or periodic check, which was previously clumped up in one log message.
work_item: For work items, store the item toString()

---------
### Progress

### Issue
 * [SKARA-971](https://bugs.openjdk.java.net/browse/SKARA-971): Stop clumping up logs in logstash

### Reviewers
 * [Robin Westberg](https://openjdk.java.net/census#rwestberg) (@rwestberg - **Reviewer**) ⚠️ Review applies to f1ec8ffd7dfe9d930d994a7e1ee7c1f5d9038fcd

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1120/head:pull/1120` \
`$ git checkout pull/1120`

`$ git checkout pull/1120` \
`$ git pull https://git.openjdk.java.net/skara pull/1120/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1120`

`$ git pr show -t 1120`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1120.diff">https://git.openjdk.java.net/skara/pull/1120.diff</a>

</details>

### DIFF
--- a/bots/cli/src/main/java/org/openjdk/skara/bots/cli/BotLogstashHandler.java
+++ b/bots/cli/src/main/java/org/openjdk/skara/bots/cli/BotLogstashHandler.java
@@ -23,22 +23,27 @@
 package org.openjdk.skara.bots.cli;
 
 import org.openjdk.skara.bot.LogContextMap;
-import org.openjdk.skara.network.RestRequest;
 import org.openjdk.skara.json.JSON;
 
-import java.io.*;
 import java.net.URI;
+import java.net.http.*;
 import java.time.*;
 import java.time.format.DateTimeFormatter;
 import java.util.*;
+import java.util.concurrent.Future;
 import java.util.logging.*;
 import java.util.regex.Pattern;
 
+/**
+ * Handles logging to logstash. Be careful not to call anything that creates new
+ * log records from this class as that can cause infinite recursion.
+ */
 public class BotLogstashHandler extends StreamHandler {
-    private final RestRequest endpoint;
+    private final URI endpoint;
+    private final HttpClient httpClient;
     private final DateTimeFormatter dateTimeFormatter;
-    private final Logger log = Logger.getLogger("org.openjdk.skara.bots.cli");
-
+    // Optionally store all futures for testing purposes
+    private Collection<Future<HttpResponse<Void>>> futures;
 
     private static class ExtraField {
         String name;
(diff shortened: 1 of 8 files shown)


bots/cli/src/main/java/org/openjdk/skara/bots/cli/BotLogstashHandler.java
```diff
@@ -81,6 +78,10 @@ private void publishToLogstash(Instant time, Level level, String message, Map<St
             query.put("level_value", level.intValue());
             query.put("message", message);
 
+            for (Map.Entry<String, String> entry : LogContextMap.entrySet()) {
```
rwestberg, line 81:
```suggestion
            for (var entry : LogContextMap.entrySet()) {
```

rwestberg (approved):
Looks good, this is probably how it should have been done from the beginning!

bots/cli/src/main/java/org/openjdk/skara/bots/cli/BotLogstashHandler.java
```diff
@@ -78,7 +78,7 @@ private void publishToLogstash(Instant time, Level level, String message, Map<St
             query.put("level_value", level.intValue());
             query.put("message", message);
 
-            for (Map.Entry<String, String> entry : LogContextMap.entrySet()) {
+            for (var entry : LogContextMap.entrySet()) {
```
erikj79, line 81: Right, not used to var yet. I use Intellij's templates to generate loops and the template generates the type for me. I did change it manually in some other location but missed this one.

bots/cli/src/main/java/org/openjdk/skara/bots/cli/BotLogstashHandler.java
```diff
@@ -33,12 +33,10 @@
 import java.util.*;
 import java.util.logging.*;
 import java.util.regex.Pattern;
-import java.util.stream.Collectors;
 
-public class BotLogstashHandler extends BotTaskAggregationHandler {
+public class BotLogstashHandler extends StreamHandler {
     private final RestRequest endpoint;
```
rwestberg, line 38: One thing that I realized when thinking about this, is that this change will result in a lot more POSTs. However, RestRequest implements various rate limits to adhere to GitHub policies like https://docs.github.com/en/rest/overview/resources-in-the-rest-api#rate-limiting and https://docs.github.com/en/rest/guides/best-practices-for-integrators#dealing-with-abuse-rate-limits . So it would perhaps be better to just use HttpClient directly in this context.

PR #1136, opened 2021-04-22
983: Github commit links 404

erikj79:
Sometimes our abbreviated commit web links on Github are not working. My guess is that this is caused by collisions. I think the correct fix here is to stop abbreviating commit hashes in links. Github internal links seem to always be full hashes, while pretty printing the link text with an abbreviation. We should adopt the same basic principle.

---------
### Progress

### Issue
 * [SKARA-983](https://bugs.openjdk.java.net/browse/SKARA-983): Github commit links 404

### Reviewers
 * [Tim Bell](https://openjdk.java.net/census#tbell) (@tbell29552 - **Reviewer**) ⚠️ Review applies to 430499747d61c7e1f8fed34f6d3b9a9881abfd56
 * [Robin Westberg](https://openjdk.java.net/census#rwestberg) (@rwestberg - **Reviewer**) ⚠️ Review applies to 430499747d61c7e1f8fed34f6d3b9a9881abfd56

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1136/head:pull/1136` \
`$ git checkout pull/1136`

`$ git checkout pull/1136` \
`$ git pull https://git.openjdk.java.net/skara pull/1136/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1136`

`$ git pr show -t 1136`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1136.diff">https://git.openjdk.java.net/skara/pull/1136.diff</a>

</details>

### DIFF
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
@@ -257,9 +257,11 @@ class IssueNotifier implements Notifier, PullRequestListener, RepositoryListener
 
                 var existingComments = issue.comments();
                 var hashUrl = repository.webUrl(commit.hash()).toString();
+                // We used to store URLs with just the abbreviated hash, so need to check for both
+                var shortHashUrl = repository.webUrl(new Hash(commit.hash().abbreviate())).toString();
                 var alreadyPostedComment = existingComments.stream()
-                                                           .filter(comment -> comment.author().equals(issueProject.issueTracker().currentUser()))
-                                                           .anyMatch(comment -> comment.body().contains(hashUrl));
+                        .filter(comment -> comment.author().equals(issueProject.issueTracker().currentUser()))
+                        .anyMatch(comment -> comment.body().contains(hashUrl) || comment.body().contains(shortHashUrl));
                 if (!alreadyPostedComment) {
                     issue.addComment(commitNotification);
                 }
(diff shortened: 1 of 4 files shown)


tbell29552 (approved):

kevinrushforth:
Conceptually, this looks like the right fix to me. A couple questions:

1. Do all places that generate a commit URL use the `webUrl` method?
2. Are there any places that extract the printed name from the URL (if so, those would now print the full 40 char hash)?
3. Are there place other than `IssueNotifer::onNewCommits` where you might need to check for old versus new hashes? I think this would only be a problem if there were other similar places that checked whether the URL had already been seen.
4. Are there any tests that need to be updated (or written)?

rwestberg (approved):
Looks good! Perhaps you have run the tests locally, but couldn't hurt to enable the GitHub Actions to make sure there isn't some odd dependency on the short format left somewhere. :)

erikj79:
> 1. Do all places that generate a commit URL use the `webUrl` method?
From what I can tell, yes.

> 2. Are there any places that extract the printed name from the URL (if so, those would now print the full 40 char hash)?
In bug comments, we typically print the full URL and let Jira interpret it as a link (with the full URL as link text). In other places, like in Jira links, we already generate our own link text, and Github/Gitlab seems to recognize these links and generate their own link text for them.

> 3. Are there place other than `IssueNotifer::onNewCommits` where you might need to check for old versus new hashes? I think this would only be a problem if there were other similar places that checked whether the URL had already been seen.
I searched for all uses of the webUrl method to look at what it was used for. This was the only usage that wasn't just adding the link to some kind of message.

> 4. Are there any tests that need to be updated (or written)?
All tests ran successfully so nothing need changing. Perhaps I should add a test case for the parsing of existing URLs in comments.

Erik D asked me to run this on staging for a bit (which will affect reviews in the Skara project) so we can see how it works.

edvbld (approved):
Looks good

kevinrushforth (approved):

PR #1148, opened 2021-05-11
metrics: add library

edvbld:
Hi all,

please review this patch that adds the new `metrics` library. The metrics library can be used to collect runtime metrics, such as the number of HTTP requests or the number of bot runs. The library is designed according to the [Prometheus library guide](https://prometheus.io/docs/instrumenting/writing_clientlibs/) and currently features the two Prometheus types [gauge](https://prometheus.io/docs/concepts/metric_types/#gauge) and [counter](https://prometheus.io/docs/concepts/metric_types/#counter). The library is designed to allow it to be extended with the [histogram](https://prometheus.io/docs/concepts/metric_types/#histogram) and [summary](https://prometheus.io/docs/concepts/metric_types/#summary) types if we find a need for them.

The `Counter` and `Gauge` types are designed to be thread-safe and should typically be stored in a `private final static` variable in a classes (similar to how one might store a `Logger` instance). A minimal `Counter` example looks like:

```java
class Example {
    private static final Counter counter = Counter.name("test").register();
}
```

I won't go into details here on how Prometheus naming and labelling works, please see the Prometheus [documentation](https://prometheus.io/docs/introduction/overview/) for those concepts. The goal is that library should be fairly self-explanatory for someone with an understanding of how Prometheus works. Note though that the `metrics` library itself is not tied to Prometheus in any way, it only supports the collection of runtime metrics. It is however designed to make it easy to write a Prometheus exporter, something I will show in a follow-up patch. 

I also added a bunch of unit tests for the new library.

Thanks,
Erik

---------
### Progress

### Reviewers
 * [Erik Joelsson](https://openjdk.java.net/census#erikj) (@erikj79 - **Reviewer**) ⚠️ Review applies to 38e0ee00dc5fc02647dbf0b1396504b8f18d661a

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1148/head:pull/1148` \
`$ git checkout pull/1148`

`$ git checkout pull/1148` \
`$ git pull https://git.openjdk.java.net/skara pull/1148/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1148`

`$ git pr show -t 1148`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1148.diff">https://git.openjdk.java.net/skara/pull/1148.diff</a>

</details>


edvbld:
Thanks @erikj79 for reviewing!

/integrate

PR #1167, opened 2021-05-25
1052: Memory leak in RestRequestCache

erikj79:
This patch adds a regularly scheduled cache eviction call to the RestRequestCache.

---------
### Progress

### Issue
 * [SKARA-1052](https://bugs.openjdk.java.net/browse/SKARA-1052): Memory leak in RestRequestCache

### Reviewers
 * [Kevin Rushforth](https://openjdk.java.net/census#kcr) (@kevinrushforth - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1167/head:pull/1167` \
`$ git checkout pull/1167`

`$ git checkout pull/1167` \
`$ git pull https://git.openjdk.java.net/skara pull/1167/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1167`

`$ git pr show -t 1167`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1167.diff">https://git.openjdk.java.net/skara/pull/1167.diff</a>

</details>

### DIFF
--- a/bot/src/main/java/org/openjdk/skara/bot/BotRunnerConfiguration.java
+++ b/bot/src/main/java/org/openjdk/skara/bot/BotRunnerConfiguration.java
@@ -320,6 +320,20 @@ public class BotRunnerConfiguration {
     }
 
     /**
+     * The amount of time to wait between runs of the RestResponseCache evictions.
+     * @return
+     */
+    Duration cacheEvictionInterval() {
+        if (!config.contains("runner") || !config.get("runner").contains("cache_eviction_interval")) {
+            var defaultValue = Duration.ofMinutes(5);
+            log.info("No cache eviction interval defined, using default value " + defaultValue);
+            return defaultValue;
+        } else {
+            return Duration.parse(config.get("runner").get("cache_eviction_interval").asString());
+        }
+    }
+
+    /**
      * Number of WorkItems to execute in parallel.
      * @return
      */
--- a/bot/src/main/java/org/openjdk/skara/bot/BotRunner.java
+++ b/bot/src/main/java/org/openjdk/skara/bot/BotRunner.java
@@ -36,6 +36,7 @@ import java.util.logging.*;
 import java.util.stream.Collectors;
 
 import com.sun.net.httpserver.*;
+import org.openjdk.skara.network.RestRequest;
 
 class BotRunnerError extends RuntimeException {
     BotRunnerError(String msg) {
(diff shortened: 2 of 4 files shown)


bot/src/main/java/org/openjdk/skara/bot/BotRunnerConfiguration.java
```diff
@@ -319,6 +319,20 @@ Duration scheduledExecutionPeriod() {
         }
     }
 
+    /**
+     * The amount of time to wait between runs of the RestResponseCache evictions.
+     * @return
+     */
+    Duration cacheEvictionInterval() {
+        if (!config.contains("runner") || !config.get("runner").contains("cache_eviction_interval")) {
+            var defaultValue = Duration.ofMinutes(5);
```
kevinrushforth, line 328: Have you done any testing to see if a default interval of 5 minutes is the right frequency for cache evictions? Naively, it seems like it might be too often. Is there a performance impact?

kevinrushforth (approved):
Looks fine, although I raised a question about the default frequency.

bot/src/main/java/org/openjdk/skara/bot/BotRunnerConfiguration.java
edvbld: I agree with Kevin, I _think_ we could probably keep at least 60 minutes of cache, if not even more. Or I might have missed something 🤷😄

edvbld (approved):
Looks good

bot/src/main/java/org/openjdk/skara/bot/BotRunnerConfiguration.java
erikj79: That is a good point. In my initial patch I used the scheduling frequency (which is what the watchdog uses) instead of this new configuration value, which is 10 seconds, and it worked fine on staging. I thought that was a bit excessive so I increased to a default of 5 mins, which matches the lowest "maxAge" in the cache.

While we could certainly live with a longer interval here, 24h would still solve our problems, I don't think we have much to gain from increasing it longer than my suggested 5 mins. The nature of the bots is that they run pretty much non stop, using a lot of cpu. I can't see how running this little method will have any noticeable impact on performance, even if the cache has thousands or even millions of valid entries. It shouldn't take more than milliseconds on one of 8-32 threads (depending on bot runner config).

If we start seeing problems, this value can always be changed in the future, or configured for any specific bot runner configuration.

kevinrushforth: Seems OK to me.

erikj79: Also just to clarify as I realized there was a misunderstanding in offline discussion. This patch is not changing the functional behavior of the cache in any way. The entries that are being removed are just those older than "maxAge", which were already disqualified in the send method.

I have also verified the behavior in staging over night and the metrics show that we have the same amount of cache hits/min with the patch as before.

PR #1180, opened 2021-06-02
1063: Lower log severity on HTTP errors

erikj79:
The Skara bots currently log all HTTP errors as SEVERE. With the log configuration we use, this results in admin alarms being sent whenever this happens (with some throttling). Now that we have metrics, we should instead leverage those and rely on alarms from our metrics gathering system instead. Typically HTTP errors aren't severe on their own, but if we get a lot of them, across multiple bot runners, then we need to react.

This patch does the following:

1. Introduces a new specific exception which the RestRequest class throws when a bad HTTP response code is received.
2. Changes the logging behavior in the BotRunner when this new exception is caught.
3. Improves the logging of bad RestRequests so we actually log the request (instead of just the class name of the RequestBuilder).
4. Adds a new label on the skara_http_response counter to track if an error code was handled or not. We have a defined way of handling a bad HTTP response, and if this was used, we shouldn't send alarms as it's expected behavior.
5. Applies the http counters in two places where they were currently missing.

---------
### Progress

### Issue
 * [SKARA-1063](https://bugs.openjdk.java.net/browse/SKARA-1063): Lower log severity on HTTP errors

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1180/head:pull/1180` \
`$ git checkout pull/1180`

`$ git checkout pull/1180` \
`$ git pull https://git.openjdk.java.net/skara pull/1180/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1180`

`$ git pr show -t 1180`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1180.diff">https://git.openjdk.java.net/skara/pull/1180.diff</a>

</details>

### DIFF
--- a/bot/src/main/java/org/openjdk/skara/bot/BotRunner.java
+++ b/bot/src/main/java/org/openjdk/skara/bot/BotRunner.java
@@ -39,6 +39,7 @@ import com.sun.management.ThreadMXBean;
 
 import com.sun.net.httpserver.*;
 import org.openjdk.skara.network.RestRequest;
+import org.openjdk.skara.network.UncheckedRestException;
 
 class BotRunnerError extends RuntimeException {
     BotRunnerError(String msg) {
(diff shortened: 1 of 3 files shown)


kevinrushforth (approved):
Looks good.

PR #1182, opened 2021-06-04
1070: An invalid /backport command target causes bot to fail endlessly

erikj79:
If a user issues a /backport command on a commit on a repository that exists, but for which we haven't properly configured backports to work (they need a fork defined where the PR can be created from), we currently end up with endless failures. This patch adds handling of the situation by printing a helpful message for the user, which also includes a list of all valid repositories.

---------
### Progress

### Issue
 * [SKARA-1070](https://bugs.openjdk.java.net/browse/SKARA-1070): An invalid /backport command target causes bot to fail endlessly

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1182/head:pull/1182` \
`$ git checkout pull/1182`

`$ git checkout pull/1182` \
`$ git pull https://git.openjdk.java.net/skara pull/1182/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1182`

`$ git pr show -t 1182`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1182.diff">https://git.openjdk.java.net/skara/pull/1182.diff</a>

</details>

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/BackportCommand.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/BackportCommand.java
@@ -89,7 +89,9 @@ public class BackportCommand implements CommandHandler {
 
         var potentialTargetRepo = forge.repository(repoName);
         if (potentialTargetRepo.isEmpty()) {
-            reply.println("@" + username + " the target repository `" + repoName + "` does not exist");
+            reply.println("@" + username + " the target repository `" + repoName + "` does not exist. ");
+            reply.print("List of valid repositories: ");
+            reply.println(String.join(", ", bot.forkRepoNames()));
             return;
         }
         var targetRepo = potentialTargetRepo.get();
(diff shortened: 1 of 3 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/BackportCommand.java
```diff
@@ -104,7 +106,14 @@ public void handle(PullRequestBot bot, HostedCommit commit, CensusInstance censu
 
         try {
             var hash = commit.hash();
-            var fork = bot.writeableForkOf(targetRepo);
+            var optionalFork = bot.writeableForkOf(targetRepo);
+            if (optionalFork.isEmpty()) {
+                reply.print("@" + username + " [" + repoName + "](" + targetRepo.webUrl() + ") is not a valid target for backports. ");
+                reply.print("List of valid repositories: ");
+                reply.println(String.join(", ", bot.forkRepoNames()));
```
kevinrushforth, line 113: Should this be further filtered to only list repos that have the same "base" (share some commits) as the source repo? For example, if we some day configure one of the `jfx` or `jmc` update release repos to allow backport APIs, you wouldn't want to target it from a commit in the `jdk` repo, so listing those repos could be confusing. This could be done as a follow-up, since it is only a hypothetical problem for now.

kevinrushforth (approved):
Looks good. I made a suggestion (for a follow-up).

bots/pr/src/main/java/org/openjdk/skara/bots/pr/BackportCommand.java
erikj79: I agree that it would be much better to filter the list of repos, but I'm not sure how to easily implement this. I figured this list was easily available and better than nothing.

PR #1183, opened 2021-06-07
SKARA-1069

erikj79:
This change adds a new transaction step when integrating a PR, to better handle if the bot gets interrupted mid integration. Currently, if the change is pushed and the bot is interrupted before closing, changing labels or adding the final "commit pushed" comment, the PR can end up in a limbo state.

The new step is another comment "Going to push commit as ..." which gets added right before the git push command is run. Using this comment, it's now possible to automatically recover if the bot gets interrupted. The integration command checks for any such comments and if found, checks if the commit hash is present in the target. If it is, the PR was already pushed, and the command will just close it out as normal.

I also decided to move the output of any rebase command to this prepush comment, so we don't risk losing it.

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/IntegrateCommand.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/IntegrateCommand.java
@@ -40,6 +40,8 @@ import java.util.regex.Matcher;
 public class IntegrateCommand implements CommandHandler {
     private final Logger log = Logger.getLogger("org.openjdk.skara.bots.pr");
     private static final Pattern BACKPORT_PATTERN = Pattern.compile("<!-- backport ([0-9a-z]{40}) -->");
+    private static final String PRE_PUSH_MARKER = "<!-- prepush %s -->";
+    private static final Pattern PRE_PUSH_PATTERN = Pattern.compile("<!-- prepush ([0-9a-z]{40}) -->");
 
     private void showHelp(PrintWriter reply) {
         reply.println("usage: `/integrate [auto|manual|<hash>]`");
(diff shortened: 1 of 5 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/IntegrateCommand.java
```diff
@@ -124,7 +159,6 @@ public void handle(PullRequestBot bot, PullRequest pr, CensusInstance censusInst
         }
 
         // Run a final jcheck to ensure the change has been properly reviewed
-        var success = false;
```
erikj79, line 127: This variable was dead.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/IntegrateCommand.java
```diff
@@ -159,8 +193,7 @@ public void handle(PullRequestBot bot, PullRequest pr, CensusInstance censusInst
                 return;
             }
 
-            var botUser = pr.repository().forge().currentUser();
-            var backportLines = pr.comments()
+            var backportLines = allComments
```
erikj79, line 196: This is a slight optimization. We already have a list of all comments, so no need to reach out to github to fetch them again. I can't see any need in this situation to refresh either.

forge/src/main/java/org/openjdk/skara/forge/PullRequestUtils.java
```diff
@@ -163,6 +163,11 @@ public static Repository materialize(HostedRepositoryPool hostedRepositoryPool,
         return localRepo;
     }
 
+    public static boolean isAncestorOfTarget(Repository localRepo, Hash hash) throws IOException {
```
erikj79, line 166: I put this in the utils class because it needs the "prutils_tagetref" reference, which is currently only used in this class.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/IntegrateCommand.java
kevinrushforth: So there's no need to ever refresh after acquiring the integration lock?

bots/pr/src/main/java/org/openjdk/skara/bots/pr/IntegrateCommand.java
```diff
@@ -196,20 +229,9 @@ public void handle(PullRequestBot bot, PullRequest pr, CensusInstance censusInst
             // Rebase and push it!
             if (!localHash.equals(PullRequestUtils.targetHash(pr, localRepo))) {
                 var amendedHash = checkablePr.amendManualReviewers(localHash, censusInstance.namespace(), original);
+                addPrePushComment(pr, amendedHash, rebaseMessage.toString());
                 localRepo.push(amendedHash, pr.repository().url(), pr.targetRef());
```
kevinrushforth, line 233: If the push fails and the bot dies, will it still be retried after this fix?

kevinrushforth (approved):
It looks OK to me. I added a couple inline questions.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/IntegrateCommand.java
erikj79: Not sure, I don't think the IntegrationLock is relevant for this, as it's not involved in the CheckWorkItem where this comment would potentially be created, but maybe I should still leave this alone as I'm not sure about how the concurrency could play out here.

Investigating this made me discover that there is a duplicate implementation of this whole thing in the SponsorCommand that I need to address.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/IntegrateCommand.java
erikj79: Yes, that's the idea at least, but I should write a test for it. If the push fails, the next time around, the prepush comment is found, but the hash is not found in the target, so the IntegrationCommand will continue as if nothing has happened.

erikj79:
Adding new patch addressing review comments. Reverted the allComments change. Added testing of interruption between posting the prepush comment and doing the push.

The big change here is I discovered the SponsorCommand had a lot of duplicate code that needed to be updated in the same way. The Skara codebase seem to shun inheritance to share logic, so I tried to just move some things to static methods on IntegrateCommand and have SponsorCommand call them when suitable. There are more duplicates that could be given the same treatment, but I tried to limit the changes here to what was relevant for this fix. I'm thinking a super class could help a lot here in a future change.

I also duplicated the main test I wrote from IntegrateTests to SponsorTests to verify the same behaviors since we have two almost identical code paths for the same thing.

kevinrushforth (approved):
Looks good as far as I can tell.

PR #1184, opened 2021-06-07
1072: Skara should validate the commit hash of a Backport PR

erikj79:
Add a check for "Backport <hash>" titles that the given hash isn't the head of the PR itself. If the user mistakenly uses that hash, Skara is very likely to get fooled into thinking the backport is clean.

While investigating this, I also discovered that the CheckWorkItem was inconsistent with not repeating "backport" errors, so I implemented a general mechanism for this, which guarantees that the same exact error message is never repeated. I believe this will be good enough, and certainly better than the existing behavior.

---------
### Progress

### Issue
 * [SKARA-1072](https://bugs.openjdk.java.net/browse/SKARA-1072): Skara should validate the commit hash of a Backport PR

### Reviewers
 * [Kevin Rushforth](https://openjdk.java.net/census#kcr) (@kevinrushforth - **Reviewer**) ⚠️ Review applies to f0cfcfff9be706dabdb4fb3094515916fe909ca8

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1184/head:pull/1184` \
`$ git checkout pull/1184`

`$ git checkout pull/1184` \
`$ git pull https://git.openjdk.java.net/skara pull/1184/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1184`

`$ git pr show -t 1184`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1184.diff">https://git.openjdk.java.net/skara/pull/1184.diff</a>

</details>

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
@@ -27,6 +27,7 @@ import org.openjdk.skara.forge.*;
 import org.openjdk.skara.host.HostUser;
 import org.openjdk.skara.issuetracker.Comment;
 import org.openjdk.skara.vcs.Hash;
+import org.openjdk.skara.vcs.Repository;
 import org.openjdk.skara.vcs.openjdk.CommitMessageParsers;
 
 import java.io.*;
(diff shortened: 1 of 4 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -229,6 +229,14 @@ public String toString() {
             var m = BACKPORT_TITLE_PATTERN.matcher(pr.title());
             if (m.matches()) {
                 var hash = new Hash(m.group(1));
+                if (pr.headHash().equals(hash)) {
```
kevinrushforth, line 232: Isn't it the hash of the first commit in the PR, rather than the HEAD commit of the PR, that is likely to cause the problem? In case there is more than one commit in the PR, might it be better to check whether it is an ancestor of the HEAD commit?

erikj79: Right, if there are multiple changes in the PR branch, any one of them could potentially be the original change. Checking that is a bit more complicated though as I think it requires a local clone to inspect. I will need to investigate how best to solve it.

erikj79:
Added test as asked.

PR #1185, opened 2021-06-10
Refactor and cleanup

erikj79:
I have been working quite a bit in the area of the Integrate and SponsorCommand classes recently, and while doing so noted that they have a lot of duplicate code. Some of that I fixed, but I didn't want to include unrelated refactoring and cleanup in a functional change, so here is a separate change with just that. Nothing should change from a functional perspective. This change only touches a limited set of classes where I had noted specific issues. Here is a breakdown of the changes.

1. Moved logic for finding the original change hash in a backport PR to the CheckablePullRequest class. All three locations where this was done already had an instance of this class, and that class has all the data needed for the operation, so seemed like a decent fit.
2. Put a lazily initialized wrapper of PullRequestUtils.targetHash(localRepo) in CheckablePullRequest. This method is called quite often, and each time it results in running a git command to resolve the hash from the ref name. When browsing logs, this git command is sprinkled all over the place so reducing it will improve both performance and log browsing. Unfortunately I can't figure out a good way to completely reduce it as PullRequestUtils is a static class and it calls this method quite a bit itself.
3. Moved a duplicate piece of code (for running and reacting to jcheck checks) in IntegrateCommand.handle() and SponsorCommand.handle() to a new static method in IntegrateCommand. I still didn't dare dabbling in a super class. 
4. Removed all dead parameters and variables that I could find in these classes.
5. Applied some other Intellij suggested code fixes, like removing exceptions never thrown and unneeded calls to toString(). I also let it convert a switch statements to the new fancy switch expression, which looks a lot better.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1185/head:pull/1185` \
`$ git checkout pull/1185`

`$ git checkout pull/1185` \
`$ git pull https://git.openjdk.java.net/skara pull/1185/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1185`

`$ git pr show -t 1185`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1185.diff">https://git.openjdk.java.net/skara/pull/1185.diff</a>

</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ external Git source code hosting providers are available:
 
 ## Building
 
-[JDK 15](http://jdk.java.net/15/) or later and [Gradle](https://gradle.org/)
+[JDK 14](http://jdk.java.net/14/) or later and [Gradle](https://gradle.org/)
 6.6 or later is required for building. To build the project on macOS or
 GNU/Linux x64, just run the following command from the source tree root:
 
@@ -69,7 +69,7 @@ also want to build the bot images run `sh gradlew images` on GNU/Linux or
 
 If you want to build on an operating system other than GNU/Linux, macOS or
 Windows _or_ if you want to build on a CPU architecture other than x64, then
-ensure that you have JDK 15 or later installed locally and JAVA_HOME set to
+ensure that you have JDK 14 or later installed locally and JAVA_HOME set to
 point to it. You can then run the following command from the source tree root:
 
 ```bash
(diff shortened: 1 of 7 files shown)


erikj79:
I reverted the text block change and the bump of the required JDK version to 15. If we are to do that, it should be in a proper change.

kevinrushforth (approved):

PR #1194, opened 2021-07-06
1084: Automatically update repository labels from Skara config

erikj79:
This patch adds a new WorkItem to the MailingListBridgeBot. The new LabelsUpdaterWorkItem runs once at startup and makes sure all mailing list labels configured for the repository actually exist as labels in the hosted repository (in GitHub or GitLab). I chose to implement this as a WorkItem, even though it's just run once. It will also block any other WorkItems in the bot from running until successful. By using the existing framework for this, we get automatic retries and I believe it will play better with other bots for resource usage, even though it doesn't quite fit in.

The new WorkItem also keeps the description field updated with the mailing list email address. This use of the description field was initially added manually for the main jdk repository on GitHub, but has not been kept up to date since, and has not been propagated to other repositories (jdk16 and jdk17). These description fields are both informative for the user, as well as being used by the CLI client to present information to users.

To verify the new interaction for managing labels on a repository with both GitLab and GitHub, I have implemented a new type of manual tests. They are disabled from automatic running in Gradle using the @Disabled annotation. To run them you need to provide a properties file with data about account and keys to use to access the servers. We don't want these kinds of tests to run against GitHub/GitLab automatically, but at least now there is a way to semi automatically verify that kind of functionality.

---------
### Progress

### Issue
 * [SKARA-1084](https://bugs.openjdk.java.net/browse/SKARA-1084): Automatically update repository labels from Skara config

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1194/head:pull/1194` \
`$ git checkout pull/1194`

`$ git checkout pull/1194` \
`$ git pull https://git.openjdk.java.net/skara pull/1194/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1194`

`$ git pr show -t 1194`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1194.diff">https://git.openjdk.java.net/skara/pull/1194.diff</a>

</details>

### DIFF
--- a/bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveWorkItem.java
+++ b/bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveWorkItem.java
@@ -61,14 +61,19 @@ class ArchiveWorkItem implements WorkItem {
 
     @Override
     public boolean concurrentWith(WorkItem other) {
-        if (!(other instanceof ArchiveWorkItem)) {
+        if (!(other instanceof ArchiveWorkItem otherArchiveItem)) {
+            if (!(other instanceof LabelsUpdaterWorkItem otherLabelsUpdaterItem)) {
+                return true;
+            }
+            if (!bot.equals(otherLabelsUpdaterItem.bot())) {
+                return true;
+            }
+            return false;
+        }
+        if (!pr.id().equals(otherArchiveItem.pr.id())) {
             return true;
         }
-        ArchiveWorkItem otherItem = (ArchiveWorkItem)other;
-        if (!pr.id().equals(otherItem.pr.id())) {
-            return true;
-        }
-        if (!bot.codeRepo().name().equals(otherItem.bot.codeRepo().name())) {
+        if (!bot.codeRepo().name().equals(otherArchiveItem.bot.codeRepo().name())) {
             return true;
         }
         return false;
--- a/forge/src/main/java/org/openjdk/skara/forge/github/GitHubRepository.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/github/GitHubRepository.java
@@ -582,4 +582,37 @@ public class GitHubRepository implements HostedRepository {
                       .map(o -> new Label(o.get("name").asString(), o.get("description").asString()))
                       .collect(Collectors.toList());
     }
+
+    @Override
+    public void addLabel(Label label) {
+        var params = JSON.object()
+                .put("name", label.name())
+                // Color is Gray and matches all current labels
+                .put("color", "ededed");
+        if (label.description().isPresent()) {
+            params.put("description", label.description().get());
+        }
+        request.post("labels")
+                .body(params)
+                .execute();
+    }
+
+    @Override
+    public void updateLabel(Label label) {
+        var params = JSON.object();
+        if (label.description().isPresent()) {
+            params.put("description", label.description().get());
+        } else {
+            params.put("description", JSONValue.fromNull());
+        }
+        request.post("labels/" + label.name())
+                .body(params)
+                .execute();
+    }
+
+    @Override
+    public void deleteLabel(Label label) {
+        request.delete("labels/" + label.name())
+                .execute();
+    }
 }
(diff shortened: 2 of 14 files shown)


bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveWorkItem.java
```diff
@@ -61,14 +61,19 @@ public String toString() {
 
     @Override
     public boolean concurrentWith(WorkItem other) {
-        if (!(other instanceof ArchiveWorkItem)) {
-            return true;
+        if (!(other instanceof ArchiveWorkItem otherArchiveItem)) {
+            if (!(other instanceof LabelsUpdaterWorkItem otherLabelsUpdaterItem)) {
```
kevinrushforth, line 65: Pattern-matching `instanceof` was finalized in JDK 16. Has Skara already bumped its minimum JDK to 16?

forge/src/main/java/org/openjdk/skara/forge/github/GitHubRepository.java
```diff
@@ -582,4 +582,37 @@ public void restrictPushAccess(Branch branch, HostUser user) {
                       .map(o -> new Label(o.get("name").asString(), o.get("description").asString()))
                       .collect(Collectors.toList());
     }
+
+    @Override
+    public void addLabel(Label label) {
+        var params = JSON.object()
+                .put("name", label.name())
+                // Color is Gray and matches all current labels
+                .put("color", "ededed");
```
kevinrushforth, line 591: Will this overwrite the existing color scheme? For a new repo, will it always use gray for all labels?

kevinrushforth (commented):
Not a complete review, but I left a couple questions inline.

forge/src/main/java/org/openjdk/skara/forge/github/GitHubRepository.java
erikj79: If addLabel() is called on an existing label, then I think the color scheme will be overwritten. The intention not to change any label, just to add the ones that are missing. The updateLabel method will not change the color.

Note that the labels we are adding with this patch are just the ones for mailing lists. Do you think we need to add automation for colors for the other set of labels we use (e.g. rfr, ready, csr etc)?

bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveWorkItem.java
erikj79: We are running the Skara bots on JDK 16, but officially we are still listing JDK 14 as minimum. I don't think it's worth bumping the minimum for this.

kevinrushforth: In that case, you will need to use explicit casts here rather than the pattern-matching `instanceof`.

forge/src/main/java/org/openjdk/skara/forge/github/GitHubRepository.java
kevinrushforth: I think if we do add automation for colors for the other set of labels, that could be done as a follow-on fix (not sure it's needed though). I mainly wanted to make sure we wouldn't overwrite the once for rfr, ready, etc. with gray.

bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveWorkItem.java
erikj79: Doh, now I understand what you mean. I must have my intellij configured incorrectly and let it just rewrite this to pattern matching style without thinking about it. Then I was out sick for a week and forgot about it.

erikj79: I changed my settings and tried compiling with JDK 14 and discovered that we have already have several JDK 15 and 16 dependencies that have crept in. Given that, I think I will just bump the required version to 16 (as a separate change). Skara has always been close to the bleeding edge anyway.

kevinrushforth (approved):
Looks good.

PR #1202, opened 2021-07-15
1058: The mlbridge bot occasionally posts the same comments twice on Github

erikj79:
This patch (hopefully) fixes an issue where sometimes the mlbridge bot would post the same email as comment twice in a PR. This happens because two instances of the CommentPosterWorkItem are running at the same time for the same PullRequest.

This is supposed to be protected by the method "concurrentWith(WorkItem other)". In this particular case, the check is faulty. It checks if the PullRequest object in this WorkItem is equal to the object in the other WorkItem. This is default Object equality, and there is definitely no guarantee that they both contain the exact same object.

Looking around in other implementations of this method, there is mix of how stringent the checks are, some only check if the PullRequest.id() field is the same, but most also check something about the HostedRepository to which the PR belongs (either name or url). I think calculating the url seems a bit expensive for an equals check.

I propose a new method on the PullRequest interface: "isSame(PullRequest other)". This isn't trying to be an equals, as that's a bit hard to define (and I don't want to get into implementing hashCode at this time), but will just return true if both PullRequest instances are logically referring to the same hosted pull request. I have replaced all uses of PullRequest.equals() and PullRequest.id().equals() with this new method for consistency.

---------
### Progress

### Issue
 * [SKARA-1058](https://bugs.openjdk.java.net/browse/SKARA-1058): The mlbridge bot occasionally posts the same comments twice on Github

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1202/head:pull/1202` \
`$ git checkout pull/1202`

`$ git checkout pull/1202` \
`$ git pull https://git.openjdk.java.net/skara pull/1202/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1202`

`$ git pr show -t 1202`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1202.diff">https://git.openjdk.java.net/skara/pull/1202.diff</a>

</details>

### DIFF
--- a/bots/bridgekeeper/src/main/java/org/openjdk/skara/bots/bridgekeeper/PullRequestCloserBot.java
+++ b/bots/bridgekeeper/src/main/java/org/openjdk/skara/bots/bridgekeeper/PullRequestCloserBot.java
@@ -86,10 +86,7 @@ class PullRequestCloserBotWorkItem implements WorkItem {
             return true;
         }
         PullRequestCloserBotWorkItem otherItem = (PullRequestCloserBotWorkItem)other;
-        if (!pr.id().equals(otherItem.pr.id())) {
-            return true;
-        }
-        if (!repository.name().equals(otherItem.repository.name())) {
+        if (!pr.isSame(otherItem.pr)) {
             return true;
         }
         return false;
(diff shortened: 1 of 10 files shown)


kevinrushforth (approved):
Looks good.

PR #1203, opened 2021-07-15
1068: [Backport] Bot finds original commit in wrong repository

erikj79:
When the PR bot searches for the original commit for a backport it's more or less random from which repository it will be found. Whatever it finds is posted as a link in a PR comment, and if it points to something weird, that can be confusing (though not strictly incorrect as the change will be the exact same regardless of which repo it is in).

The search feature already has what looks to be a configurable way to limit searching to a set of organizations. Unfortunately, that feature was never used and is currently broken for Github (for Gitlab it's already in use as it's required for the search method we use). Because of this, we usually pick a match from adoptium or SAP today. This patch fixes this for Github (where the query separator is space, not '+'), and I intend to limit the organizations to "openjdk" when deploying this fix.

In addition to this, it would be nice if the exact repo that gets picked is the "correct" one. Unfortunately, there is no really good definition of what the correct repo would be in all cases, and it would be even trickier to implement this in a good way. As an approximate workaround, I've added sorting on repository name length, so we always pick the repo with the shortest name. I believe this will at least give us a better answer than "random" in most cases for a relatively modest cost in performance.

I have experimented with this patch in staging with the playground repo and the behavior I've seen so far is good.

---------
### Progress

### Issue
 * [SKARA-1068](https://bugs.openjdk.java.net/browse/SKARA-1068): [Backport] Bot finds original commit in wrong repository

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1203/head:pull/1203` \
`$ git checkout pull/1203`

`$ git checkout pull/1203` \
`$ git pull https://git.openjdk.java.net/skara pull/1203/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1203`

`$ git pr show -t 1203`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1203.diff">https://git.openjdk.java.net/skara/pull/1203.diff</a>

</details>

### DIFF
--- a/forge/src/main/java/org/openjdk/skara/forge/github/GitHubHost.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/github/GitHubHost.java
@@ -360,8 +360,13 @@ public class GitHubHost implements Forge {
         if (items.isEmpty()) {
             return Optional.empty();
         }
-        var first = items.get(0);
-        var repo = repository(first.get("repository").get("full_name").asString());
-        return repo.get().commit(hash);
+        // When searching for a commit, there may be hits in multiple repositories.
+        // There is no good way of knowing for sure which repository we would rather
+        // get the commit from, but a reasonable default is to go by the shortest
+        // name as that is most likely the main repository of the project.
+        var shortestName = items.stream()
+                .map(o -> o.get("repository").get("full_name").asString())
+                .min(Comparator.comparing(String::length));
+        return shortestName.flatMap(this::repository).flatMap(r -> r.commit(hash));
     }
 }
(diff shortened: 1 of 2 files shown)


kevinrushforth (approved):
It looks good. I presume you have tested it? Especially since there aren't any new unit tests added.

erikj79:
> It looks good. I presume you have tested it? Especially since there aren't any new unit tests added.

I have tested it manually in the playground repo by deploying to staging yes. I could have written new manual junit tests I suppose. If I think I have time for that today I will, but otherwise I would rather get these changes live this week.

PR #1211, opened 2021-08-30
1145: NPE in GitHubPullRequest::reviews

erikj79:
The core of this change is to fix an NPE in GitHubPullRequest::reviews, which happens when a commit that a review refers to no longer exists (presumably because of a rebase or forced update of the review branch). The introduction of a possible null value in Review::hash has some ripple effects however, so this gets a bit more complicated.

I've opted to wrap the return value of Review::hash and ReviewComment::hash with Optional to properly force every caller to handle the situation. I haven't seen this happen for ReviewComment, but I would be surprised if the same thing didn't apply there from the GitHub side.

---------
### Progress

### Issue
 * [SKARA-1145](https://bugs.openjdk.java.net/browse/SKARA-1145): NPE in GitHubPullRequest::reviews

### Reviewers
 * [Magnus Ihse Bursie](https://openjdk.java.net/census#ihse) (@magicus - **Reviewer**) ⚠️ Review applies to dcbc4f1cb0c672f0d9735f81d8661acff4cb9b97

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1211/head:pull/1211` \
`$ git checkout pull/1211`

`$ git checkout pull/1211` \
`$ git pull https://git.openjdk.java.net/skara pull/1211/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1211`

`$ git pr show -t 1211`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1211.diff">https://git.openjdk.java.net/skara/pull/1211.diff</a>

</details>

### DIFF
--- a/forge/src/main/java/org/openjdk/skara/forge/Review.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/Review.java
@@ -58,7 +58,7 @@ public class Review {
     }
 
     /**
-     * The hash for the commit for which this review was created. Can be null if the commit
+     * The hash for the commit for which this review was created. Can be empty if the commit
      * no longer exists.
      */
     public Optional<Hash> hash() {
--- a/forge/src/main/java/org/openjdk/skara/forge/Review.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/Review.java
@@ -57,8 +57,12 @@ public class Review {
         return verdict;
     }
 
-    public Hash hash() {
-        return hash;
+    /**
+     * The hash for the commit for which this review was created. Can be null if the commit
+     * no longer exists.
+     */
+    public Optional<Hash> hash() {
+        return Optional.ofNullable(hash);
     }
 
     public int id() {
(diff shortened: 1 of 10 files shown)


forge/src/main/java/org/openjdk/skara/forge/Review.java
```diff
@@ -57,8 +57,12 @@ public Verdict verdict() {
         return verdict;
     }
 
-    public Hash hash() {
-        return hash;
+    /**
+     * The hash for the commit for which this review was created. Can be null if the commit
+     * no longer exists.
```
magicus, line 62: I think technically the Optional is empty, rather than "null".

forge/src/main/java/org/openjdk/skara/forge/ReviewComment.java
```diff
@@ -50,8 +50,12 @@ public ReviewComment(ReviewComment parent, String threadId, Hash hash, String pa
         return Optional.ofNullable(parent);
     }
 
-    public Hash hash() {
-        return hash;
+    /**
+     * The hash for the commit for which this review comment was created. Can be null if the commit
+     * no longer exists.
```
magicus, line 55: Same here...

magicus (approved):

forge/src/main/java/org/openjdk/skara/forge/Review.java
erikj79: Right, in my mind I was documenting the state of the field rather than the return value of the method. I can change it.

PR #1214, opened 2021-09-02
1153: Some emails still not posted as comments in PR

erikj79:
After fixing SKARA-1148, there were still some emails that weren't being posted as comments in PRs. The underlying cause this time was emails that Skara thinks it has sent, but they never were, so they weren't present in the mail archive. Emails sent in response to such missing emails have bad In-Reply-To headers as they point to emails that aren't in the archive. When the Mbox parser tries to piece together conversations, it fails to find the correct parents in these cases.

The reason we need to piece conversations together is to find the original RFR email, which contains the pull request link. This is how each email is associated with the correct pull request.

The fix for this is to also look at the References header as a backup. The References header contains the whole ancestor chain of email IDs, so using that, we can connect conversations regardless of missing emails in between, especially since we only really care about the first email in the conversation. Looking at the pipermail thread view, this is what it must be doing as it would otherwise fail miserably. I also tweaked the loops here to hopefully do a bit less work, which I think is worth it due to the pretty large number of emails being parsed here.

In addition to this small change to Mbox, I'm also tinkering with some other things that are related. These were things I stumbled over and needed to properly debug and test the issue. 

As I noted in SKARA-1148 already, the MailingListArchiveReaderBot is doing a lot of redundant work. I blame this on SKARA-843, where this bot was changed from one global instance to running one instance for each configured repository. There were two problems introduced with this change.

1. Each instance still has all the configured repositories, so every found conversation is evaluated against every repository in every instance. This creates an unnecessary N^2 complexity in number of configured repositories.
2. Each instance has its own MailingListReader, configured for exactly the set of mailing lists used for the repository. We have a lot of repositories that share the exact same mailing list config. This means that each of these instances will read all the archives for themselves, with no sharing of this data. The reader already does caching, so after the first time around, it's much faster. Still, the first time around we read jdk-updates-dev a pretty large number of times.

My solution for 1 is to only have one single repository in each MailingListArchiveReaderBot. This looks like a simple oversight in the previous patch. 

For 2, I make sure to only create one MailingListReader for each unique set of mailing lists. This will not remove all the redundant reads, but it will bring them down significantly. I also think it's the functionally correct solution as we will then only consider email threads on the relevant mailing lists for each repository, so less unnecessary (and potentially bad) cross evaluation between mailing lists and repositories that aren't configured as related.

To get 2 to actually work correctly, I needed to tweak the logic that protects us from running the wrong WorkItems concurrently. To be able to share MailingListReader between multiple WorkItems, we have WorkItem::concurrentWith. Unfortunately, this method was also used to detect duplicate WorkItems in the pending pool of the scheduler. This is correct in most cases, but not here. An ArchiveReaderWorkItem for jdk11u and jdk15u shares MailingListReader, but queueing one should not replace the other. To solve this I added another method WorkItem::replaces for this particular check (with a default method to preserve the current behavior for all other WorkItems).

Finally I also changed ArchiveReaderWorkItem::toString to print the repository name rather than the list of mailing lists. This made a lot more sense to me while debugging as we now create an instance per repository rather than based on mailinglists.

---------
### Progress

### Issue
 * [SKARA-1153](https://bugs.openjdk.java.net/browse/SKARA-1153): Some emails still not posted as comments in PR

### Reviewers
 * [Kevin Rushforth](https://openjdk.java.net/census#kcr) (@kevinrushforth - **Reviewer**) ⚠️ Review applies to fda3fbb94e7e0cbcee2c250ffbc740e8fb5a41c1

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1214/head:pull/1214` \
`$ git checkout pull/1214`

`$ git checkout pull/1214` \
`$ git pull https://git.openjdk.java.net/skara pull/1214/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1214`

`$ git pr show -t 1214`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1214.diff">https://git.openjdk.java.net/skara/pull/1214.diff</a>

</details>

### DIFF
--- a/bot/src/main/java/org/openjdk/skara/bot/BotRunner.java
+++ b/bot/src/main/java/org/openjdk/skara/bot/BotRunner.java
@@ -251,7 +251,7 @@ public class BotRunner {
 
                     for (var pendingItem : pending.entrySet()) {
                         // If there are pending items of the same type that we cannot run concurrently with, replace them.
-                        if (pendingItem.getKey().getClass().equals(item.getClass()) && !pendingItem.getKey().concurrentWith(item)) {
+                        if (item.replaces(pendingItem.getKey())) {
                             log.finer("Discarding obsoleted item " + pendingItem.getKey() +
                                               " in favor of item " + item);
                             DISCARDED_COUNTER.labels(item.botName(), item.workItemName()).inc();
--- a/mailinglist/src/main/java/org/openjdk/skara/mailinglist/Mbox.java
+++ b/mailinglist/src/main/java/org/openjdk/skara/mailinglist/Mbox.java
@@ -97,9 +97,6 @@ public class Mbox {
             unhandledEmails = new ArrayList<>();
 
             for (var email : emailsToCheck) {
-                if (email.id().toString().contains("<VtLlAA6xMBK-6Ib-4SbkYZJssOuyqPl5Dr7kdiJUf6A=.9c549bc6-867f-4ba6-8727-efafe7cc18e2@github.com>")) {
-                    log.warning("Found first email of jdk#5300 in mbox");
-                }
                 if (!idToConversation.containsKey(email.id())) {
                     EmailAddress inReplyTo = findInReplyTo(idToMail, email);
                     if (inReplyTo != null) {
--- a/mailinglist/src/main/java/org/openjdk/skara/mailinglist/Mbox.java
+++ b/mailinglist/src/main/java/org/openjdk/skara/mailinglist/Mbox.java
@@ -89,43 +89,70 @@ public class Mbox {
                                        .filter(email -> !email.hasHeader("In-Reply-To"))
                                        .collect(Collectors.toMap(Email::id, Conversation::new));
 
-        var outOfOrder = new ArrayList<Email>();
-        var lastOutOfOrderCount = -1;
-        while (outOfOrder.size() != lastOutOfOrderCount) {
-            lastOutOfOrderCount = outOfOrder.size();
-            outOfOrder.clear();
+        var unhandledEmails = emails;
+        var lastUnhandledCount = 0;
+        while (unhandledEmails.size() != lastUnhandledCount) {
+            lastUnhandledCount = unhandledEmails.size();
+            var emailsToCheck = unhandledEmails;
+            unhandledEmails = new ArrayList<>();
 
-            for (var email : emails) {
-                if (email.hasHeader("In-Reply-To")) {
-                    var inReplyToMatcher = inReplyToPattern.matcher(email.headerValue("In-Reply-To"));
-                    if (!inReplyToMatcher.find()) {
-                        log.info("Cannot parse In-Reply-To header: " + email.headerValue("In-Reply-To"));
-                        continue;
-                    }
-                    var inReplyTo = EmailAddress.from(inReplyToMatcher.group(1));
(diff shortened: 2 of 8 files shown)


bot/src/main/java/org/openjdk/skara/bot/BotRunner.java
```diff
@@ -251,7 +251,7 @@ private void submitOrSchedule(WorkItem item) {
 
                     for (var pendingItem : pending.entrySet()) {
                         // If there are pending items of the same type that we cannot run concurrently with, replace them.
-                        if (pendingItem.getKey().getClass().equals(item.getClass()) && !pendingItem.getKey().concurrentWith(item)) {
+                        if (item.replaces(pendingItem.getKey())) {
```
kevinrushforth, line 254: I presume that `a.concurrentWith(b) == b.concurrentWith(a)` for all work items (meaning that `concurrentWith` is a commutative operation)? The reason I ask is that the old code checked `pendingItem.getKey().concurrentWith(item)` while the new code checks `item.concurrentWith(pendingItem.getKey()))`. I expect it is fine.

mailinglist/src/main/java/org/openjdk/skara/mailinglist/Mbox.java
```diff
@@ -89,45 +89,72 @@ private static String decodeFromStrings(String body) {
...
+        var unhandledEmails = emails;
+        var lastUnhandledCount = 0;
+        while (unhandledEmails.size() != lastUnhandledCount) {
+            lastUnhandledCount = unhandledEmails.size();
+            var emailsToCheck = unhandledEmails;
+            unhandledEmails = new ArrayList<>();
+
+            for (var email : emailsToCheck) {
+                if (email.id().toString().contains("<VtLlAA6xMBK-6Ib-4SbkYZJssOuyqPl5Dr7kdiJUf6A=.9c549bc6-867f-4ba6-8727-efafe7cc18e2@github.com>")) {
+                    log.warning("Found first email of jdk#5300 in mbox");
```
kevinrushforth, line 101: What is the purpose of this check?

kevinrushforth (approved):
This looks good as far as I can tell. I left a couple questions.

Is this something you can test before deploying?

erikj79:
> Is this something you can test before deploying?

With a local edit of the BotRunner, I have manually tested the MailingListArchiveReaderBot and ArchiveReaderWorkItem by running them with the full production configuration, but prevented any other WorkItems from actually running. I've then inspected the set of CommentPosterWorkItems that get scheduled.

bot/src/main/java/org/openjdk/skara/bot/BotRunner.java
erikj79: As far as I can tell, it's commutative. I flipped it around because it made more sense to me to ask "replaces?" that way, but who calls who really shouldn't matter in either case.

mailinglist/src/main/java/org/openjdk/skara/mailinglist/Mbox.java
erikj79: Ah, that's a left over debug message (and break point line) from my testing. Will remove.

kevinrushforth (approved):

erikj79:
Thanks Kevin!
/integrate

PR #1216, opened 2021-09-07
1150: Migrate skara tooling for client code to reference the new client-libs-dev list

erikj79:
Here is my attempt at trying to rewrite the rules for automatic labeling for the new "client" label and client-libs-dev mailing list. I took the combined contents of the previous 2d, awt, sound and swing labels, sorted the lines and then tried to manually merge as much as possible. It's possible I didn't get everything correct or at least not fully optimized.

---------
### Progress

### Issue
 * [SKARA-1150](https://bugs.openjdk.java.net/browse/SKARA-1150): Migrate skara tooling for client code to reference the new client-libs-dev list

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1216/head:pull/1216` \
`$ git checkout pull/1216`

`$ git checkout pull/1216` \
`$ git pull https://git.openjdk.java.net/skara pull/1216/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1216`

`$ git pr show -t 1216`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1216.diff">https://git.openjdk.java.net/skara/pull/1216.diff</a>

</details>


kevinrushforth:
@prrace should take a look at this.

It can probably be optimized further if we assume that everything under `src/java.desktop`, with the exception of the two directories that are part of `java.beans` should get the `client` label. This will pick up a couple directories (for accessibility) and a handful of other files that weren't included before, but I suspect that was probably an oversight.

With that assumption, I think this would be the minimal reduction:

```
src/java.desktop/aix
src/java.desktop/linux
src/java.desktop/macosx
src/java.desktop/share/classes/com/sun/accessibility
src/java.desktop/share/classes/com/sun/imageio
src/java.desktop/share/classes/com/sun/java
src/java.desktop/share/classes/com/sun/media
src/java.desktop/share/classes/com/sun/swing
src/java.desktop/share/classes/java/applet
src/java.desktop/share/classes/java/awt
src/java.desktop/share/classes/javax
src/java.desktop/share/classes/sun
src/java.desktop/share/conf
src/java.desktop/share/legal
src/java.desktop/share/native
src/java.desktop/unix
src/java.desktop/windows
```

And then you would probably add the 3 beans files that formerly had the `2d` label, plus the build and test files from your PR. Let's see what Phil says.

prrace:
> And then you would probably add the 3 beans files that formerly had the `2d` label, plus the build and test files from your PR. Let's see what Phil says.

Phil doesn't understand why beans is still separate.

In the JBS I wrote
"All subscribers to client mailing lists - those for the above groups and also the beans-dev list have been subscribed to the new client-libs-dev list."

So beans-dev is going away too .. so the entire java.desktop module should be mapped to client-libs-dev
It should be one trivial mapping except for a double-mapping  to client and build for the build related files
that reside in the desktop module .. although I don't know that I see them there in this change ?
Actually neither before, nor after, the change. Perhaps they moved after skara built this mapping.


You may also find it useful to look at what I put here : https://openjdk.java.net/groups/client-libs/

It enumerates all the source, demo, test, make/build files I could identify that are handled by client-libs
jdk.unsupported.desktop is there but unless I'm missing something is unmapped.
You can cross-reference the tests + demos between the json and that web page too.

erikj79:
Thanks for the input Phil and Kevin, I missed that beans was also included. I will simplify this further.

magicus (approved):

kevinrushforth (approved):

PR #1221, opened 2021-09-17
1079: Support ability to defer and delegate timing of an integration

erikj79:
Here is my first stab at implementing a way for a PR author to defer integration to another committer. The usecase for this is if the timing of the integration is important, but the author may not be available to issue the /integrate command at the desired time. 

Here is how it works: There are two new arguments to the /integrate command "defer" and "undefer". If the author of a PR issues "/integrate defer", the PR enters the deferred state (which is marked by the label "deferred"). This enables any committer to issue "/integrate" to actually perform the integration. When a PR is integrated by someone other than the author, then that person gets attributed in the Git committer field, just like when a change is sponsored. The author can revert this ability at any time before integration happens by issuing "/integrate undefer".

This is similar to how sponsoring a change works, but not quite the same. When setting up for being sponsored, the /integrate command will run all the checks and record the specific hash that the author wishes to commit. I skipped this kind of rigorous book keeping here as I think it is more likely to get in the way in the desired usecase. The whole point is that timing is important and the author is not available, so better to leave things more open in my opinion.

I'm open to informed suggestions on how this could potentially be improved on a conceptual level.

When writing the test, I based it on the simpleMerge test already present. It took some work to properly understand what was happening in that test, and to make it clearer (at least to me), I have renamed and reorganized things a little bit. In these tests, the HostedRepository instances ("author", "integrator", "reviewer" etc) represent users and are used to create PR test instances (or views) which the test can use to interact with the PR as different users. To make this clearer, I introduced the botUser explicitly as that is also a user that sometimes issues commands and interacts with a PR, and used that for things that the bot is doing. I also changed some other users around so that they perform the role for which they are named (reviewer reviews, not the integrator).

When composing the messages that the bot will print back in comments, I noticed that "PR" and "pull request" were used interchangeably. I decided to standardize on "pull request".

---------
### Progress

### Issue
 * [SKARA-1079](https://bugs.openjdk.java.net/browse/SKARA-1079): Support ability to defer and delegate timing of an integration

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1221/head:pull/1221` \
`$ git checkout pull/1221`

`$ git checkout pull/1221` \
`$ git pull https://git.openjdk.java.net/skara pull/1221/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1221`

`$ git pr show -t 1221`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1221.diff">https://git.openjdk.java.net/skara/pull/1221.diff</a>

</details>

### DIFF
--- a/bots/pr/src/test/java/org/openjdk/skara/bots/pr/IntegrateTests.java
+++ b/bots/pr/src/test/java/org/openjdk/skara/bots/pr/IntegrateTests.java
@@ -1357,8 +1357,7 @@ class IntegrateTests {
             assertEquals(1, undeferred, "Missing undeferred message");
             assertFalse(authorPr.labelNames().contains("deferred"));
 
-            // Try integrating as another committer, which should fails since the PR is currently not deferred
-            // Try to integrate by committer
+            // Try integrating as another committer, which should fail since the PR is currently not deferred
             var integratorPr = integrator.pullRequest(authorPr.id());
             integratorPr.addComment("/integrate");
             TestBotRunner.runPeriodicItems(mergeBot);
--- a/bots/pr/src/test/java/org/openjdk/skara/bots/pr/IntegrateTests.java
+++ b/bots/pr/src/test/java/org/openjdk/skara/bots/pr/IntegrateTests.java
@@ -1357,19 +1357,29 @@ class IntegrateTests {
             assertEquals(1, undeferred, "Missing undeferred message");
             assertFalse(authorPr.labelNames().contains("deferred"));
 
+            // Try integrating as another committer, which should fails since the PR is currently not deferred
+            // Try to integrate by committer
+            var integratorPr = integrator.pullRequest(authorPr.id());
+            integratorPr.addComment("/integrate");
+            TestBotRunner.runPeriodicItems(mergeBot);
+            var nonAuthor = authorPr.comments().stream()
+                    .filter(comment -> comment.body().contains("Only the author")
+                            && comment.body().contains("is allowed to issue the `integrate` command"))
+                    .count();
+            assertEquals(1, nonAuthor, "Missing only author can integrate message");
+
             // Defer again
             authorPr.addComment("/integrate defer");
             TestBotRunner.runPeriodicItems(mergeBot);
             assertTrue(authorPr.labelNames().contains("deferred"));
 
             // Try to issue /integrate with an invalid command for a non author
-            var integratorPr = integrator.pullRequest(authorPr.id());
             integratorPr.addComment("/integrate auto");
             TestBotRunner.runPeriodicItems(mergeBot);
             var invalid = authorPr.comments().stream()
                     .filter(comment -> comment.body().contains("Only the author"))
                     .count();
-            assertEquals(1, invalid, "Missing error message");
+            assertEquals(2, invalid, "Missing error message");
 
             // Try to integrate by committer
             integratorPr.addComment("/integrate");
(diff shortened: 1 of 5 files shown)


kevinrushforth:
I'll do a review of the implementation when I can. Conceptually, this seems like a good approach to me.

I wonder if there would be a benefit in having the ability to defer to a specific Committer (or maybe list of Committers)? As you say, one of the use cases might be when a specific timing is needed. Another case might be a coordination with another developer so that both yours and their PR go in at roughly the same time. In both of those cases, it seems more likely that I would want a specific delegate to be able to integrate on my behalf. I don't know if this additional flexibility is worth the complexity.

erikj79:
I was considering maintaining a list of valid integrators, but in the end I felt like it just added complexity to solve a problem I doubt we would ever really have. I do believe we can trust committers within the OpenJDK to not go around and issue /integrate in other people's PRs.

bots/pr/src/test/java/org/openjdk/skara/bots/pr/IntegrateTests.java
```diff
@@ -1296,4 +1295,103 @@ void retryAfterInterruptExtraChange(TestInfo testInfo) throws IOException {
...
+
+            // Issue /integrate undefer and verify the PR is no longer deferred
+            authorPr.addComment("/integrate undefer");
+            TestBotRunner.runPeriodicItems(mergeBot);
+            var undeferred = authorPr.comments().stream()
+                    .filter(comment -> comment.body().contains("Integration of this pull request is no longer deferred and may only be integrated by the author"))
+                    .count();
+            assertEquals(1, undeferred, "Missing undeferred message");
+            assertFalse(authorPr.labelNames().contains("deferred"));
+
```
kevinrushforth, line 1359: Is it worth attempting to `/integrate` as the integrator here (which should fail since you've set it back to manual)?

bots/pr/src/test/java/org/openjdk/skara/bots/pr/IntegrateTests.java
```diff
@@ -1296,4 +1295,103 @@ void retryAfterInterruptExtraChange(TestInfo testInfo) throws IOException {
...
+            TestBotRunner.runPeriodicItems(mergeBot);
+
+            // The change should now be present on the master branch
+            var pushedRepo = Repository.materialize(pushedFolder.path(), author.url(), "master");
+            assertTrue(CheckableRepository.hasBeenEdited(pushedRepo));
+
+            var headHash = pushedRepo.resolve("HEAD").orElseThrow();
+            var headCommit = pushedRepo.commits(headHash.hex() + "^.." + headHash.hex()).asList().get(0);
+
+            // Author and committer should be the same
```
kevinrushforth, line 1385: Not the same as each other.

bots/pr/src/test/java/org/openjdk/skara/bots/pr/IntegrateTests.java
```diff
@@ -1296,4 +1295,103 @@ void retryAfterInterruptExtraChange(TestInfo testInfo) throws IOException {
...
+            assertEquals("Generated Committer 1", headCommit.author().name());
+            assertEquals("integrationcommitter1@openjdk.java.net", headCommit.author().email());
+            assertEquals("Generated Committer 4", headCommit.committer().name());
+            assertEquals("integrationcommitter4@openjdk.java.net", headCommit.committer().email());
+            assertTrue(authorPr.labelNames().contains("integrated"));
+
+            // Ready label should have been removed
```
kevinrushforth, line 1392: Minor: `// Ready and defer labels ...`

kevinrushforth (approved):
The fix and test both look good. I added a couple minor comments on the test.

kevinrushforth (approved):

magicus (approved):
Looks good. I think this fits well with the current model. I agree that we do not need to limit the set of eligible future integrators when deferring.

PR #1231, opened 2021-10-05
1201: Multiple command reply messages include 'HostUserDetails'

magicus:
Fixing this as the bug reporter suggested, by removing all explicit (broken or not) and duplicated constructions of "@" + username in the command replies, and adding a general @-mention in CommitCommands as well as PR Commands.


kevinrushforth (approved):
Looks fine.

kevinrushforth:
I see a lot of unit test failures, which I hadn't noticed when I approved it. Have you run the tests locally to see what the results are?

PR #1233, opened 2021-10-11
885: jbsupdater -pool record consumption modifications

erikj79:
This patch changes how -pool records are consumed, to help differentiate between OpenJDK and OracleJDK update releases. OracleJDK update releases have moved to using fixVersions with an -oracle suffix. With this change, pool record consumption for such releases will also use an -oracle suffix. I opted for a very simple implementation, where any "-suffix" on a fixVersion is now required as suffix on a -pool version to be a match. I think this makes the most sense and keeps things simple going forward.

While in the area, I also removed logic for treating N-open fixVersions the same way as N-pool versions. I have no memory of seeing that behavior in action and I believe that implementation was based on a misunderstanding. I could be wrong however, so if anyone can think of a reason to keep that logic, please let me know.

---------
### Progress

### Issue
 * [SKARA-885](https://bugs.openjdk.java.net/browse/SKARA-885): jbsupdater -pool record consumption modifications

### Reviewers
 * [Kevin Rushforth](https://openjdk.java.net/census#kcr) (@kevinrushforth - **Reviewer**) ⚠️ Review applies to b59a7468f740e5dfada82db09556a79825b5d25c

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1233/head:pull/1233` \
`$ git checkout pull/1233`

`$ git checkout pull/1233` \
`$ git pull https://git.openjdk.java.net/skara pull/1233/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1233`

`$ git pr show -t 1233`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1233.diff">https://git.openjdk.java.net/skara/pull/1233.diff</a>

</details>

### DIFF
--- a/jbs/src/main/java/org/openjdk/skara/jbs/Backports.java
+++ b/jbs/src/main/java/org/openjdk/skara/jbs/Backports.java
@@ -119,23 +119,42 @@ public class Backports {
     }
 
     /**
-     * Return true if the issue's fixVersionList is a match for fixVersion, using "-pool[-<opt>]".
-     * If the fixVersion has an opt string, then the pool record needs to have the same opt
-     * string suffix.
-     *
-     * If fixVersion has a major release of <N>, it matches the fixVersionList has an
-     * <N>-pool entry and all other entries are scratch values.
+     * If fixVersion has a major release of <N>, and opt string of <opt> it matches if
+     * the fixVersionList has an <N>-pool-<opt> entry.
+     */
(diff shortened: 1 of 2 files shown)


erikj79:
To ease the deployment of this change the and transition that will follow, I decided to change the behavior and continue to fall back to regular N-pool records if no N-pool-<opt> record is found. With this change, we can deploy this PR as it is without affecting anything currently in use.

Later we may decide to require N-pool-<opt> records to avoid stealing N-pool records.

PR #1245, opened 2021-11-24
1071: auto-set `/csr` flag from JBS state

lgxbslgx:
Hi all,

Currently, the author of the PR or a reviewer could type `/csr` in the comment to direct the need of the CSR.
It is good for the bot to automatically scan the `scr for` link and set the `csr` label for the PR.

This patch adds the automatically scanning feature. And the corresponding tests are added.

Best Regards,
-- Guoxiong

---------
### Progress

### Issue
 * [SKARA-1071](https://bugs.openjdk.java.net/browse/SKARA-1071): auto-set `/csr` flag from JBS state

### Reviewers
 * [Erik Joelsson](https://openjdk.java.net/census#erikj) (@erikj79 - **Reviewer**) ⚠️ Review applies to 37573bddc05d83f67ecce37ac9c89bf892c9cf75

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1245/head:pull/1245` \
`$ git checkout pull/1245`

`$ git checkout pull/1245` \
`$ git pull https://git.openjdk.java.net/skara pull/1245/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1245`

`$ git pr show -t 1245`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1245.diff">https://git.openjdk.java.net/skara/pull/1245.diff</a>

</details>

### DIFF
--- a/bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
+++ b/bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
@@ -122,7 +122,13 @@ class CSRBot implements Bot, WorkItem {
                     }
 
                     if (!name.asString().equals("Approved")) {
-                        if (!pr.labelNames().contains(CSR_LABEL)) {
+                        if (name.asString().equals("Withdrawn")) {
+                            // This condition is necessary to prevent the bot from adding the CSR label again.
+                            // And the bot can't remove the CSR label automatically here.
+                            // Because the PR author with the role of Committer may withdraw a CSR that
+                            // a Reviewer had requested and integrate it without satisfying that requirement.
+                            log.info("CSR closed and withdrawn for " + describe(pr) + ", not removing CSR label");
+                        } else if (!pr.labelNames().contains(CSR_LABEL)) {
                             log.info("CSR issue resolution is not 'Approved' for " + describe(pr) + ", adding the CSR label");
                             pr.addLabel(CSR_LABEL);
                         } else {
--- a/bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
+++ b/bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
@@ -127,7 +127,7 @@ class CSRBot implements Bot, WorkItem {
                             // And the bot can't remove the CSR label automatically here.
                             // Because the PR author with the role of Committer may withdraw a CSR that
                             // a Reviewer had requested and integrate it without satisfying that requirement.
-                            log.info("CSR closed and withdrawn for " + describe(pr) + ", not removing CSR label");
+                            log.info("CSR closed and withdrawn for " + describe(pr) + ", not revising (not adding and not removing) CSR label");
                         } else if (!pr.labelNames().contains(CSR_LABEL)) {
                             log.info("CSR issue resolution is not 'Approved' for " + describe(pr) + ", adding the CSR label");
                             pr.addLabel(CSR_LABEL);
--- a/bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
+++ b/bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
@@ -92,33 +92,41 @@ class CSRBot implements Bot, WorkItem {
 
                     var resolution = csr.properties().get("resolution");
                     if (resolution == null || resolution.isNull()) {
-                        log.info("CSR issue resolution is null for " + describe(pr) + ", not removing CSR label");
                         if (!pr.labelNames().contains(CSR_LABEL)) {
+                            log.info("CSR issue resolution is null for " + describe(pr) + ", adding the CSR label");
                             pr.addLabel(CSR_LABEL);
+                        } else {
+                            log.info("CSR issue resolution is null for " + describe(pr) + ", not removing the CSR label");
                         }
                         continue;
                     }
                     var name = resolution.get("name");
                     if (name == null || name.isNull()) {
-                        log.info("CSR issue resolution name is null for " + describe(pr) + ", not removing CSR label");
                         if (!pr.labelNames().contains(CSR_LABEL)) {
+                            log.info("CSR issue resolution name is null for " + describe(pr) + ", adding the CSR label");
                             pr.addLabel(CSR_LABEL);
+                        } else {
+                            log.info("CSR issue resolution name is null for " + describe(pr) + ", not removing the CSR label");
                         }
                         continue;
                     }
 
                     if (csr.state() != Issue.State.CLOSED) {
-                        log.info("CSR issue state is not closed for " + describe(pr) + ", not removing CSR label");
                         if (!pr.labelNames().contains(CSR_LABEL)) {
+                            log.info("CSR issue state is not closed for " + describe(pr) + ", adding the CSR label");
                             pr.addLabel(CSR_LABEL);
+                        } else {
+                            log.info("CSR issue state is not closed for " + describe(pr) + ", not removing the CSR label");
                         }
                         continue;
                     }
 
                     if (!name.asString().equals("Approved")) {
-                        log.info("CSR issue resolution is not 'Approved' for " + describe(pr) + ", not removing CSR label");
                         if (!pr.labelNames().contains(CSR_LABEL)) {
+                            log.info("CSR issue resolution is not 'Approved' for " + describe(pr) + ", adding the CSR label");
                             pr.addLabel(CSR_LABEL);
+                        } else {
+                            log.info("CSR issue resolution is not 'Approved' for " + describe(pr) + ", not removing the CSR label");
                         }
                         continue;
                     }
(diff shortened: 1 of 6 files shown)


bots/csr/src/test/java/org/openjdk/skara/bots/csr/CSRBotTests.java
```diff
@@ -162,13 +162,19 @@ void keepLabelForNotApprovedCSR(TestInfo testInfo) throws IOException {
             localRepo.push(editHash, repo.url(), "edit", true);
             var pr = credentials.createPullRequest(repo, "master", "edit", issue.id() + ": This is an issue");
 
+            // Run bot
+            TestBotRunner.runPeriodicItems(bot);
+
+            // The bot should have kept the CSR label
```
erikj79, line 168: Is this a check for if the bot added the label automatically?

bots/csr/src/test/java/org/openjdk/skara/bots/csr/CSRBotTests.java
```diff
@@ -162,13 +162,19 @@ void keepLabelForNotApprovedCSR(TestInfo testInfo) throws IOException {
...
+            assertTrue(pr.labelNames().contains("csr"));
+
             // Add CSR label
             pr.addLabel("csr");
```
erikj79, line 172: Adding the label here has no effect.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CSRCommand.java
```diff
@@ -76,7 +77,28 @@ public void handle(PullRequestBot bot, PullRequest pr, CensusInstance censusInst
                 return;
             }
 
+            // Change the CSR link relationship from `csr for` to `relates to`
+            // so that the CSRBot won't add the `csr` label automatically.
```
erikj79, line 81: I'm not sure this is a good idea. The only reason to keep the /csr command around would be to have a way of manually correcting the state of a PR if the bots somehow fail, but I'm leaning towards just removing the command. If a CSR is already linked in JBS, but is later deemed to not be needed, I think the user should fix that in JBS rather than through a Skara command.

I think this needs some discussion and input from more people.

erikj79: I read up more on the history here. The /csr command can be invoked prior to creating a CSR issue in JBS, so it still makes sense to have the command. Given this, then invoking "/csr unneeded" can happen in two different scenarios:

1. No CSR has been filed yet, then just remove the csr label
2. A CSR is linked in JBS, in this case the CSRBot will likely automatically add the label again

I still don't think it makes sense to have this command go and change things in JBS. In case 2, a better cause of action would be to tell the user in an error reply that a CSR already exists and needs to be closed/removed first.

kevinrushforth:
The primary way that I use the `/csr` command is as a Reviewer to indicate that a PR needs to have a CSR when one might not yet exist. In some cases, the PR author might not realize that a CSR is needed. So the `/csr` command is definitely needed for this case.

I also agree with Erik that the CSR bot should not modify anything in JBS.

The automatic detection of an already-existing CSR in JBS seems like a "nice to have" feature for the case where a CSR already exists prior to creating the PR, so that the author (or a Reviewer) doesn't have to remember to say `/csr needed` explicitly, but we still need to be able to manually set it (and clear it in the case where further discussion indicates that a CSR really isn't needed).

erikj79:
If a CSR is created in JBS, but later deemed not necessary, what would be the correct steps to handle that in the current process? Can you close a CSR type issue with something other than "approved", that would have the meaning of "not needed" that we could look for and react to in the CSRBot?

kevinrushforth:
Yes there is a closed "Withdrawn" state, but I would probably prefer a Reviewer to issue the `/csr unneeded` in that case. Otherwise, A PR author with the role of Committer could withdraw a CSR that a Reviewer had requested and integrate it without satisfying that requirement.

erikj79:
> Yes there is a closed "Withdrawn" state, but I would probably prefer a Reviewer to issue the `/csr unneeded` in that case. Otherwise, A PR author with the role of Committer could withdraw a CSR that a Reviewer had requested and integrate it without satisfying that requirement.

Yes, that sounds good. What I'm mostly after is what conditions need to be met for the `/csr unneeded` command to succeed. Given this, there are two situations where it should succeed:

1. No CSR issue exists.
2. A CSR issue exists but has been withdrawn.

If neither of those two are true, then issuing `/csr unneeded` should result in an error explaining that a CSR has already been filed and will need to be withdrawn first.

kevinrushforth:
Yes, this seems good.

bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
```diff
@@ -106,27 +93,40 @@ private String describe(PullRequest pr) {
                     var resolution = csr.properties().get("resolution");
                     if (resolution == null || resolution.isNull()) {
                         log.info("CSR issue resolution is null for " + describe(pr) + ", not removing CSR label");
```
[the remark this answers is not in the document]
lgxbslgx: I adjusted the log message to reflect these two situation.

bots/csr/src/test/java/org/openjdk/skara/bots/csr/CSRBotTests.java
lgxbslgx: yes, I revised the comment for readability.

bots/csr/src/test/java/org/openjdk/skara/bots/csr/CSRBotTests.java
lgxbslgx: yes, because the bot has already added csr label automatically. I removed this step.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CSRCommand.java
lgxbslgx: I revised the logic of the command `/csr unneeded` according to your comments below.

lgxbslgx:
@kevinrushforth @erikj79 Thanks for the reviews and comments. I updated the code just now.

lgxbslgx:
I refactor the code to reduce the nested statements.

bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
```diff
@@ -105,28 +92,55 @@ private String describe(PullRequest pr) {
...
 
                     if (!name.asString().equals("Approved")) {
-                        log.info("CSR issue resolution is not 'Approved' for " + describe(pr) + ", not removing CSR label");
+                        if (name.asString().equals("Withdrawn")) {
+                            // This condition is necessary to prevent the bot from adding the CSR label again.
+                            // And the bot can't remove the CSR label automatically here.
+                            // Because the PR author with the role of Committer may withdraw a CSR that
+                            // a Reviewer had requested and integrate it without satisfying that requirement.
+                            log.info("CSR closed and withdrawn for " + describe(pr) + ", not removing CSR label");
+                        } else if (!pr.labelNames().contains(CSR_LABEL)) {
```
erikj79, line 131: I'm not sure about this case. For this to realistically happen, a CSR would have been filed and withdrawn before the PR was created. I don't think we should automatically add the label in this situation. @kevinrushforth what do you think?

If the label is already there, then we should certainly leave it there however.

kevinrushforth: I agree that we need not (should not) add a label in this case. I also agree that if there is a label we must leave it.

This will handle the case where a PR author thinks a CSR might be needed, creates a CSR, withdraws it, creates the PR, and then later a Reviewer indicates that a CSR is needed. In that case, the bot should add the label and not auto-remove it.

lgxbslgx: @erikj79 @kevinrushforth Thanks for the comments. It seems that my current code meets your comments. Did I misunderstand anything?

Please see the code below. When the CSR has been withdrawn, the bot only prints the information and doesn't add the csr label automatically. 

```
if (!name.asString().equals("Approved")) {
    if (name.asString().equals("Withdrawn")) {
        log.info("CSR closed and withdrawn for " + describe(pr) + ", not removing CSR label");  // <----here
    } else if (!pr.labelNames().contains(CSR_LABEL)) {
        log.info("CSR issue resolution is not 'Approved' for " + describe(pr) + ", adding the CSR label");
        pr.addLabel(CSR_LABEL);
    } else {
        log.info("CSR issue resolution is not 'Approved' for " + describe(pr) + ", not removing the CSR label");
    }
    continue;
}
```


Maybe I should change the output message from `"CSR closed and withdrawn for " + describe(pr) + ", not removing CSR label"` to `"CSR closed and withdrawn for " + describe(pr) + ", not revising (not adding or not removing) CSR label"`.

erikj79: You are right, I misread your code.

erikj79 (approved):

lgxbslgx:
@erikj79 Thanks for the review. Could I get your help to sponsor this patch?

/integrate

PR #1246, opened 2021-11-27
1254: Add a checkbox for CSR requirement

lgxbslgx:
Hi all,

This patch adds a checkbox for CSR requirement in the `Progress` and adds the corresponding test cases.

Best Regards,
-- Guoxiong

---------
### Progress

### Issue
 * [SKARA-1254](https://bugs.openjdk.java.net/browse/SKARA-1254): Add a checkbox for CSR requirement

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1246/head:pull/1246` \
`$ git checkout pull/1246`

`$ git checkout pull/1246` \
`$ git pull https://git.openjdk.java.net/skara pull/1246/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1246`

`$ git pr show -t 1246`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1246.diff">https://git.openjdk.java.net/skara/pull/1246.diff</a>

</details>

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -167,6 +167,17 @@ class CheckRun {
         return ret;
     }
 
+    // Additional bot-specific progresses that are not handled by JCheck
+    private Map<String, Boolean> botSpecificProgresses() {
+        var ret = new HashMap<String, Boolean>();
+        if (pr.labelNames().contains("csr")) {
+            ret.put("Change requires a CSR request to be approved", false);
+        } else {
+            ret.put("Change doesn't require a CSR request or the CSR request has been approved", true);
+        }
+        return ret;
+    }
+
     private void setExpiration(Duration expiresIn) {
         // Use the shortest expiration
         if (this.expiresIn == null || this.expiresIn.compareTo(expiresIn) > 0) {
@@ -175,8 +186,7 @@ class CheckRun {
     }
 
     private Map<String, String> blockingIntegrationLabels() {
-        return Map.of("rejected", "The change is currently blocked from integration by a rejection.",
-                      "csr", "The change requires a CSR request to be approved.");
+        return Map.of("rejected", "The change is currently blocked from integration by a rejection.");
     }
 
     private List<String> botSpecificIntegrationBlockers() {
@@ -343,8 +353,9 @@ class CheckRun {
         }
     }
 
-    private String getChecksList(PullRequestCheckIssueVisitor visitor, boolean isCleanBackport) {
+    private String getChecksList(PullRequestCheckIssueVisitor visitor, boolean isCleanBackport, Map<String, Boolean> additionalProgresses) {
         var checks = isCleanBackport ? visitor.getReadyForReviewChecks() : visitor.getChecks();
+        checks.putAll(additionalProgresses);
         return checks.entrySet().stream()
                 .map(entry -> "- [" + (entry.getValue() ? "x" : " ") + "] " + entry.getKey())
                 .collect(Collectors.joining("\n"));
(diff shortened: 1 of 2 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -167,6 +167,17 @@ private IssueProject issueProject() {
         return ret;
     }
 
+    // Additional bot-specific progresses that are not handled by JCheck
+    private Map<String, Boolean> botSpecificProgresses() {
+        var ret = new HashMap<String, Boolean>();
+        if (pr.labelNames().contains("csr")) {
+            ret.put("Change requires a CSR request to be approved", false);
+        } else {
+            ret.put("Change doesn't require a CSR request or the CSR request has been approved", true);
```
erikj79, line 176: I don't think we should show anything here if a CSR was never requested. The difference between "CSR was never requested" and "CSR is approved" isn't really tracked currently, so that would be tricky. I'm guessing this is a big reason why the original implementation didn't add checkboxes, but only a blocker message when active. For this to work we will need to add some state that tracks if a CSR is currently required. We basically have 3 states:

1. No CSR needed
2. CSR needed and not yet approved
3. CSR needed and approved

Every PR starts in state 1. To get to state 2, you can either issue the /csr command, or with SKARA-1071 file a CSR and let Skara find it. To get to state 3, the CSR must be approved. At any time, the command "/csr unneeded" can take you back to state 1.

This could be implemented with another label, or it could rely on special comments with markers. In this case, I think we want to make it visible to users, so labels probably make more sense. For backwards compatibility, to keep the current meaning of the "csr" label, the new label would need to be something like "csr-approved" and be added at the same time the csr label is removed when moving to state 3.

lgxbslgx: @erikj79  The three states you mentioned above is clear. From your comment, I conclude the following things:

1. No CSR needed: the PR doesn't have `csr` and `csr-approved` labels.
2. CSR needed and not yet approved: the PR has `csr` label and doesn't have `csr-approved` label.
3. CSR needed and approved: the PR has `csr-approved` label and doesn't have `csr` label.

And the `csr` label and `csr-approved` label can't appear at the same time.

But currently, the CSRBot and CSRCommand have the following logic:

1. No CSR needed: the PR doesn't have `csr` label and the csr issue doesn't exist or has been withdrawn.
2. CSR needed and not yet approved: the PR has `csr` label.
3. CSR needed and approved: the PR doesn't have `csr` label and the csr issue has been Approved.

If we obey the new logic and create the new label `csr-approved`, some checks can be achieved easily. But the logic change may cause the unexpected regression. As you can see the discussion in the [PR-1245](https://github.com/openjdk/skara/pull/1245) of the SKARA-1071, both the authors of the patch and the reviewers need to be very careful to avoid the misunderstanding and avoid making a mistake.

So I would like to obey the origin logic. In this patch, I can check the csr link and issue to identify the state 1 and state 3 which is like the CSRBot and CSRCommand. So we don't need to add another label such as `csr-approved` to solve this problem. What is your opinion about this?

---
And a related thing:
There are many places which need to get the csr issue from the original issue. We should provide a common method to solve this common problem. So I filed [SKARA-1256](https://bugs.openjdk.java.net/browse/SKARA-1256) to solve it.

erikj79: I needed to think about this for a bit, but I think you are right. 

I was instinctively trying to keep the PR bot from having to query JBS for CSR status, because we currently have a label which made checking the CSR state very cheap and obvious. But on the other hand, we are fetching the issue from JBS anyway during every CheckRun, so adding this extra step isn't that expensive.

lgxbslgx: @erikj79 I re-think about this. Even though we decide to use the new logic and add the new label `csr-approved` at the end,  we shouldn't implement the feature at this patch. Because the goal of this patch is only to add a checkbox for CSR requirement.

So I advice to file a new issue to continue the discussion. The new issue title can be `Investigating to add a label 'csr-approved' to mark the CSR is approved`.

Back to this patch, I think it is good to use the old logic to implement the enhancement so that the users (OpenJDK developers) can take advantage of this patch as soon as possible.

erikj79: I don't think we need a new label. Your suggestion to tell state 1 and 3 apart by checking if a CSR issue exists seems good enough to me.

magicus (approved):
This looks good to me, now.

erikj79 (approved):
Looks good!

lgxbslgx:
@erikj79 @magicus Thanks for the reviews. Could I get your help to sponsor this patch?

/integrate

PR #1247, opened 2021-11-30
1236: Add jcheck option to check for large binary files

lgxbslgx:
Hi all,

This patch adds the large binary file check to the jcheck.

The corresponding configuration is like the following code. Each key/value in the `[checks "binary"]` is a mapper. The `key` means the file pattern and the `value` means the limited file size. For example, `.*\\.bin=200B` means the file which ends with `.bin` should not exceed 200 Bytes.

```
[checks]
error = binary

[checks "binary"]
.*\.bin=200B
.*\.o=1k
```

The limited file size can use one of these several units:  b(Byte), kb(KiloByte), mb(MegaByte), gb(GigaByte). The units is case insensitive, which means the `Kb`, `kB` and `KB` are equals to `kb(KiloByte)`. If the unit is not provided, it defauts to `b(Byte)`. And these is no a unit called `bit`.

The corresponding test cases are added.

Best Regards,
-- Guoxiong

---------
### Progress

### Issue
 * [SKARA-1236](https://bugs.openjdk.java.net/browse/SKARA-1236): Add jcheck option to check for large binary files

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1247/head:pull/1247` \
`$ git checkout pull/1247`

`$ git checkout pull/1247` \
`$ git pull https://git.openjdk.java.net/skara pull/1247/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1247`

`$ git pr show -t 1247`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1247.diff">https://git.openjdk.java.net/skara/pull/1247.diff</a>

</details>

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2021, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -262,8 +262,11 @@ class PullRequestCheckIssueVisitor implements IssueVisitor {
     }
 
     @Override
-    public void visit(BinaryIssue issue) {
-        log.fine("ignored: binary file");
+    public void visit(BinaryFileTooLargeIssue issue) {
+        addFailureMessage(issue.check(), "The size of the binary file `" + issue.path() + "` is " + issue.fileSize()
+                + (issue.fileSize() > 1 ? " Bytes" : " Byte") + ", which is larger than the limited file size: "
+                + issue.limitedFileSize() + (issue.limitedFileSize() > 1 ? " Bytes." : " Byte."));
+        readyForReview = false;
     }
 
     @Override
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
@@ -270,7 +270,7 @@ class PullRequestCheckIssueVisitor implements IssueVisitor {
     @Override
     public void visit(BinaryFileTooLargeIssue issue) {
         addFailureMessage(issue.check(), "The size of the binary file `" + issue.path() + "` is " + issue.fileSize()
-                + " Bytes, which is larger than the limited file size: "
+                + " Bytes, which is larger than the binary file size limit: "
                 + issue.maxSize() + (issue.maxSize() > 1 ? " Bytes." : " Byte."));
         readyForReview = false;
     }
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
@@ -262,10 +262,16 @@ class PullRequestCheckIssueVisitor implements IssueVisitor {
     }
 
     @Override
+    public void visit(BinaryIssue issue) {
+        addFailureMessage(issue.check(), "The binary file " + issue.path().toString() + " is not allowed in this repository.");
+        readyForReview = false;
+    }
+
+    @Override
     public void visit(BinaryFileTooLargeIssue issue) {
         addFailureMessage(issue.check(), "The size of the binary file `" + issue.path() + "` is " + issue.fileSize()
-                + (issue.fileSize() > 1 ? " Bytes" : " Byte") + ", which is larger than the limited file size: "
-                + issue.limitedFileSize() + (issue.limitedFileSize() > 1 ? " Bytes." : " Byte."));
+                + " Bytes, which is larger than the limited file size: "
+                + issue.maxSize() + (issue.maxSize() > 1 ? " Bytes." : " Byte."));
         readyForReview = false;
     }
 
--- a/cli/src/main/java/org/openjdk/skara/cli/JCheckCLIVisitor.java
+++ b/cli/src/main/java/org/openjdk/skara/cli/JCheckCLIVisitor.java
@@ -300,7 +300,7 @@ class JCheckCLIVisitor implements IssueVisitor {
     public void visit(BinaryFileTooLargeIssue i) {
         if (!ignore.contains(i.check().name())) {
             println(i, "The size of the binary file `" + i.path() + "` is " + i.fileSize()
-                    + " Bytes, which is larger than the limited file size: "
+                    + " Bytes, which is larger than the binary file size limit: "
                     + i.maxSize() + (i.maxSize() > 1 ? " Bytes." : " Byte."));
             hasDisplayedErrors = true;
         }
--- a/cli/src/main/java/org/openjdk/skara/cli/JCheckCLIVisitor.java
+++ b/cli/src/main/java/org/openjdk/skara/cli/JCheckCLIVisitor.java
@@ -290,11 +290,18 @@ class JCheckCLIVisitor implements IssueVisitor {
         }
     }
 
+    public void visit(BinaryIssue i) {
+        if (!ignore.contains(i.check().name())) {
+            println(i, "The binary file " + i.path().toString() + " is not allowed in this repository.");
+            hasDisplayedErrors = true;
+        }
+    }
+
     public void visit(BinaryFileTooLargeIssue i) {
         if (!ignore.contains(i.check().name())) {
             println(i, "The size of the binary file `" + i.path() + "` is " + i.fileSize()
                     + (i.fileSize() > 1 ? " Bytes" : " Byte") + ", which is larger than the limited file size: "
-                    + i.limitedFileSize() + (i.limitedFileSize() > 1 ? " Bytes." : " Byte."));
+                    + i.maxSize() + (i.maxSize() > 1 ? " Bytes." : " Byte."));
             hasDisplayedErrors = true;
         }
     }
--- a/cli/src/main/java/org/openjdk/skara/cli/JCheckCLIVisitor.java
+++ b/cli/src/main/java/org/openjdk/skara/cli/JCheckCLIVisitor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2021, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -290,9 +290,11 @@ class JCheckCLIVisitor implements IssueVisitor {
         }
     }
 
-    public void visit(BinaryIssue i) {
+    public void visit(BinaryFileTooLargeIssue i) {
         if (!ignore.contains(i.check().name())) {
-            println(i, "adds binary file: " + i.path().toString());
+            println(i, "The size of the binary file `" + i.path() + "` is " + i.fileSize()
+                    + (i.fileSize() > 1 ? " Bytes" : " Byte") + ", which is larger than the limited file size: "
+                    + i.limitedFileSize() + (i.limitedFileSize() > 1 ? " Bytes." : " Byte."));
             hasDisplayedErrors = true;
         }
     }
--- a/cli/src/main/java/org/openjdk/skara/cli/JCheckCLIVisitor.java
+++ b/cli/src/main/java/org/openjdk/skara/cli/JCheckCLIVisitor.java
@@ -300,7 +300,7 @@ class JCheckCLIVisitor implements IssueVisitor {
     public void visit(BinaryFileTooLargeIssue i) {
         if (!ignore.contains(i.check().name())) {
             println(i, "The size of the binary file `" + i.path() + "` is " + i.fileSize()
-                    + (i.fileSize() > 1 ? " Bytes" : " Byte") + ", which is larger than the limited file size: "
+                    + " Bytes, which is larger than the limited file size: "
                     + i.maxSize() + (i.maxSize() > 1 ? " Bytes." : " Byte."));
             hasDisplayedErrors = true;
         }
--- a/jcheck/src/main/java/org/openjdk/skara/jcheck/BinaryCheck.java
+++ b/jcheck/src/main/java/org/openjdk/skara/jcheck/BinaryCheck.java
@@ -45,13 +45,13 @@ public class BinaryCheck extends CommitCheck {
         for (var diff : commit.parentDiffs()) {
             for (var patch : diff.patches()) {
                 if (patch.isTextual()) {
+                    // Excluded the textual patch.
                     continue;
                 }
-                // Here, the file is surely binary.
-                if (!patch.status().isAdded() && !patch.status().isCopied()) {
+                if (patch.status().isDeleted()) {
+                    // Excluded the deleted file.
                     continue;
                 }
-                // Here, the binary file is newly added or copied.
                 if (maxSize == 0) {
                     // If the maxSize is not set or is set to 0, any binary file can't be added.
                     issues.add(new BinaryIssue(patch.target().path().get(), metadata));
--- a/jcheck/src/main/java/org/openjdk/skara/jcheck/BinaryCheck.java
+++ b/jcheck/src/main/java/org/openjdk/skara/jcheck/BinaryCheck.java
@@ -31,9 +31,7 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.util.Iterator;
 import java.util.ArrayList;
-import java.util.Map;
 import java.util.logging.Logger;
-import java.util.regex.Pattern;
 
 public class BinaryCheck extends CommitCheck {
     private final Logger log = Logger.getLogger("org.openjdk.skara.jcheck.binary");
--- a/jcheck/src/main/java/org/openjdk/skara/jcheck/BinaryCheck.java
+++ b/jcheck/src/main/java/org/openjdk/skara/jcheck/BinaryCheck.java
@@ -52,20 +52,25 @@ public class BinaryCheck extends CommitCheck {
                     // Excluded the deleted file.
                     continue;
                 }
-                if (maxSize == 0) {
+                if (maxSize <= 0) {
                     // If the maxSize is not set or is set to 0, any binary file can't be added.
                     issues.add(new BinaryIssue(patch.target().path().get(), metadata));
+                }
+                if (patch.status().isRenamed() || patch.status().isCopied()) {
+                    // Excluded the renamed and copied files now. Note that the moved file is identified as renamed.
+                    // It should be seperated from the deleted file, because the deleted file don't need
+                    // to be checked, but the renamed file should be checked for the `BinaryIssue`.
+                    continue;
                 }
                 if (maxSize > 0) {
                     long fileSize = 0;
                     var binaryPatch = (BinaryPatch) patch;
-                    var path =  binaryPatch.target().path().get();
-                    try {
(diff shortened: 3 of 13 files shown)


kevinrushforth:
Oops. I added the `/csr` label by mistake (although I think it's a no-op for this repo).

lgxbslgx:
> I would rather see two separate parameters, a size and a list of files (with regex support, like we have for the whitespace check), so that we don't have to repeat the size. We don't need the flexibility to set the limit per-extension.

This binary file check is not much like the whitespace check. The different file types have different basic sizes. The image type may be 500B-4kB and the audio type may be 2kB-32kB. So we can set the config as the following:

```
[checks]
error = binary

[checks "binary"]
; the image type
.*\.jpg|.*\.png|.*\.gif=5k

; the audio type
.*\.wav|.*\.mp3=40k
```

If we have the only limited size as 40k, it is too large for the image type. If we have the only limited size as 5k, it is too small for the audio type.

The demo above only consider two file types.  The other types, such as video type and so on, have the same issue we need to solve. So I think the configuration in this patch is more flexible than the whitespache check and can meet the requirements of our repositories.

erikj79:
First, thank you for taking on this issue!

I agree with Kevin, I don't think we have any need for configuring different file types having different size requirements. I think that it's neat to support it, but we just wouldn't use it for any OpenJDK repo that I know of. If we remove that requirement, then there is no need to list file patterns at all, since the check only applies to binary files anyway, and Git tells us which files are binary already.

It looks like you have taken the existing "binary" check and changed it to be a size check. The existing check wasn't used in the JDK repos, but I suspect someone is using it to forbid any binary files. That functionality seems to be lost with this patch. We can't change existing behavior as that would cause incompatibility with existing .jcheck/conf files.

I propose the following extension to the existing "binary" check:

```
[checks]
error = binary

[checks "binary"]
max-size=1M
```

The max-size parameter is optional. If not set, then the binary check works like today, triggering on any binary file, with a reasonable error message that isn't mentioning the size limit, or being too large.

lgxbslgx:
> I agree with Kevin, I don't think we have any need for configuring different file types having different size requirements.

If no one agrees with me or gives another idea these days, I will follow this idea.

> It looks like you have taken the existing "binary" check and changed it to be a size check. 

Because when I dove into the class `PullRequestCheckIssueVisitor`, I fould that the `BinaryIssue` is never used and the check result has no effect in the PR. So I think `BinaryIssue` is not necessary and rename it to `BinaryFileTooLargeIssue`.

```
// class PullRequestCheckIssueVisitor
    @Override
    public void visit(BinaryIssue issue) {
        log.fine("ignored: binary file");
        // <---- miss to add the failed message to the `failedChecks`.
    }
```

But now I find the `JCheckCLIVisitor` use the `BinaryIssue` which means the `BinaryIssue` is effective. So now I think the `PullRequestCheckIssueVisitor` just has a small bug that it miss to add the failed message to the `failedChecks`. And actually, the result of the `binary` check in the PR has never activated in the past.

```
// class JCheckCLIVisitor
    public void visit(BinaryIssue i) {
        if (!ignore.contains(i.check().name())) {
            println(i, "adds binary file: " + i.path().toString());
            hasDisplayedErrors = true;
        }
    }
```

I will solve this small bug of the `PullRequestCheckIssueVisitor` in this patch and will keep the `BinaryIssue` instead of removing `BinaryIssue`.

magicus:
I agree with the other reviewers that we only need a single binary limit. 

But be aware that actually deploying a change in the main JDK repo which enables this in jcheck is a separate discussion. I don't think we've ever had a hard limit on binary file sizes, only just that you should "try to avoid too large files". I foresee a highly opinionated debate on what size limit to use, if we should even activate it at all. Having a very conservative upper value is more likely to be accepted (let's say the currently largest file times two), and if that seems to work fine, we might try lowering the value.

magicus:
In a separate, but related note, I'd really like to let jcheck have some kind of "warnings", information that is not strictly blockers, but signals that you should make sure you really intend to do what you're doing. If we had that, then we could have a (generous) upper bound for binary sizes where jcheck makes it an error, and a much lower bound for which jcheck warns you that you perhaps should reconsider.

jcheck/src/main/java/org/openjdk/skara/jcheck/BinaryCheck.java
```diff
@@ -23,33 +23,85 @@
...
+                if (binaryPatch.status().isAdded()) {
+                    // This is a new-added file, so the BinaryHunk#inflatedSize of the first hunk should be the file size.
+                    fileSize = binaryPatch.hunks().get(0).inflatedSize();
+                    needCheck = true;
+                } else if (binaryPatch.status().isRenamed() || binaryPatch.status().isCopied()
+                        || binaryPatch.status().isModified() || binaryPatch.status().isUnmerged()) {
+                    // Use the file size in the file system.
+                    try {
+                        fileSize = path != null ? Files.size(path) : 0;
+                        needCheck = true;
```
magicus, line 69: Does that mean that jcheck will block renaming an existing binary file if it is larger than the currently accepted maximum? I don't think that is the right behavior. Only adding a new file should trigger this.

erikj79:
> In a separate, but related note, I'd really like to let jcheck have some kind of "warnings", information that is not strictly blockers, but signals that you should make sure you really intend to do what you're doing. If we had that, then we could have a (generous) upper bound for binary sizes where jcheck makes it an error, and a much lower bound for which jcheck warns you that you perhaps should reconsider.

Having two levels, warn and error, is a good idea. I'm not sure how well the current jcheck model fits with that, but if it does, I'm all for that. Then we would need two parameters for configuration. One to trigger warning and one to trigger error.

erikj79:
> > I agree with Kevin, I don't think we have any need for configuring different file types having different size requirements.
> 
> If no one agrees with me or gives another idea these days, I will follow your idea.
> 
> > It looks like you have taken the existing "binary" check and changed it to be a size check.
> 
> Because when I dove into the class `PullRequestCheckIssueVisitor`, I fould that the `BinaryIssue` is never used and the check result has no effect in the PR. So I think `BinaryIssue` is not necessary and rename it to `BinaryFileTooLargeIssue`.
> 
> ```
> // class PullRequestCheckIssueVisitor
>     @Override
>     public void visit(BinaryIssue issue) {
>         log.fine("ignored: binary file");
>         // <---- miss to add the failed message to the `failedChecks`.
>     }
> ```
> 
> But now I find the `JCheckCLIVisitor` use the `BinaryIssue` which means the `BinaryIssue` is effective. So now I think the `PullRequestCheckIssueVisitor` just has a small bug that it miss to add the failed message to the `failedChecks`. And actually, the result of the `binary` check in the PR has never activated in the past.
> 
> ```
> // class JCheckCLIVisitor
>     public void visit(BinaryIssue i) {
>         if (!ignore.contains(i.check().name())) {
>             println(i, "adds binary file: " + i.path().toString());
>             hasDisplayedErrors = true;
>         }
>     }
> ```
> 
> I will solve this small bug of the `PullRequestCheckIssueVisitor` in this patch and will keep the `BinaryIssue` instead of removing `BinaryIssue`.

I found the original PR where this was added https://github.com/openjdk/skara/pull/159. It looks like they never got around to implementing handling in the PR visitor. I think it's good if you fix that.

lgxbslgx:
> I found the original PR where this was added #159. It looks like they never got around to implementing handling in the PR visitor. I think it's good if you fix that.

I find that the field `JCheck#commitChecks` has no the item `BinaryCheck`. So it seems that the `BinaryCheck` never run in both PR and git-jcheck command. I wiil fix it, too.

jcheck/src/main/java/org/openjdk/skara/jcheck/BinaryCheck.java
lgxbslgx: I intended to check it. But if you think it don't need to check, I could remove it.

lgxbslgx:
About the errors and warnings, I re-read the related code.

Currently, the warning config is not the actual warnings in the PR. The PR bot doesn't distinguish them in the method `hiddenMessages` and `messages` of the class PullRequestCheckIssueVisitor. Both the errors and the warnings will be shown at the `### Errors` part of the PR first comment body and all of them will block the integration. Is it the intent of the PR bot?

> we could have a (generous) upper bound for binary sizes where jcheck makes it an error, and a much lower bound for which jcheck warns you that you perhaps should reconsider.

The goal is good. But we need to handle it carefully. Because the check config has the errors and warnings config, which I mentioned above.

If we add the check level, `error` and `warning`, in a concrete check, the understanding, usage and implementation seem more difficult to the user and the skara developer.  Note what I mention above firstly:  the current code about current errors and warnings config is not clear and is not equal to the git-jcheck before.

kevinrushforth:
I think having a single value for errors is sufficient. I can see how allowing for a lower "warning" limit might be useful, but I don't recommend holding up this PR to implement it. It could be added in the future.

erikj79:
> I think having a single value for errors is sufficient. I can see how allowing for a lower "warning" limit might be useful, but I don't recommend holding up this PR to implement it. It could be added in the future.

I agree with Kevin, since the current implementation isn't readily letting us do this, then skip it for now.

lgxbslgx:
I updated the code just now. The change is listed below.

1. Use the simple config `max-size`. If the `max-size` is not set or is set to zero, the binary file is not allowed, or the BinaryIssue is raised. If the `max-size` is set and is not zero, the binary file can't be larger than `max-size`, or the BinaryFileTooLargeIssue is raised.
2. Restore the BinaryIssue, which was removed mistakenly.
3. Only check the newly added or copied files. The modified or renamed  binary files won't be checked.
4. Fix the bug that the binary check is not enabled.

kevinrushforth:
> Only check the newly added or copied files. The modified or renamed binary files won't be checked.

A modified file is just as much of a problem, and much more likely to occur, as adding a new file. I strongly recommend checking all files, including modified or renamed files.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
```diff
@@ -263,7 +263,16 @@ public void visit(SymlinkIssue issue) {
...
     public void visit(BinaryIssue issue) {
-        log.fine("ignored: binary file");
+        addFailureMessage(issue.check(), "The binary file " + issue.path().toString() + " is not allowed in this repository.");
+        readyForReview = false;
+    }
+
+    @Override
+    public void visit(BinaryFileTooLargeIssue issue) {
+        addFailureMessage(issue.check(), "The size of the binary file `" + issue.path() + "` is " + issue.fileSize()
+                + " Bytes, which is larger than the limited file size: "
```
erikj79, line 273:
```suggestion
                + " Bytes, which is larger than the binary file size limit: "
```

cli/src/main/java/org/openjdk/skara/cli/JCheckCLIVisitor.java
```diff
@@ -292,7 +292,16 @@ public void visit(CommitterEmailIssue i) {
...
-            println(i, "adds binary file: " + i.path().toString());
+            println(i, "The binary file " + i.path().toString() + " is not allowed in this repository.");
+            hasDisplayedErrors = true;
+        }
+    }
+
+    public void visit(BinaryFileTooLargeIssue i) {
+        if (!ignore.contains(i.check().name())) {
+            println(i, "The size of the binary file `" + i.path() + "` is " + i.fileSize()
+                    + " Bytes, which is larger than the limited file size: "
```
erikj79, line 303:
```suggestion
                    + " Bytes, which is larger than the binary file size limit: "
```

erikj79:
> > Only check the newly added or copied files. The modified or renamed binary files won't be checked.
> 
> A modified file is just as much of a problem, and much more likely to occur, as adding a new file. I strongly recommend checking all files, including modified or renamed files.

To expand on this a bit. My main motivation for this feature is to keep the repo clean of big binaries, as that's a bad thing in general. We also want to protect ourselves from hitting similar checks in other places, typically by downstream consumers of the JDK repo. We know of one such case today (but I wouldn't be surprised if there are more cases), and unfortunately, that check isn't smart enough to tell different types of changes apart, and we aren't able to affect the implementation of that check. So for that reason, I would also ask you to revert what Magnus asked, and let this apply to any binary file, regardless of how it was changed. We may revisit this later and relax this requirement if possible in the future.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
lgxbslgx: Fixed.

cli/src/main/java/org/openjdk/skara/cli/JCheckCLIVisitor.java
lgxbslgx: Fixed.

lgxbslgx:
>  let this apply to any binary file,

Fixed.

I have a small problem. Can the following code get the file size actually? The `path` seems a local relative path.

```
try {
    fileSize = Files.size(path);
} catch (IOException e) {
    log.warning("The file '" + path + "' doesn't exist.");
}
```

erikj79:
> > let this apply to any binary file,
> 
> Fixed.
> 
> I have a small problem. Can the following code get the file size actually? The `path` seems a local relative path.
> 
> ```
> try {
>     fileSize = Files.size(path);
> } catch (IOException e) {
>     log.warning("The file '" + path + "' doesn't exist.");
> }
> ```

The path you get from a patch should be local to the repository. Assuming the file is in the workspace, it should be resolvable against the repository root, but I don't think a jcheck check can make such an assumption. AFAIK it's expected to just interact with the commit data. It's never updating the workspace to each inspected commit to perform the checks. I think you need to figure out the size by inspecting the binary patch. I'm not sure exactly how, but it looks like all the binary data is encoded in the patch hunks.

magicus:
I just want to confirm that Erik is correct. You need to check even updated and renamed files. 

I also want to confirm that I don't think getting a warning level should hold up this PR. As I said, we don't currently support warnings, and the entire concept needs to be developed, for all stages, up to and including how to present them to the user in the GitHub UI, before we can start to actually emit any warnings. I still think this would be a good thing to have, though.

lgxbslgx:
>  I think you need to figure out the size by inspecting the binary patch. I'm not sure exactly how, but it looks like all the binary data is encoded in the patch hunks.

I dive into the diff information to find the size. Only the added and deleted files have the entire size. The modified files have the delta size (modified size) which we can't use. And the renamed and copied files have no size, they only have the source path and target patch.

May need help from the experts who are famaliar with Git/HG.

erikj79:
> > I think you need to figure out the size by inspecting the binary patch. I'm not sure exactly how, but it looks like all the binary data is encoded in the patch hunks.
> 
> I dive into the diff information to find the size. Only the added and deleted files have the entire size. The modified files have the delta size (modified size) which we can't use. And the renamed and copied files have no size, they only have the source path and target patch.
> 
> May need help from the experts who are famaliar with Git/HG.

I think size delta in the case of modified is fine. In my experience, it's very rare for binary patches to actually be partial in real situations. When replacing an image or archive, the whole file usually changes enough. If the binary patch is empty, due to file moved/rename, then we can't really do anything about that unfortunately. Have you looked at actual patch output from Git to see what it looks like for renamed binary files? Git doesn't track file renames explicitly, so I'm not sure what would actually happen here.

lgxbslgx:
> I think size delta in the case of modified is fine. In my experience, it's very rare for binary patches to actually be partial in real situations. When replacing an image or archive, the whole file usually changes enough. If the binary patch is empty, due to file moved/rename, then we can't really do anything about that unfortunately. Have you looked at actual patch output from Git to see what it looks like for renamed binary files? Git doesn't track file renames explicitly, so I'm not sure what would actually happen here.

Moved file is identifed as `renamed`. The `renamed` diff information is shown below.

```
:100644 100644 c9eef64 c9eef64 R100     t/1.png t/2.png

diff --git a/t/1.png b/t/2.png
similarity index 100%
rename from t/1.png
rename to t/2.png
```

It only has the source file and target file and have no information about size.

erikj79:
> > I think size delta in the case of modified is fine. In my experience, it's very rare for binary patches to actually be partial in real situations. When replacing an image or archive, the whole file usually changes enough. If the binary patch is empty, due to file moved/rename, then we can't really do anything about that unfortunately. Have you looked at actual patch output from Git to see what it looks like for renamed binary files? Git doesn't track file renames explicitly, so I'm not sure what would actually happen here.
> 
> Moved file is identifed as `renamed`. The `renamed` diff information is shown below.
> 
> ```
> :100644 100644 c9eef64 c9eef64 R100     t/1.png t/2.png
> 
> diff --git a/t/1.png b/t/2.png
> similarity index 100%
> rename from t/1.png
> rename to t/2.png
> ```
> 
> It only has the source file and target file and have no information about size.

Then we can't check the size of files in such changes. We are limited to what Git will tell us here.

lgxbslgx:
> And the renamed and copied files have no size, they only have the source path and target patch.

Please note the copied file has no size, too:

```
:100644 100644 c9eef64 c9eef64 C100     1.png   t/1.png

diff --git a/1.png b/t/1.png
similarity index 100%
copy from 1.png
copy to t/1.png
```

kevinrushforth:
How is this diff generated? The commit itself does have the complete information. Here is a test I just did in the jfx repo:

```
$ git diff --no-renames --binary HEAD
diff --git a/tests/manual/web/dnd/blue_renamed.png b/tests/manual/web/dnd/blue_renamed.png
new file mode 100644
index 0000000000000000000000000000000000000000..d8f1d7de2baf7851a8c63506709ea0b83806c172
GIT binary patch
literal 207
zcmeAS@N?(olHy`uVBq!ia0vp^4j|0I1SD0tpLGJMX`U{QAsLNtZyEA7I54mrd~CX>
zH9xbkMOI6SW#R%xl?RMmDgilmHv<c9-{0`=eA$ydm*<8V`%4~sy3;D=f2Qfu`(HiJ
zS-0N%FrCqaIfLy2&jLvY<AyYbvkbFH5zt<HNw_Y<G<Dvk%){mT>luqo1)RTg`Mv?V
OkipZ{&t;ucLK6Uiphv*~

literal 0
HcmV?d00001

diff --git a/tests/manual/web/dnd/red_copied.png b/tests/manual/web/dnd/red_copied.png
new file mode 100644
index 0000000000000000000000000000000000000000..66d776a846d3671512c5afba2623fb3a0dac847e
GIT binary patch
literal 284
zcmeAS@N?(olHy`uVBq!ia0vp^4j|0I1SD0tpLGJM$DS^ZAsLNtFDh~!N?>Sv_&55Z
z(KclXF&Vv%@QqP1qHJdUiWyh3Iht}g+EfoF{kYt;ZmsUw2b?7jQ$N_M|A@2OynoyI
z$7SCvdS30mEPu_v%c6JMO!1iob<yo-yXSqGy`i%Iba5Q(y{v$5d4H!D-b##}RpZ3}
z^4L=LlBEo@m@>E)Xf`Y(i}5VVexdoF^S?Tyk7Q<7{dCU0qvW^u=({Q52RG;)e&*}{
itMTjP(<Q>XoB98W>z)d%Zc+kzm%-E3&t;ucLK6Tl6m?4g

literal 0
HcmV?d00001
```

lgxbslgx:
> How is this diff generated? The commit itself does have the complete information. Here is a test I just did in the jfx repo:

Good caght. It seems that the option `--no-renames` should be added to the corresponding code.

@kevinrushforth Do you know which method about the git/hg should we revise?
If you don't know, I will do more search tomorrow.

kevinrushforth:
No, I don't know. And one thing I should mention is that there are almost certainly other cases where you do want `git diff` to do rename detection (`webrev` comes to mind), so I recommend only inhibiting rename detection when checking binary files.

lgxbslgx:
> I found the original PR where this was added https://github.com/openjdk/skara/pull/159. It looks like they never got around to implementing handling in the PR visitor. I think it's good if you fix that.

I don't think this patch would be integrated in short time. So I advice that we can use another patch to enable the existing binary check firstly, which is a bug in the past. @erikj79 What's your opinion?

erikj79:
> I don't think this patch would be integrated in short time. So I advice that we can use another patch to enable the existing binary check firstly, which is a bug in the past. @erikj79 What's your opinion?

I agree that it's better to change PullRequestCheckIssueVisitor to react to the existing binary check in a separate change.

lgxbslgx:
I filed [SKARA-1347](https://bugs.openjdk.java.net/browse/SKARA-1347) to follow up.

lgxbslgx:
I would like to close this PR and let other developers, who are familiar with Git/Mercurial, solve this issue.

PR #1248, opened 2021-11-30
1256: Use stream api to simplify the csr issue lookup

lgxbslgx:
Hi all,

This patch adds the common methods `Issue#linksWithRelationships`, `Issue#csrLink` and `Issue#csrIssue` and refactors the code to reduce the loop nested statements by using these common methods.

All the existing tests passed.

Best Regards,
-- Guoxiong

---------
### Progress

### Issue
 * [SKARA-1256](https://bugs.openjdk.java.net/browse/SKARA-1256): Use stream api to simplify the csr issue lookup

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1248/head:pull/1248` \
`$ git checkout pull/1248`

`$ git checkout pull/1248` \
`$ git pull https://git.openjdk.java.net/skara pull/1248/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1248`

`$ git pr show -t 1248`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1248.diff">https://git.openjdk.java.net/skara/pull/1248.diff</a>

</details>

### DIFF
--- a/bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
+++ b/bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
@@ -79,69 +79,62 @@ class CSRBot implements Bot, WorkItem {
                 continue;
             }
 
-            for (var link : jbsIssue.get().links()) {
-                var relationship = link.relationship();
-                if (relationship.isPresent() && relationship.get().equals("csr for")) {
-                    log.info("Found CSR for " + describe(pr));
+            var csr = jbsIssue.get().csrIssue().orElse(null);
+            if (csr == null) {
+                log.info("Not found CSR for " + describe(pr));
+            }
+            log.info("Found CSR for " + describe(pr) + ". It has id " + csr.id());
 
-                    var csr = link.issue().orElseThrow(
-                            () -> new IllegalStateException("Link with title 'csr for' does not contain issue")
-                    );
-
-                    log.info("CSR for " + describe(pr) + " has id " + csr.id());
-
-                    var resolution = csr.properties().get("resolution");
-                    if (resolution == null || resolution.isNull()) {
-                        if (!pr.labelNames().contains(CSR_LABEL)) {
-                            log.info("CSR issue resolution is null for " + describe(pr) + ", adding the CSR label");
-                            pr.addLabel(CSR_LABEL);
-                        } else {
-                            log.info("CSR issue resolution is null for " + describe(pr) + ", not removing the CSR label");
-                        }
-                        continue;
-                    }
-                    var name = resolution.get("name");
-                    if (name == null || name.isNull()) {
-                        if (!pr.labelNames().contains(CSR_LABEL)) {
-                            log.info("CSR issue resolution name is null for " + describe(pr) + ", adding the CSR label");
-                            pr.addLabel(CSR_LABEL);
-                        } else {
-                            log.info("CSR issue resolution name is null for " + describe(pr) + ", not removing the CSR label");
-                        }
-                        continue;
-                    }
-
-                    if (csr.state() != Issue.State.CLOSED) {
-                        if (!pr.labelNames().contains(CSR_LABEL)) {
-                            log.info("CSR issue state is not closed for " + describe(pr) + ", adding the CSR label");
-                            pr.addLabel(CSR_LABEL);
-                        } else {
-                            log.info("CSR issue state is not closed for " + describe(pr) + ", not removing the CSR label");
-                        }
-                        continue;
-                    }
-
-                    if (!name.asString().equals("Approved")) {
-                        if (name.asString().equals("Withdrawn")) {
-                            // This condition is necessary to prevent the bot from adding the CSR label again.
-                            // And the bot can't remove the CSR label automatically here.
-                            // Because the PR author with the role of Committer may withdraw a CSR that
-                            // a Reviewer had requested and integrate it without satisfying that requirement.
-                            log.info("CSR closed and withdrawn for " + describe(pr) + ", not revising (not adding and not removing) CSR label");
-                        } else if (!pr.labelNames().contains(CSR_LABEL)) {
-                            log.info("CSR issue resolution is not 'Approved' for " + describe(pr) + ", adding the CSR label");
-                            pr.addLabel(CSR_LABEL);
-                        } else {
-                            log.info("CSR issue resolution is not 'Approved' for " + describe(pr) + ", not removing the CSR label");
-                        }
-                        continue;
-                    }
-
-                    if (pr.labelNames().contains(CSR_LABEL)) {
-                        log.info("CSR closed and approved for " + describe(pr) + ", removing CSR label");
-                        pr.removeLabel(CSR_LABEL);
-                    }
+            var resolution = csr.properties().get("resolution");
+            if (resolution == null || resolution.isNull()) {
+                if (!pr.labelNames().contains(CSR_LABEL)) {
+                    log.info("CSR issue resolution is null for " + describe(pr) + ", adding the CSR label");
+                    pr.addLabel(CSR_LABEL);
+                } else {
+                    log.info("CSR issue resolution is null for " + describe(pr) + ", not removing the CSR label");
                 }
+                continue;
+            }
+            var name = resolution.get("name");
+            if (name == null || name.isNull()) {
+                if (!pr.labelNames().contains(CSR_LABEL)) {
+                    log.info("CSR issue resolution name is null for " + describe(pr) + ", adding the CSR label");
+                    pr.addLabel(CSR_LABEL);
+                } else {
+                    log.info("CSR issue resolution name is null for " + describe(pr) + ", not removing the CSR label");
+                }
+                continue;
+            }
+
+            if (csr.state() != Issue.State.CLOSED) {
+                if (!pr.labelNames().contains(CSR_LABEL)) {
+                    log.info("CSR issue state is not closed for " + describe(pr) + ", adding the CSR label");
+                    pr.addLabel(CSR_LABEL);
+                } else {
+                    log.info("CSR issue state is not closed for " + describe(pr) + ", not removing the CSR label");
+                }
+                continue;
+            }
+
+            if (!name.asString().equals("Approved")) {
+                if (name.asString().equals("Withdrawn")) {
+                    // This condition is necessary to prevent the bot from adding the CSR label again.
+                    // And the bot can't remove the CSR label automatically here.
+                    // Because the PR author with the role of Committer may withdraw a CSR that
+                    // a Reviewer had requested and integrate it without satisfying that requirement.
+                    log.info("CSR closed and withdrawn for " + describe(pr) + ", not revising (not adding and not removing) CSR label");
+                } else if (!pr.labelNames().contains(CSR_LABEL)) {
+                    log.info("CSR issue resolution is not 'Approved' for " + describe(pr) + ", adding the CSR label");
+                    pr.addLabel(CSR_LABEL);
+                } else {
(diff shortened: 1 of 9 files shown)


issuetracker/src/main/java/org/openjdk/skara/issuetracker/Issue.java
```diff
@@ -188,6 +188,50 @@ default URI nonTransformedWebUrl() {
...
+        return result;
+    }
+
+    /**
+     * Get the corresponding csr link of current issue if it exists.
+     * @return the csr link
+     */
+    default Optional<Link> csrLink() {
+        var links = linksWithRelationships(List.of("csr for"));
+        if (links != null && !links.isEmpty()) {
```
erikj79, line 215: From what I can tell, `links` can never be null here. 

I think this method could be expressed a bit more elegantly with stream usage, like this:
```
return links.stream().findAny();
```

issuetracker/src/main/java/org/openjdk/skara/issuetracker/Issue.java
```diff
@@ -188,6 +188,50 @@ default URI nonTransformedWebUrl() {
 
     List<Link> links();
 
+    /**
+     * Get the links which have the relationships listed in the arguments.
+     * @param relationships the link relationship
+     * @return the related links
+     */
+    default List<Link> linksWithRelationships(List<String> relationships) {
```
erikj79, line 196: I'm not sure there will ever be a need to filter links with multiple different relationships. Do you know of any such case that you are planning to convert?

The rest of this method could be converted to streams like this:

```
return links().stream()
    .filter(l -> link.relationship().isPresent() && relationships.contains(link.relationship().get())
    .toList();
```

issuetracker/src/main/java/org/openjdk/skara/issuetracker/Issue.java
```diff
@@ -188,6 +188,50 @@ default URI nonTransformedWebUrl() {
...
+        return Optional.empty();
+    }
+
+    /**
+     * Get the corresponding csr issue of current issue if it exists.
+     * @return the csr issue
+     */
+    default Optional<Issue> csrIssue() {
+        var csrLink = csrLink();
+        var csrIssue = Optional.<Issue>empty();
```
erikj79, line 228: This could be simplified to:

```
return csrLink().map(Link::issue);
```

erikj79 (commented):
I'm not sure about this change. The Issue interface seems over used and already contains a bunch of methods that do not apply to several implementors. I don't like the idea of adding even more such methods.

I think we need to try to extract a sub interface from Issue, which is specialized for bugtrackers or Jira or something. Not sure what to name it at this point though. All the links  methods should be moved there, and methods that definitely can't return a pull request, like IssueProject::issue should return the new interface instead. This is part of what I meant with this being hard to solve properly.

lgxbslgx:
@erikj79 Thanks for the comment.

From your suggestion in the code, I find that we don't need these 3 method to write the clean code. We can only use the stream api to simply the nest loop statement, which is more flexible.

It seems we need a method named `linksWithRelationships`, which contains a loop statement, to find the needed links. But actually, if we use the stream api, we only need a simple invoking statement.

I would like remove these three method and use the stream api to refactor the code.
And this issue title may change to `Use stream api to simplify the csr issue lookup`.

lgxbslgx:
> I'm not sure there will ever be a need to filter links with multiple different relationships. Do you know of any such case that you are planning to convert?

It can be used in the method `Backports#findMainIssue`:

```
// method Backports#findMainIssue
var links = issue.linksWithRelationships(List.of("backported by", "backport of"));
```

But If we use the stream api, it is not need now.

lgxbslgx:
@erikj79 I update the code by use the stream api just now.

erikj79:
> > I'm not sure there will ever be a need to filter links with multiple different relationships. Do you know of any such case that you are planning to convert?
> 
> It can be used in the method `Backports#findMainIssue`:
> 
> ```
> // method Backports#findMainIssue
> var links = issue.linksWithRelationships(List.of("backported by", "backport of"));
> ```
> 
> But If we use the stream api, it is not needed now.

Ah, right good point. I forgot about the two variants for every link type in JBS.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -133,20 +136,12 @@ private boolean isTargetBranchAllowed() {
...
-                log.warning("The CSR " + link + " of the issue " + issue + " does not exist");
-            } else {
-                break;
-            }
-        }
-        if (csr != null) {
+        org.openjdk.skara.issuetracker.Issue csr = jbsIssue.get().links().stream()
+                .filter(link -> link.relationship().isPresent() && "csr for".equals(link.relationship().get()))
+                .findAny().flatMap(Link::issue).orElse(null);
+        if (csr == null) {
```
erikj79, line 142: I just noticed this pattern being used a lot. Instead of .orElse(null), I would prefer if you let csr be an optional and used `csr.isPresent()` in the if statement. You will then need to use `csr.get()` to actually get the value, but this is safe as it's protected by the condition on isPresent(). If the variable in question is used a lot, you can choose resolve it with .get() into a new variable at the top of the if/else block.

This applies to several places where the same pattern is used.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CSRCommand.java
```diff
@@ -156,16 +161,10 @@ public void handle(PullRequestBot bot, PullRequest pr, CensusInstance censusInst
...
-        for (var link : jbsIssue.get().links()) {
-            var relationship = link.relationship();
-            if (relationship.isPresent() && relationship.get().equals("csr for")) {
-                csr = link.issue().orElseThrow(
-                        () -> new IllegalStateException("Link with title 'csr for' does not contain issue")
-                );
-            }
-        }
 
+        var csr = jbsIssue.get().links().stream()
```
erikj79, line 165: This particular pattern now occurs twice in this file. Perhaps worth moving to a private method at least, to reduce some duplication.

erikj79 (commented):
I think this kind of refactoring is good enough for now. I still think the Issue interface should get refactored along the lines I described at some point, so that concepts like different types of links can be handled by a single appropriate class/interface. The clash with the other "Issue" class which represents something completely different is very unfortunate as well.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
lgxbslgx: Fixed.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CSRCommand.java
lgxbslgx: Fixed.

bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
```diff
@@ -79,69 +80,66 @@ private String describe(PullRequest pr) {
...
-                        continue;
-                    }
-
-                    if (pr.labelNames().contains(CSR_LABEL)) {
-                        log.info("CSR closed and approved for " + describe(pr) + ", removing CSR label");
-                        pr.removeLabel(CSR_LABEL);
-                    }
+            var csrOptional = jbsIssue.get().links().stream()
+                    .filter(link -> link.relationship().isPresent() && "csr for".equals(link.relationship().get()))
+                    .findAny().flatMap(Link::issue);
```
magicus, line 85: I realize you'd be breaking abstractions by using csrLink() from CSRCommand, but maybe you can create a duplicated csrLink method here? I think it would help with readability, at least. And if we ever refactor this more, it would be a clear candidate to move to a class that is accessible from both here and CSRCommand.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -133,21 +136,13 @@ private boolean isTargetBranchAllowed() {
...
-            csr = link.issue().orElse(null);
-            if (csr == null) {
-                log.warning("The CSR " + link + " of the issue " + issue + " does not exist");
-            } else {
-                break;
-            }
-        }
-        if (csr != null) {
-            return Issue.fromStringRelaxed(csr.id() + ": " + csr.title());
+        var csr = jbsIssue.get().links().stream()
```
magicus, line 139: ... and here as well. It's really unfortunate that we have no good framework for where to place such common functionality. :-(

bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
lgxbslgx: Encapsulating such a method can be reused in the future. Fixed

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
lgxbslgx: Fixed.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CSRCommand.java
```diff
@@ -96,37 +102,42 @@ public void handle(PullRequestBot bot, PullRequest pr, CensusInstance censusInst
...
+                reply.println("determined that a [CSR](https://wiki.openjdk.java.net/display/csr/Main) request " +
+                        "is not needed for this pull request.");
+                return;
             }
-            // The issue has no csr or the csr has been withdrawn, the bot should just remove the csr label.
+
+            var csrOptional = csrLink.flatMap(Link::issue);
+            if (csrOptional.isEmpty()) {
+                // The csr link exists but the csr issue doesn't exist.
+                // We should remind the user to remove the link firstly.
```
erikj79, line 118: I missed this in the previous patch.

I don't think we can ask the user to remove a broken link in JBS. I'm pretty sure that would require admin actions. Here we should just treat it as if the link didn't exist and remove the label.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CSRCommand.java
```diff
@@ -96,37 +102,42 @@ public void handle(PullRequestBot bot, PullRequest pr, CensusInstance censusInst
...
+                return;
+            }
+            var csrIssue = csrOptional.get();
+            var resolution = csrIssue.properties().get("resolution");
+            if (resolution == null || resolution.isNull()
+                    || resolution.get("name") == null || resolution.get("name").isNull()
+                    || csrIssue.state() != Issue.State.CLOSED
+                    || !resolution.get("name").asString().equals("Withdrawn")) {
+                // The issue has a non-withdrawn csr issue, the bot should direct the user to withdraw the csr firstly.
+                reply.println("the issue for this pull request, [" + jbsIssue.get().id() + "](" + jbsIssue.get().webUrl() + "), has " +
```
erikj79, line 132: I missed this in the previous patch. The language of this reply needs some work. Here is my suggestion.

"The CSR requirement cannot be removed as there already is a CSR associated with the main issue of this pull request. Please withdraw the CSR [link] and then use the command '/csr unneeded' again."

lgxbslgx: Fixed.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CSRCommand.java
lgxbslgx: Fixed.

erikj79 (approved):

lgxbslgx:
@erikj79 thanks for the review.

/integrate

PR #784, opened 2020-09-04
Avoid posting Jira link comments multiple times

rwestberg:
Hi all,

Please review this change that avoids potentially posting the same Jira link comment multiple times.

Best regards,
Robin

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Change must be properly reviewed

### Reviewers
 * [Erik Helin](https://openjdk.java.net/census#ehelin) (@edvbld - **Reviewer**)

### Download
`$ git fetch https://git.openjdk.java.net/skara pull/784/head:pull/784`
`$ git checkout pull/784`

PR #1250, opened 2021-12-03
1258: Add review link/details to comments section

lgxbslgx:
Hi all,

This patch adds a comment to the jira issue when the corresponding PR is submitted. Because currently, the class `Issue` has no an api to remove a comment. So I need to create and implement it, too. And the test cases are added.

Best Regards,
-- Guoxiong

---------
### Progress

### Issue
 * [SKARA-1258](https://bugs.openjdk.java.net/browse/SKARA-1258): Add review link/details to comments section

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1250/head:pull/1250` \
`$ git checkout pull/1250`

`$ git checkout pull/1250` \
`$ git pull https://git.openjdk.java.net/skara pull/1250/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1250`

`$ git pr show -t 1250`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1250.diff">https://git.openjdk.java.net/skara/pull/1250.diff</a>

</details>

### DIFF
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
@@ -58,7 +58,7 @@ class IssueNotifier implements Notifier, PullRequestListener, RepositoryListener
 
     private final Logger log = Logger.getLogger("org.openjdk.skara.bots.notify");
 
-    private static final String pullRequestTip = "A pull request is submitted for review.";
+    private static final String pullRequestTip = "A pull request was submitted for review.";
 
     IssueNotifier(IssueProject issueProject, boolean reviewLink, URI reviewIcon, boolean commitLink, URI commitIcon,
                   boolean setFixVersion, Map<String, String> fixVersions, Map<String, List<String>> altFixVersions,
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
@@ -191,13 +191,13 @@ class IssueNotifier implements Notifier, PullRequestListener, RepositoryListener
             }
 
             realIssue.get().addLink(linkBuilder.build());
+        }
 
-            var alreadyPostedComment = realIssue.get().comments().stream()
-                    .filter(comment -> comment.author().equals(issueProject.issueTracker().currentUser()))
-                    .anyMatch(comment -> comment.body().contains(pullRequestTip) && comment.body().contains(pr.webUrl().toString()));
-            if (!alreadyPostedComment) {
-                realIssue.get().addComment(pullRequestToTextBrief(pr));
-            }
+        var alreadyPostedComment = realIssue.get().comments().stream()
+                .filter(comment -> comment.author().equals(issueProject.issueTracker().currentUser()))
+                .anyMatch(comment -> comment.body().contains(pullRequestTip) && comment.body().contains(pr.webUrl().toString()));
+        if (!alreadyPostedComment) {
+            realIssue.get().addComment(pullRequestToTextBrief(pr));
         }
     }
 
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2021, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -36,6 +36,7 @@ import org.openjdk.skara.vcs.openjdk.*;
 import java.io.*;
 import java.net.URI;
 import java.nio.file.Path;
+import java.time.format.DateTimeFormatter;
 import java.util.*;
 import java.util.logging.Logger;
 
@@ -56,6 +57,8 @@ class IssueNotifier implements Notifier, PullRequestListener, RepositoryListener
     private final String namespace;
 
     private final Logger log = Logger.getLogger("org.openjdk.skara.bots.notify");
+
+    private static final String pullRequestTip = "A pull request is submitted for review.";
 
     IssueNotifier(IssueProject issueProject, boolean reviewLink, URI reviewIcon, boolean commitLink, URI commitIcon,
                   boolean setFixVersion, Map<String, String> fixVersions, Map<String, List<String>> altFixVersions,
@@ -188,6 +191,13 @@ class IssueNotifier implements Notifier, PullRequestListener, RepositoryListener
             }
 
             realIssue.get().addLink(linkBuilder.build());
+
+            var alreadyPostedComment = realIssue.get().comments().stream()
+                    .filter(comment -> comment.author().equals(issueProject.issueTracker().currentUser()))
+                    .anyMatch(comment -> comment.body().contains(pullRequestTip) && comment.body().contains(pr.webUrl().toString()));
+            if (!alreadyPostedComment) {
+                realIssue.get().addComment(pullRequestToTextBrief(pr));
+            }
         }
     }
 
@@ -201,6 +211,21 @@ class IssueNotifier implements Notifier, PullRequestListener, RepositoryListener
 
         var link = Link.create(pr.webUrl(), "").build();
         realIssue.get().removeLink(link);
+
+        var postedComment = realIssue.get().comments().stream()
+                .filter(comment -> comment.author().equals(issueProject.issueTracker().currentUser()))
+                .filter(comment -> comment.body().contains(pullRequestTip) && comment.body().contains(pr.webUrl().toString()))
+                .findAny();
+        postedComment.ifPresent(comment -> realIssue.get().removeComment(comment));
+    }
+
+    private String pullRequestToTextBrief(PullRequest pr) {
+        var writer = new StringWriter();
+        var printer = new PrintWriter(writer);
+        printer.println(pullRequestTip + "\n");
+        printer.println("URL: " + pr.webUrl().toString());
+        printer.println("Date: " + pr.createdAt().format(DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss +0000")));
+        return writer.toString();
     }
 
     @Override
--- a/bots/notify/src/test/java/org/openjdk/skara/bots/notify/issue/IssueNotifierTests.java
+++ b/bots/notify/src/test/java/org/openjdk/skara/bots/notify/issue/IssueNotifierTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2020, 2021, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -46,6 +46,8 @@ import static org.openjdk.skara.issuetracker.Issue.State.OPEN;
 import static org.openjdk.skara.issuetracker.Issue.State.RESOLVED;
 
 public class IssueNotifierTests {
+    private static final String pullRequestTip = "A pull request is submitted for review.";
+
     private Set<String> fixVersions(Issue issue) {
         if (!issue.properties().containsKey("fixVersions")) {
             return Set.of();
@@ -196,15 +198,19 @@ public class IssueNotifierTests {
             pr.setBody("\n\n### Issue\n * [" + issue.id() + "](http://www.test.test/): The issue");
             TestBotRunner.runPeriodicItems(notifyBot);
 
-            // The issue should not yet contain a link to the PR
+            // The issue should not yet contain a link to the PR or a comment which contains the link to the PR
             var links = issue.links();
             assertEquals(0, links.size());
+            var comments = issue.comments();
+            assertEquals(0, comments.size());
 
             // Just a label isn't enough
             pr.addLabel("rfr");
             TestBotRunner.runPeriodicItems(notifyBot);
             links = issue.links();
             assertEquals(0, links.size());
+            comments = issue.comments();
+            assertEquals(0, comments.size());
 
             // Neither is just a comment
             pr.removeLabel("rfr");
@@ -213,16 +219,22 @@ public class IssueNotifierTests {
             TestBotRunner.runPeriodicItems(notifyBot);
             links = issue.links();
             assertEquals(0, links.size());
+            comments = issue.comments();
+            assertEquals(0, comments.size());
 
             // Both are needed
             pr.addLabel("rfr");
             TestBotRunner.runPeriodicItems(notifyBot);
 
-            // The issue should now contain a link to the PR
+            // The issue should now contain a link to the PR and a comment which contains the link to the PR
             links = issue.links();
             assertEquals(1, links.size());
             assertEquals(pr.webUrl(), links.get(0).uri().orElseThrow());
             assertEquals(reviewIcon, links.get(0).iconUrl().orElseThrow());
+            comments = issue.comments();
+            assertEquals(1, comments.size());
+            assertTrue(comments.get(0).body().contains(pullRequestTip));
+            assertTrue(comments.get(0).body().contains(pr.webUrl().toString()));
 
             // Add another issue
             var issue2 = issueProject.createIssue("This is another issue", List.of("Yes indeed"), Map.of("issuetype", JSON.of("Enhancement")));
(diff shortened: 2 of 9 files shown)


erikj79:
If you look in JiraIssue::addWebLink, it already uses a comment instead of a link in certain cases (specifically when the link is considered confidential). This is the behavior we now wish to emulate all the time, but in addition to the weblink. IMO, we just need to change this method a bit, so that it always calls ::addWebLinkAsComment and then skips the weblink part if needsSecurity is true.

I don't think we should remove comments when removing links. Issue comments are meant to represent a log of things happening to an Issue. The links section is meant to represent the current valid/relevant links. Removing the link when it's not relevant is good, but removing comments is like editing history and should only be done rare cases. In this situation, I would leave it to the user to clean up if they think it's needed.

Also, if you add new API calls to various REST providers, those need thorough testing which can be hard to perform. We unfortunately do not have any automated way of testing the interaction between Skara and Github/Gitlab/Jira.

lgxbslgx:
>  If you look in JiraIssue::addWebLink, it already uses a comment instead of a link in certain cases (specifically when the link is considered confidential). This is the behavior we now wish to emulate all the time, but in addition to the weblink. IMO, we just need to change this method a bit, so that it always calls ::addWebLinkAsComment and then skips the weblink part if needsSecurity is true.

I agree to adjust the method `addWebLinkAsComment` to complete this enhancement.

> I don't think we should remove comments when removing links. Issue comments are meant to represent a log of things happening to an Issue. The links section is meant to represent the current valid/relevant links. Removing the link when it's not relevant is good, but removing comments is like editing history and should only be done rare cases. In this situation, I would leave it to the user to clean up if they think it's needed.

Curently, the method `removeWebLink` will remove the corresponding comment. Please see the code below. The line `request.delete("/comment/" + comment.id()).execute()` will remove the related web link comment.

Because such comment is related to the weblink, I think this step is reasonable. But removing the arbitrary comment is not good. I agree that we shouldn't provide the api about removing comment.

```
    private void removeWebLink(Link link) {
        request.delete("/remotelink")
               .param("globalId", "skaralink=" + link.uri().orElseThrow().toString())
               .onError(e -> e.statusCode() == 404 ? Optional.of(JSON.object().put("already_deleted", true)) : Optional.empty())
               .execute();

        for (var comment : comments()) {
            var commentLink = parseWebLinkComment(comment);
            if (commentLink.isEmpty()) {
                continue;
            }
            if (!commentLink.get().uri().equals(link.uri())) {
                continue;
            }
            request.delete("/comment/" + comment.id()).execute();  // <---- here, remove the comment.
        }
    }

```

> Also, if you add new API calls to various REST providers, those need thorough testing which can be hard to perform. We unfortunately do not have any automated way of testing the interaction between Skara and Github/Gitlab/Jira.

I noticed it. Hope the automated tests can be built in the future.

magicus:
I don't like the thought of removing things from JBS. What is it you want to remove, and why? The Review link? As I argued in the bug report, I think it is useful to have and should not be removed.

erikj79:
> > I don't think we should remove comments when removing links. Issue comments are meant to represent a log of things happening to an Issue. The links section is meant to represent the current valid/relevant links. Removing the link when it's not relevant is good, but removing comments is like editing history and should only be done rare cases. In this situation, I would leave it to the user to clean up if they think it's needed.
> 
> Curently, the method `removeWebLink` will remove the corresponding comment. Please see the code below. The line `request.delete("/comment/" + comment.id()).execute()` will remove the related web link comment.
> 
> Because such comment is related to the weblink, I think this step is reasonable. But removing the arbitrary comment is not good. I agree that we shouldn't provide the api about removing comment.

You are right, I missed this, and thinking more about this, I think it's the correct behavior. That also means we don't need to change anything with regards to removing links in this enhancement.

> 
> > Also, if you add new API calls to various REST providers, those need thorough testing which can be hard to perform. We unfortunately do not have any automated way of testing the interaction between Skara and Github/Gitlab/Jira.
> 
> I noticed it. Hope the automated tests can be built in the future.

I wouldn't know where to start. I have created a couple of manual tests that interact directly with Github and Gitlab for functionality that I added. To run them you need to feed in a properties file with URLs and authentication data. See ManualForgeTests for an example.

erikj79:
> I don't like the thought of removing things from JBS. What is it you want to remove, and why? The Review link? As I argued in the bug report, I think it is useful to have and should not be removed.

The operations here are "add an issue" to a PR (which adds a review link to the issue), and remove an issue from a PR (which removes the review link from the issue). Both of these are already supported today and as Guoxiong pointed out to me, both link and comment are already automatically removed from the issue when the user does `/issue remove`.

lgxbslgx:
> If you look in JiraIssue::addWebLink, it already uses a comment instead of a link in certain cases (specifically when the link is considered confidential). This is the behavior we now wish to emulate all the time, but in addition to the weblink. IMO, we just need to change this method a bit, so that it always calls ::addWebLinkAsComment and then skips the weblink part if needsSecurity is true.

@erikj79 I re-read the code. If we make `addWebLinkAsComment` always be invoked, it may cause some unexpected behavior. 

For example, when a patch integrates, a link will be added to the issue (The code is at `IssueNotifier#onIntegratedPullRequest`). At the same time, a comment will be added to the issue (The code is at `IssueNotifier#onNewCommits`). If we enable the `addWebLinkAsComment`, two similar comments would be added to the same issue.

We can remove the related code in `IssueNotifier#onNewCommits` to let the bot only add one comment. But actually, the comment message of the `IssueNotifier#onNewCommits` is more readable and precise. If we always remove the related code like  `IssueNotifier#onNewCommits`, I can't find all the places to modify now and the change is more large and more difficult.

So I advice that we can add a warning comment in the api `removeComment` so that the developers use this api carefully.

> Also, if you add new API calls to various REST providers, those need thorough testing which can be hard to perform. We unfortunately do not have any automated way of testing the interaction between Skara and Github/Gitlab/Jira.

If you agree to add the api `removeComment`, this patch needs to be reviewed carefully according to the documentation of the `Github/Gitlab/Jira`.

----
When I type this comment, I receive the your new comment just now. If something is duplicated, just ignore it.

lgxbslgx:
> I don't like the thought of removing things from JBS. What is it you want to remove, and why? The Review link? As I argued in the bug report, I think it is useful to have and should not be removed.

I understand your meaning in the comment of [SKARA-1258](https://bugs.openjdk.java.net/browse/SKARA-1258). This patch is following your idea.

The original logic is:
1. When the user types `/issue add XXXX` at a comment of the PR, the bot will add a review link to the issue `XXXX`.
2. When the user types `/issue remove XXXX` at a comment of the PR, the bot will remove the review link which is added in the step 1. This removing step is necessary because the issue `XXXX` is not related to the PR if the user typed `/issue remove XXXX`.

The logic of this patch is:
1. Like the step 1 above. Add the review link and **a comment which makes the JBS can notify the watchers**.
2. Like the step 2 above. Remove the added link and **remove the added comment**.

The review link is added and removed as before. And this patch only cares about the comment.

erikj79:
> > If you look in JiraIssue::addWebLink, it already uses a comment instead of a link in certain cases (specifically when the link is considered confidential). This is the behavior we now wish to emulate all the time, but in addition to the weblink. IMO, we just need to change this method a bit, so that it always calls ::addWebLinkAsComment and then skips the weblink part if needsSecurity is true.
> 
> @erikj79 I re-read the code. If we make `addWebLinkAsComment` always be invoked, it may cause some unexpected behavior.
> 
> For example, when a patch integrates, a link will be added to the issue (The code is at `IssueNotifier#onIntegratedPullRequest`). At the same time, a comment will be added to the issue (The code is at `IssueNotifier#onNewCommits`). If we enable the `addWebLinkAsComment`, two similar comments would be added to the same issue.
> 
> We can remove the related code in `IssueNotifier#onNewCommits` to let the bot only add one comment. But actually, the comment message of the `IssueNotifier#onNewCommits` is more readable and precise. If we always remove the related code like `IssueNotifier#onNewCommits`, I can't find all the places to modify now and the change is more large and more difficult.
> 

I've now read through the code in question and you are partly right. First of all, internally at Oracle we are already using the code path where comments are used instead of issue links. The "needsSecurity" conditional basically means "is this Oracle confidential". You of course can't see this in action, but this makes me comfortable that it currently actually works. 

We did have an issue with duplicate comments being posted in the past. This was fixed in #784. I don't really like that solution, but it does the trick. Looking at actual (confidential) bugs, the onNewCommits call seems to run first, so we do get the nicer comment link.

Given this, we could get away with just changing addWebLink as I suggested. However, this would have the effect that also commit links would be added as both comments and links, which the enhancement request did not ask for. That's probably not desired.

> So I advice that we can add a warning comment in the api `removeComment` so that the developers use this api carefully.

If we need to add Issue::removeComment, then only implement it where it's going to be used, which is in JiraIssue. For GithubPullRequest and GitlabMergeRequest, better leave it not implemented than implemented but not tested. (Here the failed abstraction of Issue and lack of a sub interface for bug system issues strikes again.) If you go down this route, I won't be able to integrate this until I have time to test it properly. Given how little time I have to spend on Skara these days, that may be a while.

kevinrushforth:
> Given this, we could get away with just changing addWebLink as I suggested. However, this would have the effect that also commit links would be added as both comments and links, which the enhancement request did not ask for. That's probably not desired.

Actually, I _do_ think having both the link and the comment is desired. See the comment from David Homles in the JBS issue (and my comment agreeing with him).

erikj79:
> > Given this, we could get away with just changing addWebLink as I suggested. However, this would have the effect that also commit links would be added as both comments and links, which the enhancement request did not ask for. That's probably not desired.
> 
> Actually, I _do_ think having both the link and the comment is desired. See the comment from Magnus in the JBS issue (and my comment agreeing with him).

Sorry if I wasn't clear. We do want both link and comment for review links, but not for commit links. My suggested solution would apply to both kinds of links. The enhancement request was only for review links.

lgxbslgx:
> We do want both link and comment for review links, but not for commit links.

The comment for commit links is added by `IssueNotifier#onNewCommits`. It is a existing feature.

lgxbslgx:
> Given this, we could get away with just changing addWebLink as I suggested. However, this would have the effect that also commit links would be added as both comments and links, which the enhancement request did not ask for. That's probably not desired.

Do you mean that if we change the method `addWebLink` as your suggested, there will be two commit comments, which is unexpected?

lgxbslgx:
I have one thing to confirm. In a confidential issue, when the patch is integrated (commit links), which comment does you see?

```
Changeset: 
Author: 
Committer: 
Date: 
URL: 
```

or 

```
Remote link:
URL:
Summary:
Relationship:
```

From the code, I guess you can see the second style. Is my guess right?

erikj79:
> > Given this, we could get away with just changing addWebLink as I suggested. However, this would have the effect that also commit links would be added as both comments and links, which the enhancement request did not ask for. That's probably not desired.
> 
> Do you mean that if we change the method `addWebLink` as you suggested, there will be two similar comments for commit links, which is unexpected?

No, I seem to have mixed things up in this paragraph, please disregard it. There will never be duplicate link comments because #784 fixed that. I think my proposed fix in `addWebLink` should do the trick. To clarify, addWebLink would do the following.

1. If needsSecurity is false, add an IssueLink (same as today)
2. If a comment with this link doesn't already exist, add a comment with the link (new behavior when needsSecurity is false, otherwise same as today)

> I have one thing to confirm. In a confidential issue, when the patch is integrated (commit links), which comment does you see?

You see the first one, as onNewCommit seems to be executing first. I have searched back JBS and the newest bugs with where the second style comment can be found were fixed for JDK 16, so before #784 (in these issues both comments are present).

lgxbslgx:
> > From the code, I guess you can see the second style. Is my guess right?
>
> You see the first one, as onNewCommit seems to be executing first. I have searched back JBS and the newest bugs with where the second style comment can be found were fixed for JDK 16, so before #784 (in these issues both comments are present).

The `see the first one` confused me. Please see the following code analysis.

The method `parseWebLinkComment` can only identify the comment lines `Remote link:` and `URL:`. 

```
    private static final Pattern titlePattern = Pattern.compile("^Remote link: (.*)");
    private static final Pattern urlPattern = Pattern.compile("^URL: (.*)");
    private static final Pattern summaryPattern = Pattern.compile("^Summary: (.*)");
    private static final Pattern relationshipPattern = Pattern.compile("^Relationship: (.*)");

    private Optional<Link> parseWebLinkComment(Comment comment) {
        var lines = comment.body().lines().collect(Collectors.toList());
        if ((lines.size() < 2 ) || (lines.size() > 4)) {
            return Optional.empty();
        }
        var titleMatcher = titlePattern.matcher(lines.get(0));  // <----  Remote link: XXXX
        var urlMatcher = urlPattern.matcher(lines.get(1));   // <----  URL: XXXX
        if (!titleMatcher.matches() || !urlMatcher.matches()) {   // <---- Both `Remote link:` and `URL:` should exist
            return Optional.empty();   // <---- if not, it will return Optional.empty()
        }
        // ignore other code...
    }
``` 

So when the #784 added the following code, the variable `alreadyPosted` just can identify the comment which has `Remote link:` and `URL:`. But it can't identify the comment created by the method `onNewCommit`, which has `Changeset: Author: Committer: Date: URL: `.

```
    private void addWebLinkAsComment(Link link) {
        var alreadyPosted = comments().stream()
                                      .map(this::parseWebLinkComment)
                                      .filter(Optional::isPresent)
                                      .map(Optional::get)
                                      .anyMatch(l -> l.uri().equals(link.uri()) && l.title().equals(link.title()));
        if (alreadyPosted) {
            return;
        }
        // ignore other code...
    }
```

As @rwestberg mentioned in the #784:

> Please review this change that avoids potentially posting the same Jira link comment multiple times.

The #784 only can avoid potentially posting **the same** Jira link comment multiple times.

And the method `onNewCommit` has the following code which can identify the comment created by `addWebLinkAsComment`. It only checks the url and not checks the other pattern, which is not like `parseWebLinkComment`.

``` 
    // only check the url
    var alreadyPostedComment = existingComments.stream()
            .filter(comment -> comment.author().equals(issueProject.issueTracker().currentUser()))
            .anyMatch(comment -> comment.body().contains(hashUrl) || comment.body().contains(shortHashUrl));
    if (!alreadyPostedComment) {
        issue.addComment(commitNotification);
    }
```

---

From the code, we can simulate the run result.

**If the issues are public (I can see):**
1. if the `onNewCommit` runs earier than `onIntegratedPullRequest`.
  (1) The `onNewCommit` will create the comment as `Changeset: Author: Committer: Date: URL:`.
  (2) Then the `onIntegratedPullRequest` will  create the issue link for the commit.

  It is reasonable and expected.

2. if the `onIntegratedPullRequest` runs earier than `onNewCommit`.
  (1) The `onIntegratedPullRequest` will  create the issue link for the commit.
  (2) Then the `onNewCommit` will create the comment as `Changeset: Author: Committer: Date: URL:`.

  The result is the same as the point 1. And it is the situation I always see in the OpenJDK JBS.


**If the issues are confidential (not public) (I can't see):**
1. if the `onNewCommit` runs earier than `onIntegratedPullRequest`.
  (1) The `onNewCommit` will create the comment as `Changeset: Author: Committer: Date: URL:`.
  (2) Then the `onIntegratedPullRequest` runs. Because `parseWebLinkComment` can't identify the comment `Changeset: Author: Committer: Date: URL:`, so `onIntegratedPullRequest` creates another comment `Remote link:` and `URL:` for the commit.

  In this situation, there are two comments added. It is unexpected.

2. if the `onIntegratedPullRequest` runs earier than `onNewCommit`.  
  (1) The `onIntegratedPullRequest` runs. Because there is no comment now, so `onIntegratedPullRequest` creates comment `Remote link:  URL:` for the commit.
  (2) Then the `onNewCommit` runs. It only identifies the url, which can identify the comment `Remote link: URL:`. So it won't create the comment again.

  In this situation, only one comment like `Remote link:  URL:` is added to the issue. It is unexpected, too. Because @erikj79 said `see the first one`.

----

I think the problem is not in method `parseWebLinkComment` and `addWebLink`.  I re-read the code, it seems the problem is caused by the config `commitLink`.

In the public issues, the `commitLink` is `true`, so the issue link can be created by using the method `onIntegratedPullRequest`.
So the analysis above about this situation is right.

```
// method onIntegratedPullRequest
            if (commitLink) {   // <---- Here
                var linkBuilder = Link.create(repository.webUrl(hash), "Commit")
                                      .summary(repository.name() + "/" + hash.abbreviate());
                if (commitIcon != null) {
                    linkBuilder.iconTitle("Commit");
                    linkBuilder.iconUrl(commitIcon);
                }
                issue.addLink(linkBuilder.build());  // <---- if the commitLink is false, this line never runs.
            }
```

In the confidential issues, the `commitLink` is false, so the issue comment `Remote link:  URL:`  never be created because the `issue.addLink` never runs. And the method `onNewCommit` will always create the comment `Changeset: Author: Committer: Date: URL:`, which you can see in current confidential (oracle internal) issues.

Could I get your help to read the configuration about the `commitLink` to verify my suspection?

lgxbslgx:
> You are correct, we have "commitLink": "false" in the configuration for the confidential notifier.

@erikj79 Given this, if we change the method `addWebLink` as you suggested. We may get two comments in public issues because the method `addWebLinkAsComment` can't identify the comment added by the method `onNewCommit`. We need to remove the relate code about adding comment in `onNewCommit`. The related code, which adds both link and comment like `onNewCommit`, need to be removed, too.(I am not sure whether these code exist at other classes)

So I suggest that we can set the config `reviewLink` to `false`, too. And then add the comment or remove the comment when needed or unneeded. It follows the convention : if we need to add both the link and comment, we need to disable the link at the confidential issues.(such as "commitLink": "false").

> If we need to add Issue::removeComment, then only implement it where it's going to be used, which is in JiraIssue. For GithubPullRequest and GitlabMergeRequest, better leave it not implemented than implemented but not tested.

Agree.

bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
```diff
@@ -57,6 +58,8 @@
 
     private final Logger log = Logger.getLogger("org.openjdk.skara.bots.notify");
 
+    private static final String pullRequestTip = "A pull request is submitted for review.";
```
erikj79, line 61:
```suggestion
    private static final String pullRequestTip = "A pull request was submitted for review.";
```

erikj79:
I have given this a lot of thought, and I think I'm ok with your current solution. It's a bit unfortunate to have differently formatted comments for confidential vs non confidential, but it's good enough for now.

lgxbslgx:
I merged the master branch to re-run the pre-submit tests in macOS, which failed in the past. And ping for review. Thanks.

erikj79 (approved):

lgxbslgx:
@erikj79 Thanks for your review. Could I get your help to sponsor this patch?

/integrate

lgxbslgx:
Please notice to set the config `reviewLink` to `false`.

PR #1251, opened 2021-12-03
1269: Handle Jira host with path element in URL

erikj79:
The IssueTracker/JiraHost classes in Skara make the assumption that the base URL for Jira is just a hostname with no path elements. In order to use Skara with a Jira instance with a URL that has a path element, this needs to be fixed.

Instead of always basing REST call URLs on the configured base path and appending to it, there are many cases where the URIBuilder::setPath is used, which overwrites any existing path element. This is especially troublesome when combined with using IssueProject::webUrl as the base for these URLs.

I would like to fix this by adding a supported way of getting the base URI from an IssueTracker (IssueTracker::uri), replace all uses of ::webUrl for constructing other URLs with this new call, and replacing all ::setPath with ::appendPath (in IssueTracker/Jira related code).

This change will require all configurations to be careful not to add a trailing slash to the URL for the Jira instance.

I intend to run this for a while in the staging environment to make sure it works.

---------
### Progress

### Issue
 * [SKARA-1269](https://bugs.openjdk.java.net/browse/SKARA-1269): Handle Jira host with path element in URL

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1251/head:pull/1251` \
`$ git checkout pull/1251`

`$ git checkout pull/1251` \
`$ git pull https://git.openjdk.java.net/skara pull/1251/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1251`

`$ git pr show -t 1251`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1251.diff">https://git.openjdk.java.net/skara/pull/1251.diff</a>

</details>

### DIFF
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifierBuilder.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifierBuilder.java
@@ -116,7 +116,7 @@ class IssueNotifierBuilder {
     }
 
     IssueNotifier build() {
-        var jbsBackport = new JbsBackport(issueProject.webUrl(), vault);
+        var jbsBackport = new JbsBackport(issueProject.issueTracker().uri(), vault);
         return new IssueNotifier(issueProject, reviewLink, reviewIcon, commitLink, commitIcon,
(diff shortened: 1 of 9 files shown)


kevinrushforth (approved):
Looks good.

PR #1257, opened 2021-12-08
1275: An unknown sub-command gives a confusing error message

lgxbslgx:
Hi all,

When using the unknown command, the client will output the unexpected message. For example:

```
$ git-pr lis 123

error: unexpected input: 123
usage: git-pr [options] [<COMMAND>]
	-h, --help      Show help
	    --verbose   Turn on verbose output
	    --debug     Turn on debugging output
	    --version   Print the version of this tool
```

When the client can't identify the sub-command `git-pr lis`, it will execute the default command `git-pr help` and will pass `lis 123` as the argument to the `git-pr help` command. And the `git-pr help` can resolve only one argument, which is `lis` in this example, so it would output `error: unexpected input: 123`.

This patch will output more information to the user:

```
$ git-pr lis 123

Can't find the sub-command 'lis', the default sub-command 'help' will be executed.
The arguments [lis, 123] will be passed to the default sub-command.

error: unexpected input: 123
usage: git-pr [options] [<COMMAND>]
	-h, --help      Show help
	    --verbose   Turn on verbose output
	    --debug     Turn on debugging output
	    --version   Print the version of this tool
```

The added information can let the user know why the following output is shown.

Best Regards,
-- Guoxiong

---------
### Progress

### Issue
 * [SKARA-1275](https://bugs.openjdk.java.net/browse/SKARA-1275): An unknown sub-command gives a confusing error message

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1257/head:pull/1257` \
`$ git checkout pull/1257`

`$ git checkout pull/1257` \
`$ git pull https://git.openjdk.java.net/skara pull/1257/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1257`

`$ git pr show -t 1257`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1257.diff">https://git.openjdk.java.net/skara/pull/1257.diff</a>

</details>

### DIFF
--- a/args/src/main/java/org/openjdk/skara/args/MultiCommandParser.java
+++ b/args/src/main/java/org/openjdk/skara/args/MultiCommandParser.java
@@ -33,8 +33,9 @@ public class MultiCommandParser {
     private final String programName;
     private final String defaultCommand;
     private final Map<String, Command> subCommands;
+    private final boolean defaultCommandWarningEnabled;
 
-    public MultiCommandParser(String programName, List<Command> commands) {
+    public MultiCommandParser(String programName, List<Command> commands, boolean defaultCommandWarningEnabled) {
         var defaults = commands.stream().filter(Default.class::isInstance).collect(Collectors.toList());
         if (defaults.size() != 1) {
             throw new IllegalArgumentException("Expecting exactly one default command");
@@ -46,6 +47,7 @@ public class MultiCommandParser {
                                    .collect(Collectors.toMap(
                                            Command::name,
                                            Function.identity()));
+        this.defaultCommandWarningEnabled = defaultCommandWarningEnabled;
         if (!commands.stream().anyMatch(c -> c.name().equals("help"))) {
             this.subCommands.put("help", helpCommand());
         }
@@ -62,7 +64,7 @@ public class MultiCommandParser {
                 var forwardedArgs = Arrays.copyOfRange(args, 1, args.length);
                 return () -> p.main(forwardedArgs);
             }
-            if (!"git-webrev".equals(programName) && !"git webrev".equals(programName)) {
+            if (defaultCommandWarningEnabled) {
                 System.err.println("warning: unknown sub-command: " + args[0]);
                 System.err.println("the default sub-command '" + defaultCommand +
                         "' will be executed with the arguments " + Arrays.toString(args) + "\n");
--- a/args/src/main/java/org/openjdk/skara/args/MultiCommandParser.java
+++ b/args/src/main/java/org/openjdk/skara/args/MultiCommandParser.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2021, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -62,6 +62,10 @@ public class MultiCommandParser {
                 var forwardedArgs = Arrays.copyOfRange(args, 1, args.length);
                 return () -> p.main(forwardedArgs);
             }
+            System.err.println("Can't find the sub-command '" + args[0] +
+                    "', the default sub-command '" + defaultCommand + "' will be executed.");
+            System.err.println("The arguments " + Arrays.toString(args) +
+                    " will be passed to the default sub-command.\n");
         }
         return () -> subCommands.get(defaultCommand).main(args);
     }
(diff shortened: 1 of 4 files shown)


erikj79:
The new message does explain in great detail how the implementation interprets the given input, but I think it's unnecessarily verbose. I think something like this would be clear enough:

```
$ git-pr lis 123

error: unknown sub-command: lis
error: unexpected input: 123
usage: git-pr [options] [<COMMAND>]
	-h, --help      Show help
	    --verbose   Turn on verbose output
	    --debug     Turn on debugging output
	    --version   Print the version of this tool
```

However, not all users of MultiCommandParser have the help command as default. The webrev command does not and printing all of this text when running `webrev` with no explicit sub-command would be a regression IMO. Printing this error would need to be made conditional on if a sub-command is required or not. For `pr` it is, while for `webrev` it's not.

lgxbslgx:
If we only output `error: unknown sub-command: lis`, the following output is confused:

```
error: unexpected input: 123
usage: git-pr [options] [<COMMAND>]  // <-- here should be `git-pr help [options] [<COMMAND>] `. See SKARA-1274
	-h, --help      Show help
	    --verbose   Turn on verbose output
	    --debug     Turn on debugging output
	    --version   Print the version of this tool
```

The user would confuse why the client outputs the usage of the `git-pr help`. Because the user expects the following message.

```
error: unknown sub-command: lis

Available sub-commands are:
- apply
- approve
- cc
```

So I think it is necessary to let user know what is happening. As for `git webrev`, it will output the following message in my patch. Because the `git-webrev generate` will execute.

```
Can't find the sub-command 'lis', the default sub-command 'generate' will be executed.
The arguments [lis, 123] will be passed to the default sub-command.
```

Now I suggest the following information:

```
$ git-pr lis 123

error: unknown sub-command: lis
the default sub-command 'help' will be executed with the arguments [lis, 123].

error: unexpected input: 123
usage: git-pr [options] [<COMMAND>]
	-h, --help      Show help
	    --verbose   Turn on verbose output
	    --debug     Turn on debugging output
	    --version   Print the version of this tool
```

```
$ git-webrev lis 123

error: unknown sub-command: lis
the default sub-command 'generate' will be executed with the arguments [lis, 123].

error: unexpected input: 123
usage: git webrev [options] [<FILE>]
// ignore
```

lgxbslgx:
Because the code is still running and not stop. It is good to change the `error` to `warning`. The message could be:

```
$ git-pr lis 123

warning: unknown sub-command: lis
the default sub-command 'help' will be executed with the arguments [lis, 123].

error: unexpected input: 123
usage: git-pr [options] [<COMMAND>]
// ignore
```

erikj79:
I like the last suggestion much better, but I still don't think it should be printed for `git webrev`. Running webrev without a sub-command is normal procedure and should not result in a warning.

lgxbslgx:
@erikj79 I updated the code just now. The new code uses the `warning` and excludes the `git webrev` command as you suggested.

lgxbslgx:
I merged the master branch to re-run the pre-submit tests in macOS, which failed in the past. And ping for review. Thanks.

args/src/main/java/org/openjdk/skara/args/MultiCommandParser.java
```diff
@@ -62,6 +62,11 @@ public Executable parse(String[] args) {
                 var forwardedArgs = Arrays.copyOfRange(args, 1, args.length);
                 return () -> p.main(forwardedArgs);
             }
+            if (!"git-webrev".equals(programName) && !"git webrev".equals(programName)) {
```
erikj79, line 65: I was expecting a more elegant solution to this. I would add a property/method on the Command interface that  MutliCommandParser can use to check this. Something like Command::defaultCommandExpected.

lgxbslgx: > I was expecting a more elegant solution to this. I would add a property/method on the Command interface that MutliCommandParser can use to check this. Something like Command::defaultCommandExpected.

Good idea. Should we add it in this patch? Or we should file another issue to implement it?

erikj79: This was meant as a suggestion for this patch. I don't think the current solution is ok to go in.

I'm not sure about the name of the new method, it's hard to come up with a good name for this property of a command. It's trying to express if the default sub-command is expected to be run with arguments or not, or if it takes arguments at all, but putting all that in a method name becomes quite a mouthful.

lgxbslgx: What about `defaultCommandWarningEnabled`? Actually, the default sub-command is always invoked if no sub-command is provided. But the warning message are outputted optionally.

lgxbslgx: What about `defaultCommandWarningEnabled`? Actually, the default sub-command is always invoked if no sub-command is provided. But the warning message are outputted optionally.

erikj79: Works for me. Thanks.

lgxbslgx: I updated the code just now by using the property `defaultCommandWarningEnabled`.

erikj79 (approved):

PR #1262, opened 2021-12-15
1288: Adjust the implementation of the 'reviewers' command to match the documentation

lgxbslgx:
Hi all,

The `reviewers` command [documentation](https://wiki.openjdk.java.net/display/SKARA/Pull+Request+Commands#PullRequestCommands-/reviewers) states:

> Description
> Require that at least N users with given role (defaults to Author) review the pull request before it can be integrated. The requirements are in addition to the ones specified by the .jcheck/conf file.

But the implementation doesn't match the documentation now, especially the bug reported at [JDK-PULL-6752](https://github.com/openjdk/jdk/pull/6752) ([SKARA-1288](https://bugs.openjdk.java.net/browse/SKARA-1288) also recorded this bug).

This patch fixes the bug and revises the code to match the documentation.

The logic has small change(Please see the following code). In the past, the reviewer number of the role will be the sum of the `updatedLimits.get(r)` and `remainingAdditional`. But I don't know it meets our expectation. Because the higher roles can reduce the `remainingAdditional`, I think the remaining number should be only set to `updatedLimits` but not added to `updatedLimits`. This matches the documentation `Require that at least N users with given role (defaults to Author) review the pull request`, especially the key work `at least`.

```
-                updatedLimits.replace(r, updatedLimits.get(r) + remainingAdditional);
+                if (remainingAdditional > updatedLimits.get(r)) {
+                    // Set the number for the lower roles to remove.
+                    remainingRemovals = remainingAdditional - updatedLimits.get(r);
+                   // The new number of the reviewers should larger or equal than the requirement of the '.jcheck/conf' file.
+                    // Because the '.jcheck/conf' file means the minimal reviewer requirement.
+                   updatedLimits.replace(r, remainingAdditional);
+              }
```

The other code changes are listed below:
- add negative number check.
- remove the updated limits check. Because it is not needed after adding the negative number check.
- adjust the reply message, always output the concrete reviewers. Because the merged logic of jcheck config and user `reviewers` command is a little hard, we should let the user know the result directly instead of guessing the result.
- `decrease lower roles` in the method `ReviewersTracker#updatedRoleLimits` doesn't work as expected. I fix it.
- add more test cases.

Best Regards,
-- Guoxiong

---------
### Progress

### Issue
 * [SKARA-1288](https://bugs.openjdk.java.net/browse/SKARA-1288): Adjust the implementation of the 'reviewers' command to match the documentation

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1262/head:pull/1262` \
`$ git checkout pull/1262`

`$ git checkout pull/1262` \
`$ git pull https://git.openjdk.java.net/skara pull/1262/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1262`

`$ git pr show -t 1262`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1262.diff">https://git.openjdk.java.net/skara/pull/1262.diff</a>

</details>

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/ReviewersCommand.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/ReviewersCommand.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2020, 2021, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -97,6 +97,11 @@ public class ReviewersCommand implements CommandHandler {
             reply.println("Cannot increase the required number of reviewers above 10 (requested: " + numReviewers + ")");
             return;
         }
+        if (numReviewers < 0) {
+            showHelp(reply);
+            reply.println("Cannot decrease the required number of reviewers below 0 (requested: " + numReviewers + ")");
+            return;
+        }
 
         String role = "authors";
         if (splitArgs.length > 1) {
@@ -124,24 +129,19 @@ public class ReviewersCommand implements CommandHandler {
         }
 
         var updatedLimits = ReviewersTracker.updatedRoleLimits(censusInstance.configuration(), numReviewers, role);
-        if (updatedLimits.get(role) > numReviewers) {
-            showHelp(reply);
-            reply.println("Number of required reviewers of role " + role + " cannot be decreased below " + updatedLimits.get(role));
-            return;
-        }
 
         reply.println(ReviewersTracker.setReviewersMarker(numReviewers, role));
         var totalRequired = updatedLimits.values().stream().mapToInt(Integer::intValue).sum();
-        reply.print("The number of required reviews for this PR is now set to " + totalRequired);
+        reply.print("The total number of required reviews for this PR, " +
+                "which is combined by the configuration and the command, is now set to " + totalRequired);
 
         // Create a helpful message regarding the required distribution (if applicable)
         var nonZeroDescriptions = updatedLimits.entrySet().stream()
                 .filter(entry -> entry.getValue() > 0)
                 .map(entry -> entry.getValue() + " of role " + entry.getKey())
                 .collect(Collectors.toList());
-        if (nonZeroDescriptions.size() > 1) {
-            nonZeroDescriptions.remove(nonZeroDescriptions.size() - 1);
-            reply.print(" (with at least " + String.join(", ", nonZeroDescriptions) + ")");
+        if (nonZeroDescriptions.size() > 0) {
+            reply.print(" (with " + String.join(", ", nonZeroDescriptions) + ")");
         }
 
         reply.println(".");
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/ReviewersCommand.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/ReviewersCommand.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2020, 2022, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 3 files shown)


kevinrushforth:
Unless there is a corner case that I haven't hit, the implementation is working as expected now. I will need to look at your proposed change more closely, but I don't see the need to change the existing logic.

What problem are you trying to solve?

erikj79:
> Unless there is a corner case that I haven't hit, the implementation is working as expected now. I will need to look at your proposed change more closely, but I don't see the need to change the existing logic.
> 
> What problem are you trying to solve?

In the linked PR, they tried to run `/reviewers 2 reviewer`, which gave a very weird error. At least that part is completely wrong in the current implementation. There isn't really any test coverage for using the command with a role argument. As for the rest, I'm still investigating. 

The command as implemented does not do what the documentation at https://wiki.openjdk.java.net/display/SKARA/Pull+Request+Commands#PullRequestCommands-/reviewers says it's supposed to do, that I definitely agree on. I think one point of confusion is if multiple command invocations should be added together, or just override the previous one, and should the override apply globally or just to the specified role? 

Before we go any further with an implementation, I think we need to agree on what the desired behavior actually is. The bug comments seems like a good forum for suggesting and commenting on bot behavior.

lgxbslgx:
Keep the patch alive. And waiting for comments in the issue or this PR.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/ReviewersCommand.java
```diff
@@ -124,24 +129,19 @@ public void handle(PullRequestBot bot, PullRequest pr, CensusInstance censusInst
...
-            showHelp(reply);
-            reply.println("Number of required reviewers of role " + role + " cannot be decreased below " + updatedLimits.get(role));
-            return;
-        }
 
         reply.println(ReviewersTracker.setReviewersMarker(numReviewers, role));
         var totalRequired = updatedLimits.values().stream().mapToInt(Integer::intValue).sum();
-        reply.print("The number of required reviews for this PR is now set to " + totalRequired);
+        reply.print("The total number of required reviews for this PR, " +
+                "which is combined by the configuration and the command, is now set to " + totalRequired);
```
erikj79, line 136: I would suggest:

"The total number of required reviews for this PR (including the jcheck configuration and the last /reviewers command) is now set to"

bots/pr/src/main/java/org/openjdk/skara/bots/pr/ReviewersTracker.java
```diff
@@ -59,19 +60,30 @@ static String setReviewersMarker(int numReviewers, String role) {
                     break;
                 }
             } else {
-                updatedLimits.replace(r, updatedLimits.get(r) + remainingAdditional);
+                if (remainingAdditional > updatedLimits.get(r)) {
+                    // Set the number for the lower roles to remove.
+                    remainingRemovals = remainingAdditional - updatedLimits.get(r);
+                    // The new number of the reviewers should larger or equal than the requirement of the '.jcheck/conf' file.
```
erikj79, line 66: I would suggest moving this comment to before line 63 and express it something like this:

// The new value cannot the lower than the value in .jcheck/conf

bots/pr/src/main/java/org/openjdk/skara/bots/pr/ReviewersTracker.java
```diff
@@ -59,19 +60,30 @@ static String setReviewersMarker(int numReviewers, String role) {
...
         Collections.reverse(roles);
-        var remainingRemovals = count;
         for (var r : roles) {
             if (!r.equals(role)) {
-                var removed = Math.max(0, updatedLimits.get(r) - remainingRemovals);
+                var originalVal = updatedLimits.get(r);
+                var removed = Math.max(0, originalVal - remainingRemovals);
                 updatedLimits.replace(r, removed);
-                remainingRemovals -= removed;
+                remainingRemovals -= (originalVal - updatedLimits.get(r));
```
erikj79, line 86: Maybe use the local variable `removed` instead of updatedLimits.get(r). I think that reads better.

erikj79 (commented):
Just a few nits on language in replies and comments. Copyright years need to be updated. Otherwise this looks really good!

bots/pr/src/main/java/org/openjdk/skara/bots/pr/ReviewersCommand.java
lgxbslgx: Fixed

bots/pr/src/main/java/org/openjdk/skara/bots/pr/ReviewersTracker.java
lgxbslgx: Fixed

bots/pr/src/main/java/org/openjdk/skara/bots/pr/ReviewersTracker.java
lgxbslgx: Fixed

lgxbslgx:
@erikj79 Thanks for the review. I updated the code just now.

erikj79 (approved):

lgxbslgx:
@erikj79 Could I get your help to sponsor this patch? Thanks a lot.

/integrate

PR #1274, opened 2022-01-13
1308: CSRBot only reacts to approved CSR if PR is touched

erikj79:
When a CSR issue is approved, any associated PR is only updated once someone touches the PR. The CSRBot will not evaluate anything about a PR unless there is a change to the PR. Since a CSR issue update only happens in JBS, the CSRBot cannot be limited by PR updates, it needs to check Jira for changes.

This patch removes the use of the PullRequestUpdateCache to limit when the bot evaluates a PR. If this turns out to generate too much load on JBS, we will have to implement a limiter in some way, but I don't think it will be a big issue in practice.

---------
### Progress

### Issue
 * [SKARA-1308](https://bugs.openjdk.java.net/browse/SKARA-1308): CSRBot only reacts to approved CSR if PR is touched

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1274/head:pull/1274` \
`$ git checkout pull/1274`

`$ git checkout pull/1274` \
`$ git pull https://git.openjdk.java.net/skara pull/1274/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1274`

`$ git pr show -t 1274`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1274.diff">https://git.openjdk.java.net/skara/pull/1274.diff</a>

</details>

### DIFF
--- a/bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
+++ b/bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
@@ -70,10 +70,6 @@ class CSRBot implements Bot, WorkItem {
         var prs = repo.pullRequests();
 
         for (var pr : prs) {
-            if (!cache.needsUpdate(pr)) {
-                continue;
-            }
-
             var issue = org.openjdk.skara.vcs.openjdk.Issue.fromStringRelaxed(pr.title());
             if (issue.isEmpty()) {
                 log.info("No issue found in title for " + describe(pr));
--- a/bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
+++ b/bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
@@ -25,7 +25,6 @@ package org.openjdk.skara.bots.csr;
 import org.openjdk.skara.bot.*;
 import org.openjdk.skara.forge.HostedRepository;
 import org.openjdk.skara.forge.PullRequest;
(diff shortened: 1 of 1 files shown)


bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
```diff
@@ -70,10 +67,6 @@ private String describe(PullRequest pr) {
         var prs = repo.pullRequests();
 
         for (var pr : prs) {
-            if (!cache.needsUpdate(pr)) {
-                continue;
-            }
```
kevinrushforth, line 75: An alternative to eliminating this check would be to poll JBS either if the PR was updated or if the PR has a csr label. But as you say, maybe it won't be a performance problem, in which case your proposed change is better.

kevinrushforth (approved):
Looks good. I left one comment about a possible alternative, but it might not be worth doing.

bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
erikj79: Note that we also want to react if a CSR issue is created when the label isn't there. In that case we automatically add the CSR label.

kevinrushforth: Yeah, that's the main reason your fix is better.

PR #1280, opened 2022-01-26
1329: incorrect hgupdate label for JDK 7 updates

erikj79:
This patch adds special handling of Backports for a couple of historic releases where we have 30 or more builds (and no BPRs).

---------
### Progress

### Issue
 * [SKARA-1329](https://bugs.openjdk.java.net/browse/SKARA-1329): incorrect hgupdate label for JDK 7 updates

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1280/head:pull/1280` \
`$ git checkout pull/1280`

`$ git checkout pull/1280` \
`$ git pull https://git.openjdk.java.net/skara pull/1280/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1280`

`$ git pr show -t 1280`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1280.diff">https://git.openjdk.java.net/skara/pull/1280.diff</a>

</details>

### DIFF
--- a/jbs/src/main/java/org/openjdk/skara/jbs/Backports.java
+++ b/jbs/src/main/java/org/openjdk/skara/jbs/Backports.java
@@ -308,18 +308,22 @@ public class Backports {
                     ret.add(jdkVersion.feature() + "+updates-openjdk");
                 }
             } else if (numericFeature == 7 || numericFeature == 8) {
-                var resolvedInBuild = jdkVersion.resolvedInBuild();
-                if (resolvedInBuild.isPresent()) {
-                    if (!resolvedInBuild.get().equals("team")) { // Special case - team build resolved are ignored
-                        int resolvedInBuildNumber = jdkVersion.resolvedInBuildNumber();
-                        if (resolvedInBuildNumber < 31) {
-                            ret.add(jdkVersion.feature());
-                        } else if (resolvedInBuildNumber < 60) {
-                            ret.add(jdkVersion.feature() + "+bpr");
-                        }
-                    }
-                } else {
+                if (bprException(jdkVersion, numericFeature)) {
                     ret.add(jdkVersion.feature());
+                } else {
+                    var resolvedInBuild = jdkVersion.resolvedInBuild();
+                    if (resolvedInBuild.isPresent()) {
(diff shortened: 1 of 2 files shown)


jbs/src/main/java/org/openjdk/skara/jbs/Backports.java
```diff
@@ -308,18 +308,22 @@ private static boolean isFixed(Issue issue) {
...
+                if (bprException(jdkVersion, numericFeature)) {
+                    ret.add(jdkVersion.feature());
+                } else {
+                    var resolvedInBuild = jdkVersion.resolvedInBuild();
+                    if (resolvedInBuild.isPresent()) {
+                        if (!resolvedInBuild.get().equals("team")) { // Special case - team build resolved are ignored
+                            int resolvedInBuildNumber = jdkVersion.resolvedInBuildNumber();
+                            if (resolvedInBuildNumber < 30) {
+                                ret.add(jdkVersion.feature());
+                            } else if (resolvedInBuildNumber < 60) {
```
magicus, line 320: In the JBS comment you stated that the spec says BPR builds should only be < 39. Did you chose not to change the current formulation, or did you forget?

magicus (approved):
LGTM barring my question about the upper bound for BPRs.

jbs/src/main/java/org/openjdk/skara/jbs/Backports.java
erikj79: I initially changed this to 40, as the specification we have says BPR between 30 and 39. The problem is that this leaves a gap between 41 and 59 which is undefined. Builds 60 and up are special and should never be considered for hgupdate-sync, but after contemplating this, I agree with the original author's conclusion that the best way to deal with 41-59 is to assume they are also BPRs. In reality, I don't think we have any release with builds in that range, so it doesn't actually matter.

magicus: That's okay for me. Just wanted to check that this was a conscious decision. (Maybe perhaps it should be commented upon in the code?)

PR #1286, opened 2022-02-15
1350: Altfixversions isn't checking Backport resolution

erikj79:
This patch fixes a flaw in the altfixversion feature. Currently a backport with an altfixversion is considered valid if the state is "resolved". There are two problems with this. It misses backports that are in state closed. It could also cause false positives by accepting backports that aren't set to resolution "Fixed".

To fix this, I've introduced a new method on the Issue interface "isFixed()". The default implementation (which will get used in testing) just delegates to isResolved(). For JiraIssue, it checks both the state and the resolution.

---------
### Progress

### Issue
 * [SKARA-1350](https://bugs.openjdk.java.net/browse/SKARA-1350): Altfixversions isn't checking Backport resolution

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1286/head:pull/1286` \
`$ git checkout pull/1286`

`$ git checkout pull/1286` \
`$ git pull https://git.openjdk.java.net/skara pull/1286/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1286`

`$ git pr show -t 1286`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1286.diff">https://git.openjdk.java.net/skara/pull/1286.diff</a>

</details>

### DIFF
--- a/issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraIssue.java
+++ b/issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraIssue.java
@@ -181,6 +181,21 @@ public class JiraIssue implements Issue {
         }
     }
 
+    /**
+     * A Jira issue is considered fixed if it's either resolved or closed and
+     * the resolution is "Fixed".
+     */
+    @Override
+    public boolean isFixed() {
+        if (isResolved() || isClosed()) {
+            var resolution = json.get("fields").get("resolution");
+            if (!resolution.isNull()) {
+                return "Fixed".equals(resolution.get("name").asString());
+            }
+        }
+        return false;
+    }
+
     private Map<String, String> availableTransitions() {
         var transitions = request.get("/transitions").execute();
         return transitions.get("transitions").stream()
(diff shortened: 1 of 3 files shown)


issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraIssue.java
```diff
@@ -181,6 +181,21 @@ public State state() {
         }
     }
 
+    /**
+     * A Jira issue is considered fixed if it's either resolved or closed and
+     * the resolution is "Fixed".
+     */
+    @Override
+    public boolean isFixed() {
+        if (isResolved() || isClosed()) {
```
kevinrushforth, line 190: This test might not be needed -- the resolution will be empty or null for unresolved issues. I don't see any harm in leaving it if you prefer to have the extra check.

kevinrushforth (approved):
Looks good.

issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraIssue.java
erikj79: I agree that JBS should guarantee this to be true. I went back and forth on this myself, but in the end I think it's clearer like this.

PR #1287, opened 2022-02-16
1352: Configure GithubHost as offline

erikj79:
As a followup to [SKARA-1344](https://bugs.openjdk.java.net/browse/SKARA-1344), to better support the notifier to act through a mirror, we need another configuration option. When configuring a forge for the "originalrepository" option, we need to make it possible to not need to make any remote calls to that forge.

Currently, when Skara creates a HostedRepository from a forge, it will always make a call to validate and initiate things. At least for GithubRepositories, this call isn't strictly necessary (and one could argue that it should be possible to get around this for Gitlab as well). Most callers for Forge::repository do expect this validation however, so we do not want to change the default behavior.

I propose adding a new configure parameter to github forge, "offline", which defaults to false. If set to true, then it indicates that this forge is expected to be offline and the caller is not expected to try to perform any remote operations. This concept could certainly be expanded, but for now, the only thing we explicitly need to prevent is that validation/initialization when creating a repository. I would rather not spend more time implementing a more complete feature that is unlikely to be used.

---------
### Progress

### Issue
 * [SKARA-1352](https://bugs.openjdk.java.net/browse/SKARA-1352): Configure GithubHost as offline

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1287/head:pull/1287` \
`$ git checkout pull/1287`

`$ git checkout pull/1287` \
`$ git pull https://git.openjdk.java.net/skara pull/1287/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1287`

`$ git pr show -t 1287`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1287.diff">https://git.openjdk.java.net/skara/pull/1287.diff</a>

</details>

### DIFF
--- a/forge/src/main/java/org/openjdk/skara/forge/Forge.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/Forge.java
@@ -59,7 +59,7 @@ public interface Forge extends Host {
     }
 
     static Forge from(String name, URI uri) {
-        return from(name, uri, (Credential) null);
+        return from(name, uri, null, null);
     }
 
     static Forge from(String name, URI uri, JSONObject configuration) {
--- a/forge/src/main/java/org/openjdk/skara/forge/Forge.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/Forge.java
@@ -59,7 +59,11 @@ public interface Forge extends Host {
     }
 
     static Forge from(String name, URI uri) {
-        return from(name, uri, null);
+        return from(name, uri, (Credential) null);
+    }
+
+    static Forge from(String name, URI uri, JSONObject configuration) {
+        return from(name, uri, null, configuration);
     }
 
     static Optional<Forge> from(URI uri, Credential credential, JSONObject configuration) {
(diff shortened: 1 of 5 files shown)


forge/src/main/java/org/openjdk/skara/forge/Forge.java
```diff
@@ -59,7 +59,11 @@ static Forge from(String name, URI uri, Credential credential) {
     }
 
     static Forge from(String name, URI uri) {
-        return from(name, uri, null);
+        return from(name, uri, (Credential) null);
```
magicus, line 62: Not entirely sure I like this. Maybe "from" has been overloaded too much?

magicus: Or you could return `from(name, uri, null, null)`, could you not? Anything but casting null...

magicus (approved):
LGTM, barring the null casting part. :)

magicus (approved):
Much better, thank you!

PR #1297, opened 2022-04-07
SKARA-1096: New command and label for JEPs, similar to CSR

lgxbslgx:
Hi all,

This patch implements the features about the `/jep` command and JEP bot. The detailed designs and discussions are listed at [1][2].

Best Regards,
-- Guoxiong

[1] https://mail.openjdk.java.net/pipermail/skara-dev/2021-December/005481.html
[2] https://mail.openjdk.java.net/pipermail/skara-dev/2022-March/005770.html

---------
### Progress

### Issue
 * [SKARA-1096](https://bugs.openjdk.java.net/browse/SKARA-1096): New command and label for JEPs, similar to CSR

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1297/head:pull/1297` \
`$ git checkout pull/1297`

`$ git checkout pull/1297` \
`$ git pull https://git.openjdk.java.net/skara pull/1297/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1297`

`$ git pr show -t 1297`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1297.diff">https://git.openjdk.java.net/skara/pull/1297.diff</a>

</details>


bots/notify/src/test/java/org/openjdk/skara/bots/notify/issue/IssueNotifierTests.java
```diff
@@ -47,6 +47,7 @@
 import static org.openjdk.skara.issuetracker.jira.JiraProject.SUBCOMPONENT;
 
 public class IssueNotifierTests {
+    private static final String JEP_NUMBER = "customfield_10701";
```
erikj79, line 50: You should be able to import this from JiraProject instead of duplicating the definition.

bots/pr/src/test/java/org/openjdk/skara/bots/pr/CheckTests.java
```diff
@@ -41,6 +41,8 @@
 import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
 class CheckTests {
+    private static final String JEP_NUMBER = "customfield_10701";
```
erikj79, line 44: It should be possible to import this from JiraProject as well.

issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraProject.java
```diff
@@ -433,6 +434,18 @@ public Issue createIssue(String title, List<String> body, Map<String, JSONValue>
         }
     }
 
+    @Override
+    public Optional<Issue> jepIssue(String jepId) {
```
erikj79, line 438: Have you verified that this works?

bots/pr/src/main/java/org/openjdk/skara/bots/pr/JEPCommand.java
```diff
@@ -0,0 +1,163 @@
...
+            showHelp(reply);
+            return;
+        }
+
+        var labelNames = pr.labelNames();
+        if ("unneeded".equals(args) || "uneeded".equals(args)) {
+            if (pr.author().equals(command.user()) && !censusInstance.isReviewer(command.user())) {
+                reply.println("only [Reviewers](https://openjdk.java.net/bylaws#reviewer) can determine that a JEP request is not needed.");
+                return;
+            }
```
erikj79, line 99: I don't think I agree with this check. If the author is able to link the PR to a JEP, then the author should be able to correct a mistakenly linked JEP.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/JEPCommand.java
```diff
@@ -0,0 +1,163 @@
...
+                    + "](" + jbsIssue.webUrl() + ")" + " has been targeted.");
+            if (!labelNames.contains(JEP_LABEL)) {
+                pr.addLabel(JEP_LABEL);
+            }
+        }
+    }
+
+    @Override
+    public String description() {
+        return "require a JDK enhancement proposal for this pull request";
```
erikj79, line 156:
```suggestion
        return "require a JDK Enhancement Proposal for this pull request";
```

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -151,6 +154,44 @@ private boolean isTargetBranchAllowed() {
                 .filter(link -> link.relationship().isPresent() && "csr for".equals(link.relationship().get())).findAny();
     }
 
+    private Optional<Issue> getJepIssue() {
+        var issueOpt = getJepIssueFromIssueTracker();
```
erikj79, line 158: Do we really need to get the issue from the issue tracker at this point? For other issues, CheckRun just gets the ID and description from the encoded marker pattern. See SolvesTracker.java.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -218,6 +259,25 @@ private IssueProject issueProject() {
...
+                var resolution = issue.properties().get("resolution");
+                String resolutionName = "";
+                if (resolution != null && !resolution.isNull() &&
+                        resolution.get("name") != null && !resolution.get("name").isNull()) {
+                    resolutionName = resolution.get("name").asString();
+                }
+                if ("Targeted".equals(issueStatus) || "Integrated".equals(issueStatus) ||
+                    "Completed".equals(issueStatus) || ("Closed".equals(issueStatus) && "Delivered".equals(resolutionName))) {
+                    ret.put("Change requires a JEP request to be targeted", true);
+                }
```
erikj79, line 278: We shouldn't need to repeat the work of the JEPBot here. We should just need to check if there is a comment matching the jepPattern and it's not saying unneeded.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/JEPCommand.java
```diff
@@ -0,0 +1,163 @@
...
+    static final Pattern jepMarkerPattern = Pattern.compile("<!-- jep: '(.*?)' '(.*?)' '(.*?)' -->");
+    private static final String unneededMarker = "<!-- jep: 'unneeded' 'unneeded' 'unneeded' -->";
+
+    private void showHelp(PrintWriter reply) {
+        reply.println("""
+                Command syntax:
+                 * `/jep <issue-id>`
+                 * `/jep JEP-<jep-id>`
+                 * `/jep jep-<jep-id>`
+                 * `/jep [unneeded|uneeded]`
```
erikj79, line 50: The command should accept the alternate spelling 'uneeded' but we should not list it in the help text.

erikj79 (commented):
I have looked over the design and the product code and left some comments. I have not yet looked at the tests.

bots/jep/src/main/java/org/openjdk/skara/bots/jep/JEPBot.java
```diff
@@ -0,0 +1,140 @@
...
+    private final IssueProject issueProject;
+
+    JEPBot(HostedRepository repo, IssueProject issueProject) {
+        this.repo = repo;
+        this.issueProject = issueProject;
+    }
+
+    @Override
+    public boolean concurrentWith(WorkItem other) {
+        if (!(other instanceof JEPBot otherBot))
```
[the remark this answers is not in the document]
lgxbslgx: Fixed.

bots/pr/src/test/java/org/openjdk/skara/bots/pr/CheckTests.java
lgxbslgx: Fixed.

issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraProject.java
lgxbslgx: Not, I have not.

When I request the link `https://bugs.openjdk.java.net/rest/api/2/search?jql=project = JDK AND customfield_10701 = 11` locally, it reports `{"errorMessages":["Field 'customfield_10701' does not exist or you do not have permission to view it."],"errors":{}}`.

But when I request the link `https://bugs.openjdk.java.net/rest/api/2/search?jql=project = JDK AND type = JEP`, it returns the correct data.

It seems that my account don't have permission to search the field `customfield_10701`. Could I get your help to verify it?

bots/pr/src/main/java/org/openjdk/skara/bots/pr/JEPCommand.java
lgxbslgx: The author can use the `/jep <jep-id>|<issue-id>` again if the author linked the wrong JEP. This check follows the convention of the `csr` command. It can protect the situation that the author uses the command `/jep unneeded` and integrates the patch unexpectly.

Actually, the `/jep unneeded` command is rarely used and the author can request the reviewers to do that if the patch really don't need a JEP to be targeted. Preventing the unexpected (and maybe wrong) code from integrating should be the most important thing we need to ensure even though it has not happened in the past.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
lgxbslgx: Fixed.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
lgxbslgx: Fixed.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/JEPCommand.java
lgxbslgx: Fixed.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/JEPCommand.java
erikj79: I understand that this is how it would work with the current logic, and I don't agree with it. Declaring the need for a JEP and a CSR are different, so they do not need to follow the same model (though I'm not sure I agree with the limitation on `/csr unneeded` either).

In order for a committer to mishandle this, they would have to actively issue `/jep unneeded` and then integrate. I think it's unlikely to happen by mistake, and mistakes is what this feature intends to protect against. We already trust committers to handle `/integrate` on their own, this isn't about protecting against malicious committers.

issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraProject.java
lgxbslgx: I opened a JEP issue in the browser and then pressed `F12` to get the HTML info which contains the field name.

lgxbslgx: The link `https://bugs.openjdk.java.net/rest/api/2/search?jql=project = JDK AND type = JEP` also returns this field. Such as `"customfield_10701":"419"`.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/JEPCommand.java
lgxbslgx: > We already trust committers to handle /integrate on their own, this isn't about protecting against malicious committers.

I agree with you now and I will fix it later.

kevinrushforth: > I understand that this is how it would work with the current logic, and I don't agree with it. Declaring the need for a JEP and a CSR are different, so they do not need to follow the same model (though I'm not sure I agree with the limitation on `/csr unneeded` either).

I feel that the current limitation on `/csr unneeded` is a good one, but I also agree that JEP need not (and probably should not) have this same restriction.

issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraProject.java
lgxbslgx: The previous link `https://bugs.openjdk.java.net/rest/api/2/search?jql=project = JDK AND customfield_10701 = 11` has two errors.

1. We should use the display name to query instead of the internal name. So the `customfield_10701` need to be changed to `JEP Number`.
2. We can't query text field by using the operator `=`. We need to use the operator `~`.

So the link should be `https://bugs.openjdk.java.net/rest/api/2/search?jql=project = JDK AND "JEP Number" ~ "11"`

Related documentation:
[1] https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-issue-search/#api-rest-api-2-search-post
[2] https://support.atlassian.com/jira-software-cloud/docs/advanced-search-reference-jql-operators/

erikj79: I tried to query in the webUI and then convert to advanced query. That gives me:
https://bugs.openjdk.java.net/issues/?jql=project%20%3D%20JDK%20AND%20issuetype%20%3D%20JEP%20AND%20%22JEP%20Number%22%20~%20%2211%22

Trying that with the rest api seems to work too:
https://bugs.openjdk.java.net/rest/api/2/search?jql=project=JDK%20AND%20%22JEP%20Number%22~11

bots/pr/src/main/java/org/openjdk/skara/bots/pr/JEPCommand.java
lgxbslgx: Fixed just now.

issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraProject.java
lgxbslgx: Fixed just now.

lgxbslgx: @erikj79 we tried the same thing at the same time and get the same result. So the result should be reliable.

erikj79: When I have been introducing new API calls to Github/Gitlab, I introduced manual tests so that I could verify and debug the interaction with the remote services. These tests required specialized setup that we can't put in the OpenJDK Skara repo (credentials and URLs) and the tests are of a nature that we don't want to run automatically all the time. But at least having something that you can run when you need to, helps a lot. See forge/src/test/java/org/openjdk/skara/forge/ManualForgeTests.java for an example. I don't have any example for Jira. Since this is just a read query, this test won't need authentication a least.

lgxbslgx: I added a test case about the jira api just now.

issuetracker/src/test/java/org/openjdk/skara/issuetracker/jira/JiraProjectTests.java
```diff
@@ -0,0 +1,58 @@
...
+
+import org.junit.jupiter.api.Test;
+import org.openjdk.skara.network.URIBuilder;
+
+import java.io.IOException;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class JiraProjectTests {
```
erikj79, line 33: To my knowledge, no automatic tests in the Skara project use remote services. Until we decide to change this policy, or find a reasonable way to run such tests, please mark this test as disabled.

```suggestion
/**
 * To be able to run the tests, you need to remove or comment out the @Disabled
 * annotation first.
 */
@Disabled("Manual")
public class JiraProjectTests {
```

lgxbslgx: Fixed.

dholmes-ora:
I think this is totally unwarranted and there was not enough previous discussion at the JDK level to actually conclude anything about the need for this and how it should work. A JEP is not like a CSR-request. It is extremely rare to create an issue and raise a PR and have someone then say "this really needs a JEP".

David

lgxbslgx:
> A JEP is not like a CSR-request. It is extremely rare to create an issue and raise a PR and have someone then say "this really needs a JEP".

@dholmes-ora The current design[4] and this patch have the following advantages or features at least:

**1. Prevent a JEP implemetation PR from accidentally integrating before the JEP has been targeted**

As Jon stated in the SKARA-1096[5]: 
> It would be convenient to mark a PR as dependent on a JEP being targeted. This would help prevent accidental premature integration

This situation has already occured in the past[6-9]. So it is necessary and important to prevent it.

**2. Mark a PR as a JEP implemetation**

The `jep` label can be added to the PR, so the developers can use `is:open is:pr label:jep` in the github to search all the JEPs which are under review. It is nice because we didn't have a right way to search them in the past.

**3. Show the clear JEP Progress in the PR**

This patch adds a checkbox item `Change requires a JEP request to be targeted` in the `Progresses` of the PR body. It shows the clear JEP Progress to the PR author. And the JEP progress can be changed automatically by the bot according to the JEP issue status, which is very convenient.

**4. Provide a JEP link**

This patch provides a JEP issue link in the `Issues` of the PR body. It is also convenient to the developers who want to jump to the JEP issue from the PR.

After this patch integrating, what the PR author need to do is only typing the command `/jep <jep-id>`. Then SKARA can help complete all such good features.

---

> there was not enough previous discussion at the JDK level to actually conclude anything about the need for this and how it should work.

I posted the first design to the skara-dev mailing list at Dec, 2021[1]. After 3-4 months, I only recieved 2 comments, and one is yours[2]. The skara-dev mailing list has at least dozens of  (if not hundreds or thousands of) subscribers, including all the SKARA reviewers and you @dholmes-ora. Are all of you so busy at these months that you can't give some concrete suggessions or can't give the right mailing list to discuss it?

As you said in the email[2], `I'll have to make enquiries.`:

> I don't think that is a decision to be made just in skara-dev. I'm not 
> certain where would be the best place to discuss it ... we don't have a 
> general infrastructure mailing list to discuss JBS. I'll have to make 
> enquiries.

Does you need several months to find a right mailing list? Maybe it is only because all of you are lazy or are not instereting in this issue.

At Mar. 2022, I provided another design[4] and had still waited for your ideas and suggestions. One month passed, no one gave a suggestion or expressed the objection. You have to know that most of the SKARA subscribers are the experienced OpenJDK developers who are both instereted in JDK and SKARA. Such developers can't or don't want to give any suggestion or point the right mailing list. Can the developers from other mailing list can give the actually good suggestions?

If you still think that it need to be discussed. I can post the design detail to the jdk-dev or other mailing lists to re-start the discussion. But I don't think it is necessary personally.

---

> I think this is totally unwarranted

@dholmes-ora Actually, I am so sad when I read this comment. Because it seems that you have not read the issue[5] and the design[4] (or not read carefully/patiently) before opposing to the whole design. Please read the first advantage mentioned above again, this feature is necessary and important so that @jonathan-gibbons reported the issue[5] and so that we take so much time to discuss and complete the design and patch.

---

Anyway, It @dholmes-ora you think we need to continue to discuss the design, I am happy to do that. Because I don't want to integrate a patch that all of you don't like. But can @dholmes-ora you provide some real suggesions at the following discussion and provide the reason when stating your opposition, which can let the design and patch better, at the next time?

[1] https://mail.openjdk.java.net/pipermail/skara-dev/2021-December/005481.html
[2] https://mail.openjdk.java.net/pipermail/skara-dev/2021-December/005482.html
[3] https://mail.openjdk.java.net/pipermail/skara-dev/2021-December/005490.html
[4] https://mail.openjdk.java.net/pipermail/skara-dev/2022-March/005770.html
[5] https://bugs.openjdk.java.net/browse/SKARA-1096
[6] https://github.com/openjdk/jdk/pull/290
[7] https://github.com/openjdk/jdk/pull/291
[8] https://github.com/openjdk/jdk/pull/740
[9] https://github.com/openjdk/jdk/pull/769

erikj79:
> I think this is totally unwarranted and there was not enough previous discussion at the JDK level to actually conclude anything about the need for this and how it should work. A JEP is not like a CSR-request. It is extremely rare to create an issue and raise a PR and have someone then say "this really needs a JEP".

That isn't the intended usecase here. To issue the command, a JEP must already exist, as you have to reference a bugid/jepid pointing to the JEP. This command is mostly intended for the author to run, to make it clearer that the PR is related to a JEP.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/JEPCommand.java
```diff
@@ -0,0 +1,157 @@
...
+        var jepNumber = jbsIssue.properties().get(JiraProject.JEP_NUMBER).asString();
+        reply.println(String.format(jepMarker, jepNumber, jbsIssue.id(), jbsIssue.title()));
+        if ("Targeted".equals(issueStatus) || "Integrated".equals(issueStatus) ||
+            "Completed".equals(issueStatus) || ("Closed".equals(issueStatus) && "Delivered".equals(resolutionName))) {
+            reply.println("the JEP for this pull request, [JEP-" + jepNumber + "](" + jbsIssue.webUrl() + "), has already been targeted.");
+            if (labelNames.contains(JEP_LABEL)) {
+                pr.removeLabel(JEP_LABEL);
+            }
+        } else if ("Draft".equals(issueStatus) || "Submitted".equals(issueStatus) || "Candidate".equals(issueStatus) ||
+                "Proposed to Target".equals(issueStatus) || "Proposed to Drop".equals(issueStatus) || "Closed".equals(issueStatus)) {
```
erikj79, line 139: I think this should either just be "else", or we need an else clause after this for catching any unknown states. I think I prefer the former.

bots/pr/src/test/java/org/openjdk/skara/bots/pr/CheckTests.java
```diff
@@ -1189,6 +1190,65 @@ void issueWithCsr(TestInfo testInfo) throws IOException {
...
+
+            // Populate the projects repository
+            var localRepo = CheckableRepository.init(tempFolder.path(), author.repositoryType(),
+                    Path.of("appendable.txt"), Set.of("issues"), null);
+            var masterHash = localRepo.resolve("master").orElseThrow();
+            localRepo.push(masterHash, author.url(), "master", true);
+
+            var mainIssue = issueProject.createIssue("The main issue", List.of("main"), Map.of("issuetype", JSON.of("Bug")));
+            var jepIssue = issueProject.createIssue("The jep issue", List.of("Jep body"),
+                    Map.of("issuetype", JSON.of("JEP"), "status", JSON.object().put("name", "Submitted"), JiraProject.JEP_NUMBER, JSON.of("123")));
```
erikj79, line 1215: Consider using static import for JEP_NUMBER.

bots/pr/src/test/java/org/openjdk/skara/bots/pr/CheckTests.java
```diff
@@ -1189,6 +1190,65 @@ void issueWithCsr(TestInfo testInfo) throws IOException {
...
+            jepIssue.setProperty("status", JSON.object().put("name", "Submitted"));
+            // Push a commit to trigger the check which can update the PR body.
+            var newHash = CheckableRepository.appendAndCommit(localRepo);
+            localRepo.push(newHash, author.url(), "edit", false);
+
+            // PR should have two issues
+            TestBotRunner.runPeriodicItems(checkBot);
+            assertTrue(pr.body().contains("### Issues"));
+            assertTrue(pr.body().contains("The main issue"));
+            assertTrue(pr.body().contains("The jep issue (**JEP**)"));
```
erikj79, line 1248: This part of the test isn't really doing anything. Should it perhaps remove the JEP label and verify that the JEP requirement is now met?

bots/notify/src/test/java/org/openjdk/skara/bots/notify/issue/IssueNotifierTests.java
```diff
@@ -444,6 +445,66 @@ void testCsrIssue(TestInfo testInfo) throws IOException {
...
+            assertEquals(1, issueComments.size());
+            assertTrue(issueComments.get(0).body().contains(pullRequestTip));
+            assertTrue(issueComments.get(0).body().contains(pr.webUrl().toString()));
+
+            // jep issue shouldn't have the PR link or PR comment.
+            var csrIssueLinks = updatedJepIssue.links();
+            assertEquals(0, csrIssueLinks.size());
+
+            var csrIssueComments = updatedJepIssue.comments();
+            assertEquals(0, csrIssueComments.size());
```
erikj79, line 504:
```suggestion
            var jepIssueLinks = updatedJepIssue.links();
            assertEquals(0, jepIssueLinks.size());

            var jepIssueComments = updatedJepIssue.comments();
            assertEquals(0, jepIssueComments.size());
```

bots/jep/src/main/java/org/openjdk/skara/bots/jep/JEPBot.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    @Override
+    public List<WorkItem> getPeriodicItems() {
+        return List.of(this);
+    }
+
+    @Override
+    public String name() {
+        return JEP_LABEL;
```
erikj79, line 143:
```suggestion
        return CSRBotFactory.NAME;
```

bots/jep/src/main/java/org/openjdk/skara/bots/jep/JEPBot.java
```diff
@@ -0,0 +1,145 @@
...
+                    pr.removeLabel(JEP_LABEL);
+                }
+                log.info("Found `/jep unneeded` command");
+                continue;
+            }
+
+            var issueOpt = issueProject.issue(issueId);
+            if (issueOpt.isEmpty()) {
+                if (pr.labelNames().contains(JEP_LABEL)) {
+                    pr.removeLabel(JEP_LABEL);
```
erikj79, line 89: If we remove the label in this situation, then the PR bot will mark the PR as `[x] Change requires a JEP request to be targeted`, which isn't true. I think the label should stay on. The jep command will not add the label unless it finds the JEP. I can imagine the following situations where this could happen:

* Jira is having a temporary outage - In this case we should do nothing.
* The JEP is no longer visible (e.g. made oracle confidential) - The user/author needs to decide how to handle this, by referencing another JEP or /jep unneeded.
* The JEP comment is corrupted - I don't think we can expect the bot to figure out this on its own.

Since this is an error state that shouldn't be possible to reach from normal operation, we should log SEVERE so that an admin gets notified. The same goes for the case below where the issue is of the wrong type.

bots/jep/src/main/java/org/openjdk/skara/bots/jep/JEPBot.java
```diff
@@ -0,0 +1,145 @@
...
+            var jepComment = pr.comments().stream()
+                    .filter(comment -> comment.author().equals(pr.repository().forge().currentUser()))
+                    .flatMap(comment -> comment.body().lines())
+                    .map(jepMarkerPattern::matcher)
+                    .filter(Matcher::find)
+                    .reduce((first, second) -> second)
+                    .orElse(null);
+            if (jepComment == null) {
+                if (pr.labelNames().contains(JEP_LABEL)) {
+                    pr.removeLabel(JEP_LABEL);
```
erikj79, line 71: There is a potential race here. The /jep command adds the label first and the comment after. Not sure how to solve that. Ideally the label should be added after the comment.

dholmes-ora:
The original proposal for this solicited only two responses from myself and Erik and we both expressed concerns that this was beyond the scope of discussion for skara-dev - codifying these aspects of the JEP process (particularly with the original proposed changes to JBS) needed to be done as part of the JEP process. So that "discussion" went nowhere. I apologise that I did not get back regarding looking into a more suitable place to discuss the JBS changes - there really isn't one, so jdk-dev (as suggested by Erik at the time) seems a reasonable place.

After that initial "discussion" nothing was said for three months at which point you decided to proceed with a reduced proposal. I would have taken the lack of further discussion as sign of either a general lack of interest about the proposal, or a sign that it was not being seen by the right people in the right place. 

So the current simplified proposal addresses the basic need expressed in SKARA-1096, and prevents the accidental integrations that you rightly point out have actually occurred. I withdraw, and apologise for, my comment that this feature is "unwarranted".

I would still like to see the proposal announced on a broader mailing, such as jdk-dev, to increase awareness and potentially garner some feedback.

Thanks,
David

lgxbslgx:
I posted https://mail.openjdk.java.net/pipermail/jdk-dev/2022-April/006498.html to re-start the discussion.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/JEPCommand.java
lgxbslgx: I fixed it by using the former style and added a comment.

bots/pr/src/test/java/org/openjdk/skara/bots/pr/CheckTests.java
lgxbslgx: Fixed.

bots/notify/src/test/java/org/openjdk/skara/bots/notify/issue/IssueNotifierTests.java
lgxbslgx: Fixed.

bots/jep/src/main/java/org/openjdk/skara/bots/jep/JEPBot.java
lgxbslgx: Fixed and used the static import.

bots/jep/src/main/java/org/openjdk/skara/bots/jep/JEPBot.java
lgxbslgx: Fixed.

bots/jep/src/main/java/org/openjdk/skara/bots/jep/JEPBot.java
lgxbslgx: Another case: If there is a exiting jep command comment before. We may get the wrong comment and get the wrong issue in the following code.
The error state will be fixed in the next JEPBot cycle, but it is not accepted because it seems that the error occurs frequently.

Currently, the JEPBot implements the interface `Bot` and `WorkItem`. I would like to seperate the Bot (also named JEPBot) and the WorkItem (may named JEPStateCheckWorkItem). This new WorkItem maps to the pre PR, which is similar to the PullRequestWorkItem. And we need to override the method `WorkItem#concurrentWith` carefully to let the new WorkItem not concurrent with the PullRequestWorkItem if they operate the same PR. This is a common way to solve concurrency issues in the current SKARA "framework".

erikj79: Unfortunately that won't work. The concurrentWith construct only works for WorkItems within the same bot module. An important property of the different bot modules is that they can be run independently in different Java processes. We are for instance currently not running the PullRequestBot and the CSRBot in the same JVM instance.

lgxbslgx: > An important property of the different bot modules is that they can be run independently in different Java processes. We are for instance currently not running the PullRequestBot and the CSRBot in the same JVM instance.

From your description, I think this is a synchronization problem in a distributed system. Conceptually, the correct solution is using the distributed lock, but the current SKARA project don't have such mechanism.

erikj79: If change operations are done in the right order, it wouldn't be a problem. In this case, if the jep command added the label after the comment, it would be fine without synchronization. One can also argue that this issue will self correct anyway. The JEPBot will see the correct comment eventually and restore the label. The downside is that until that happens (which can easily take enough time to be confusing to users), the PR will show that the JEP has been targeted.

erikj79: I pondered on this some more and I think we can fix it. We can extend `CommandHandler::handle` for PRs with another parameter for returning a collection of labels to add to the PR. PullRequestCommandWorkItem calls this with an empty collection and any command that needs to add labels after the reply is posted can add them to this collection.

lgxbslgx: Good idea. Actually, we also need to remove labels sometimes, such as the command `/jep unneeded`. What about returning a pair collections like `<labelsToAdd, labelsToRemove>` ?

erikj79: At least in the jep command case, removing before posting comment is fine. I don't think we should add a dead parameter at this time. This can easily be extended if we ever find a need for it.

erikj79: Hm, at second thought, I take that back. Removing it after adding a new comment is better.

lgxbslgx: I updated the code just now. The new code changes the PR labels after commenting just as we discussed before.

And I refactored the method `processCommand` to avoid the multi-level nested conditional statements.

erikj79: This wasn't quite how I envisioned it. I don't think we need to overload the handle method. I also don't think we should use the return value for this. We already use the reply parameter as a kind of return value, a parameter that you modify to return a new state to the caller. My idea was to just add two more parameters to the existing handle method:
```
..., List<String> labelsToAdd, List<String> labelsToRemove)
```
The logic in `processCommand` wouldn't need to change much. It would need to instantiate two ArrayLists in the case of `isCommit==false`, and then loop through each list to process the label changes on the pr. The `pr.addComment(writer.toString())` line would need to be moved from the end to just after each of the two `handle` calls.

lgxbslgx: I misunderstood your idea previously. It is better to add two more parameters to the existing handle method than overload it.

After adding two more parameters, the signatures of the method `handle` in all the commands need to be changed, too. Is it the expected behaviour? I need to confirm it, because it may change many files.

---

The method `CommandHandler#changeLabelsAfterComment`  can be added if we want to control it explicitly.
If the `changeLabelsAfterComment` is true, it has these three steps:
1. add/remove labels before comment (in the method `handle`)
2. comment (in the method `processCommand`)
3. add/remove labels after comment (in the method `processCommand`)

If the `changeLabelsAfterComment` is false, it has only two steps:
1. add/remove labels before comment (in the method `handle`)
2. comment (in the method `processCommand`)
 
If we choose not to add the method `CommandHandler#changeLabelsAfterComment`, we may do the test `labelsToAdd is not empty && labelsToRemove is not empty` to implicitly confirm whether the command want to change the labels after commenting. This implicit way actually make the same three steps : 
1. add/remove labels before comment (in the method `handle`)
2. comment (in the method `processCommand`)
3. add/remove labels after comment if existing (in the method `processCommand`)

What's your opinion, explicitly or implicitly?

lgxbslgx: Now, I think the implicit way is better and follow your idea above.

erikj79:
> @erikj79 During revising the code, I find two issues:
> 
> 1. the `AllowCommand` and `RejectCommand` are not used in the `PullRequestCommandWorkItem`
> 2. the `AllowCommand`, `RejectCommand` and `OpenCommand` are not documented at the wiki [1]
> 
> Is it the intentional behavious or a bug or unfinished work?
> 
> [1] https://wiki.openjdk.java.net/display/SKARA/Pull+Request+Commands

The OpenCommand should probably be documented on the wiki. The other two I don't know the purpose of. I would say unfinished/unclear/prototype work from pretty early in Skara development. I don't think we need them.

bots/jep/src/main/java/org/openjdk/skara/bots/jep/JEPBot.java
erikj79: I don't think we need `CommandHandler#changeLabelsAfterComment`, it just adds extra complexity.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/JEPCommand.java
```diff
@@ -0,0 +1,157 @@
...
+                 * `/jep JDK-1234567`
+                 * `/jep 1234567`
+                 * `/jep jep-123`
+                 * `/jep JEP-123`
+                 * `/jep unneeded`
+
+                Note:
+                The project prefix of the issue ID (`JDK-` in the above examples) is optional.
+                But the `JEP-` or `jep-` prefix is required when you provide a JEP ID.
+                And the issue type must be `JEP`.
```
magicus, line 62:
```suggestion
                The project prefix (`JDK-` in the above examples) is optional if you use an issue ID. 
                The issue type in that case must be `JEP`.
                The `JEP-` or `jep-` prefix is required if you instead provide a JEP ID.
```

bots/pr/src/main/java/org/openjdk/skara/bots/pr/JEPCommand.java
```diff
@@ -0,0 +1,157 @@
...
+                """);
+    }
+
+    private Optional<Issue> getJepIssue(String args, PullRequestBot bot) {
+        Optional<Issue> jbsIssue;
+        if (args.startsWith("jep-") || args.startsWith("JEP-")) {
```
magicus, line 68:
```suggestion
        if (args.toUpperCase().startsWith("JEP-")) {
```

Then variations like "Jep-123" is also accepted.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/JEPCommand.java
```diff
@@ -0,0 +1,157 @@
...
+            return;
+        }
+
+        var labelNames = pr.labelNames();
+        if ("unneeded".equals(args) || "uneeded".equals(args)) {
+            if (labelNames.contains(JEP_LABEL)) {
+                labelsToRemove.add(JEP_LABEL);
+            }
+            reply.println(unneededMarker);
+            reply.println("determined that the [JEP](https://openjdk.java.net/jeps) request " +
```
magicus, line 98:
```suggestion
            reply.println("determined that the JEP request " +
```

bots/pr/src/main/java/org/openjdk/skara/bots/pr/JEPCommand.java
```diff
@@ -0,0 +1,157 @@
...
+                    + "](" + jbsIssue.webUrl() + ")" + " has been targeted.");
+            if (!labelNames.contains(JEP_LABEL)) {
+                labelsToAdd.add(JEP_LABEL);
+            }
+        }
+    }
+
+    @Override
+    public String description() {
+        return "require a JDK Enhancement Proposal for this pull request";
```
magicus, line 150:
```suggestion
        return "require a JDK Enhancement Proposal (JEP) for this pull request";
```

bots/jep/src/main/java/org/openjdk/skara/bots/jep/JEPBot.java
erikj79: Yes, I meant to change the signature. Updating it in all the command implementations is fine.

erikj79:
This is starting to look good now. My only remaining concern is that the total execution time of the Skara tests goes up quite a bit with all these new and rather expensive tests. Not sure we can do much about this, perhaps something to investigate separately.

magicus (approved):
FWIW, I've looked through all the code in the PR now. I added a few suggestions on user fronting strings, otherwise I have not much to add. I'll mark this as approved, but you do need to get Erik's approval as well. (And please fix the strings :))

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CommandHandler.java
```diff
@@ -34,8 +33,8 @@
 interface CommandHandler {
     String description();
 
-    default void handle(PullRequestBot bot, PullRequest pr, CensusInstance censusInstance, Path scratchPath, CommandInvocation command, List<Comment> allComments, PrintWriter reply)
-    {
+    default void handle(PullRequestBot bot, PullRequest pr, CensusInstance censusInstance, Path scratchPath, CommandInvocation command,
+                        List<Comment> allComments, PrintWriter reply, List<String> labelsToAdd, List<String> labelsToRemove) {
     }
```
magicus, line 38: I keep coming back to this in my reviewing. It would be better if we did not have to update all those other places where handle() is implemented. 

I might be thinking wrong here, but I believe that you should be able to keep the original `handle(PullRequestBot bot, PullRequest pr, CensusInstance censusInstance, Path scratchPath, CommandInvocation command, List<Comment> allComments, PrintWriter reply)` method, and instead add the new `handle` (with `List<String> labelsToAdd, List<String> labelsToRemove` as a default method, with the implementatation
```
handle(bot, pr, censusInstance, scratchPath, command, allComments, reply);
```

This way, the commands that need the new label handling mechanism can implement the new handle override with the extended argument list. All other commands needs no change.

Once again, I don't feel 100% confident I'm thinking correctly here, but it seems to me like it would work, and be a better solution.

erikj79: You are correct Magnus, that would be nicer indeed.

erikj79:
> This is starting to look good now. My only remaining concern is that the total execution time of the Skara tests goes up quite a bit with all these new and rather expensive tests. Not sure we can do much about this, perhaps something to investigate separately.

I take that back, when I ran all the tests just now, the total time was close enough to what it usually is. Last time I ran it took considerably longer. Might have just been a fluke.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CommandHandler.java
erikj79: The drawback of doing that is that JEPCommand would still need to implement the old handle method, which doesn't make sense for it to do.

lgxbslgx: > This way, the commands that need the new label handling mechanism can implement the new handle override with the extended argument list. All other commands needs no change.

If we obey this convention which adds a new "bridge" `handle` method. Next time, we want to add another parameter and add a new `handle` method, too. The interface `CommandHandler` will have sevaral such "bridge" `handle` methods in the future which are actually unnecessary. @erikj79, @magicus and I may feel good with it because we added or reviewed it. But in the future, the new SKARA developers may be confused by it. In other words, the maintainment of the project may become a little more complex.

>  but I'm also ok leaving it as it is.

I would like to keep the code as it is now.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/JEPCommand.java
lgxbslgx: Fixed.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/JEPCommand.java
lgxbslgx: Fixed and added a test case.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/JEPCommand.java
lgxbslgx: Fixed and joined with next line and revised the corresponding tests.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/JEPCommand.java
lgxbslgx: Fixed.

lgxbslgx:
@magicus Thanks for the review. I updated the code just now.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CommandHandler.java
magicus: No, each and every implementation of `CommandHandler` would only implement a single `handle()` method.

At least, if I'm thinking correctly about this. :-)

When processing WorkItems, in `processCommand`, we call the "long" version, with the two extra arguments. This will have a default implementation in the `CommandHandler` interface, which calls the "short" version, dropping the two last arguments. The "short" version will have a default implementation, which does nothing. (This is exactly the case of the code right now, without this PR.) 

So, my suggestion is that we leave the old "short" method untouched, add a new "long" method which by default delegates to the "short" method, and call the new "long" method in `processCommand`.

Then all old implementations can keep their "short" implementations, and everything will work as before.

JEPCommand will need to override the new "long" version instead (and can just ignore the "short" version), and everything will work perfect for it too.

And if we in the future need to add further arguments, we can do the same procedure again -- add a new long-long method, which deletages to the long method, etc.

I believe this was the exact use case that the `default` method implementation in interfaces was intended for.

magicus: And I believe this solution is so much better that I'd strongly argue that we should not keep the code as it looks right now. There is no need nor good reason to push these additional argument to command implementations.

lgxbslgx:
> The OpenCommand should probably be documented on the wiki.

@erikj79 Could you help add the OpenCommand to the wiki? Or we can file an issue to record it so that it can be added in the future by others?

> The other two I don't know the purpose of. I would say unfinished/unclear/prototype work from pretty early in Skara development. I don't think we need them.

It is good to file a new issue to remove it instead of leaving it which may confuse other developers.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CommandHandler.java
lgxbslgx: > @erikj79, @magicus and I may feel good with it because we added or reviewed it. But in the future, the new SKARA developers may be confused by it. In other words, the maintainment of the project may become a little more complex.

@magicus previously, I thought it is not good to add the "bridge" "long" methods because it may confused other developers.

> There is no need nor good reason to push these additional argument to command implementations.

But now, I think that the additional parameter `labelsToAdd` and `labelsToRemove` which have not used in much cammands confused the developers, too and are more terrible. So I agree with you and updated the code just now.

lgxbslgx:
> It is good to file a new issue to remove it instead of leaving it which may confuse other developers.

I filed https://bugs.openjdk.java.net/browse/SKARA-1402 to follow up.

bots/jep/src/main/java/org/openjdk/skara/bots/jep/JEPBot.java
```diff
@@ -0,0 +1,141 @@
...
+                }
+                log.info("No jep command found in comment for PR-" + pr.id());
```
erikj79, line 75: This is going to log for most PRs every time this WorkItem runs. I would like to lower it to `fine` level. Also, please see CSR bot for a good way of logging pr name/IDs. I would recommend copying the describe method here.

bots/jep/src/main/java/org/openjdk/skara/bots/jep/JEPBot.java
```diff
@@ -0,0 +1,141 @@
...
+            }
+
+            var hasTargeted = "Targeted".equals(issueStatus) ||
+                    "Integrated".equals(issueStatus) ||
+                    "Completed".equals(issueStatus) ||
+                    ("Closed".equals(issueStatus) && "Delivered".equals(resolutionName));
+            if (hasTargeted && pr.labelNames().contains(JEP_LABEL)) {
+                pr.removeLabel(JEP_LABEL);
+            } else if (!hasTargeted && !pr.labelNames().contains(JEP_LABEL)) {
+                pr.addLabel(JEP_LABEL);
```
erikj79, line 116: Here we are taking action, I would like this logged at info level. Something like this:
```suggestion
                log.info("JEP issue " + issue.id() + " found in state " + issueStatus + ", removing JEP label from " + describe(pr);
                pr.removeLabel(JEP_LABEL);
            } else if (!hasTargeted && !pr.labelNames().contains(JEP_LABEL)) {
                log.info("JEP issue " + issue.id() + " found in state " + issueStatus + ", adding JEP label to " + describe(pr);
                pr.addLabel(JEP_LABEL);
```

bots/jep/src/main/java/org/openjdk/skara/bots/jep/JEPBot.java
```diff
@@ -0,0 +1,141 @@
...
+                    pr.removeLabel(JEP_LABEL);
+                }
+                log.info("No jep command found in comment for PR-" + pr.id());
+                continue;
+            }
+
+            var issueId = jepComment.group(2);
+            if ("unneeded".equals(issueId)) {
+                if (pr.labelNames().contains(JEP_LABEL)) {
+                    pr.removeLabel(JEP_LABEL);
```
erikj79, line 82: Please log that action was taken, see example below.

bots/jep/src/main/java/org/openjdk/skara/bots/jep/JEPBot.java
```diff
@@ -0,0 +1,141 @@
...
+                }
+                log.info("Found `/jep unneeded` command");
```
erikj79, line 84: Please include the `describe(pr)` in the log message.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
```diff
@@ -146,6 +147,23 @@ private Optional<Hash> resultingCommitHash(List<Comment> allComments) {
                  .findAny();
     }
 
+    private void changeLabelsAfterComment(List<String> labelsToAdd, List<String> labelsToRemove){
+        if (labelsToAdd != null && !labelsToAdd.isEmpty()) {
+            for (var label : labelsToAdd) {
+                if (!pr.labelNames().contains(label)) {
+                    pr.addLabel(label);
```
erikj79, line 154: Could you add an info log about adding and removing labels to the PR.

erikj79 (commented):
Looks pretty good now. I focused on logging this time as that's something I will need to trace from this in the future. 

Thanks for fixing the API change with a default method, it turned out a lot better. Thanks Magnus for pushing through. I had missed that the existing method already had a default and was considered optional to implement.

I have added the /open command to the wiki.

bots/jep/src/main/java/org/openjdk/skara/bots/jep/JEPBot.java
lgxbslgx: Fixed.

bots/jep/src/main/java/org/openjdk/skara/bots/jep/JEPBot.java
lgxbslgx: Fixed.

bots/jep/src/main/java/org/openjdk/skara/bots/jep/JEPBot.java
lgxbslgx: Fixed.

bots/jep/src/main/java/org/openjdk/skara/bots/jep/JEPBot.java
lgxbslgx: Fixed.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
lgxbslgx: Fixed.

issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraProject.java
```diff
@@ -36,6 +36,7 @@ public class JiraProject implements IssueProject {
 
     public static final String RESOLVED_IN_BUILD = "customfield_10006";
     public static final String SUBCOMPONENT = "customfield_10008";
+    public static final String JEP_NUMBER = "customfield_10701";
```
magicus, line 39: Don't think this is needed anymore?

magicus (commented):
Thanks for fixing `handle`! Looks much better now.

issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraProject.java
lgxbslgx: Why? It is used in many test cases and the convention is putting such field name at the `JiraProjec.java`.

magicus: Aha, I did not notice that. I only looked at that particular class. Nevermind, then.

magicus (approved):
Seems the bots did not like that I added a comment using the review function; they took it as me withdrawing my review...

bots/jep/src/main/java/org/openjdk/skara/bots/jep/JEPBot.java
```diff
@@ -0,0 +1,149 @@
...
+            var jepComment = pr.comments().stream()
+                    .filter(comment -> comment.author().equals(pr.repository().forge().currentUser()))
+                    .flatMap(comment -> comment.body().lines())
+                    .map(jepMarkerPattern::matcher)
+                    .filter(Matcher::find)
+                    .reduce((first, second) -> second)
+                    .orElse(null);
+            if (jepComment == null) {
+                if (pr.labelNames().contains(JEP_LABEL)) {
+                    pr.removeLabel(JEP_LABEL);
```
erikj79, line 74: Missed this remove label in the previous review due to the long thread in the middle of the block. Please add logging here too.

lgxbslgx: I missed it, too. Added. Use the `info` level instead of `fine` because the condition only occurs when the developers add the `jep` label manually in the PR.

magicus (approved):

erikj79 (approved):

lgxbslgx:
@magicus @erikj79 Thanks for the review. Could I get your help to sponsor this patch?

/integrate

PR #1305, opened 2022-04-24
282: Make the review requirements more explicit in the progress list

lgxbslgx:
Hi all,

This patch adds the explicit review requirements, such as ` (2 reviews required, with at least 1 Reviewer).`, to the progress list and adds some test cases.

Best Regards,
-- Guoxiong

---------
### Progress

### Issue
 * [SKARA-282](https://bugs.openjdk.java.net/browse/SKARA-282): Make the review requirements more explicit in the progress list

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1305/head:pull/1305` \
`$ git checkout pull/1305`

`$ git checkout pull/1305` \
`$ git pull https://git.openjdk.java.net/skara pull/1305/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1305`

`$ git pr show -t 1305`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1305.diff">https://git.openjdk.java.net/skara/pull/1305.diff</a>

</details>

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckablePullRequest.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckablePullRequest.java
@@ -45,16 +45,11 @@ public class CheckablePullRequest {
     private final boolean ignoreStaleReviews;
     private final List<String> confOverride;
 
-    // The review requirements is used by CheckRun#getChecksList to
-    // set the checkbox of the `Progress` in the PR body.
-    private String reviewRequirements;
-
     CheckablePullRequest(PullRequest pr, Repository localRepo, boolean ignoreStaleReviews,
             HostedRepository jcheckRepo, String jcheckName, String jcheckRef) {
         this.pr = pr;
         this.localRepo = localRepo;
         this.ignoreStaleReviews = ignoreStaleReviews;
-        reviewRequirements = "";
 
         if (jcheckRepo != null) {
             confOverride = jcheckRepo.fileContents(jcheckName, jcheckRef).lines().collect(Collectors.toList());
@@ -187,34 +182,6 @@ public class CheckablePullRequest {
         return new PullRequestCheckIssueVisitor(checks);
     }
 
-    String getReviewRequirements() {
-        return reviewRequirements;
-    }
-
-    private void setReviewRequirements(JCheckConfiguration conf) {
-        var reviewersConf = conf.checks().reviewers();
-        var requireList = new ArrayList<String>();
-        var sum = 0;
-        var reviewRequirementMap = new LinkedHashMap<String, Integer>();
-        reviewRequirementMap.put("lead", reviewersConf.lead());
-        reviewRequirementMap.put("reviewer", reviewersConf.reviewers());
-        reviewRequirementMap.put("committer", reviewersConf.committers());
-        reviewRequirementMap.put("author", reviewersConf.authors());
-        reviewRequirementMap.put("contributor", reviewersConf.contributors());
-        for (var reviewRequirement : reviewRequirementMap.entrySet()) {
-            var requirementNum = reviewRequirement.getValue();
-            if (requirementNum > 0) {
-                sum += requirementNum;
-                requireList.add(requirementNum+ " " + reviewRequirement.getKey() + (requirementNum > 1 ? "s" : ""));
-            }
-        }
-        if (sum == 0) {
-            reviewRequirements = " (no reviews required)";
-        } else {
-            reviewRequirements = String.format(" (%d reviews required, with at least %s)", sum, String.join(", ", requireList));
-        }
-    }
-
     void executeChecks(Hash localHash, CensusInstance censusInstance, PullRequestCheckIssueVisitor visitor, List<String> additionalConfiguration) throws IOException {
         Optional<JCheckConfiguration> conf;
         if (confOverride != null) {
@@ -225,8 +192,7 @@ public class CheckablePullRequest {
         if (conf.isEmpty()) {
             throw new RuntimeException("Failed to parse jcheck configuration at: " + targetHash() + " with extra: " + additionalConfiguration);
         }
-        // Set the review requirements for using in CheckRun#getChecksList
-        setReviewRequirements(conf.get());
+        visitor.setConfiguration(conf.get());
         try (var issues = JCheck.check(localRepo, censusInstance.census(), CommitMessageParsers.v1, localHash,
                                        conf.get())) {
             for (var issue : issues) {
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckablePullRequest.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckablePullRequest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2022, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -45,11 +45,16 @@ public class CheckablePullRequest {
     private final boolean ignoreStaleReviews;
     private final List<String> confOverride;
 
+    // The review requirements is used by CheckRun#getChecksList to
+    // set the checkbox of the `Progress` in the PR body.
+    private String reviewRequirements;
+
     CheckablePullRequest(PullRequest pr, Repository localRepo, boolean ignoreStaleReviews,
             HostedRepository jcheckRepo, String jcheckName, String jcheckRef) {
         this.pr = pr;
         this.localRepo = localRepo;
         this.ignoreStaleReviews = ignoreStaleReviews;
+        reviewRequirements = "";
 
         if (jcheckRepo != null) {
             confOverride = jcheckRepo.fileContents(jcheckName, jcheckRef).lines().collect(Collectors.toList());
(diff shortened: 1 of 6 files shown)


dholmes-ora:
These are informal rules not codified anywhere so how are you determining what criteria to report? Hotspot  generally requires two reviews, but that is not a blanket rule.

lgxbslgx:
> These are informal rules not codified anywhere so how are you determining what criteria to report? Hotspot generally requires two reviews, but that is not a blanket rule.

The concrete reviews information depends on the configuration in `.jcheck/conf` and the `reviewers` command.
For example, please see this [PR](https://github.com/openjdk/jdk/pull/7204#issuecomment-1032111596), when someone typed `/reviewers 2`, the bot replied `The number of required reviews for this PR is now set to 2 (with at least 1 of role reviewers)`. This patch puts similar information to the `Progress` of the PR body so that developers can read it earily instead of searching it in the comments.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckablePullRequest.java
```diff
@@ -192,6 +225,8 @@ void executeChecks(Hash localHash, CensusInstance censusInstance, PullRequestChe
         if (conf.isEmpty()) {
             throw new RuntimeException("Failed to parse jcheck configuration at: " + targetHash() + " with extra: " + additionalConfiguration);
         }
+        // Set the review requirements for using in CheckRun#getChecksList
+        setReviewRequirements(conf.get());
```
lgxbslgx, line 229: I don't think it is good to put such code in `CheckablePullRequest#executeChecks`. But I can't find a better way to do that because we need the complete `JCheckConfiguration` config information.

An alternative way is duplicating the following code in the `CheckRun` so that we can get the complete config information.

```
        Optional<JCheckConfiguration> conf;
        if (confOverride != null) {
            conf = JCheck.parseConfiguration(confOverride, additionalConfiguration);
        } else {
            conf = JCheck.parseConfiguration(localRepo, targetHash(), additionalConfiguration);
        }
        if (conf.isEmpty()) {
            throw new RuntimeException("Failed to parse jcheck configuration at: " + targetHash() + " with extra: " + additionalConfiguration);
        }
```

dholmes-ora:
> > These are informal rules not codified anywhere so how are you determining what criteria to report? Hotspot generally requires two reviews, but that is not a blanket rule.
> 
> The concrete reviews information depends on the configuration in `.jcheck/conf` and the `reviewers` command. For example, please see this [PR](https://github.com/openjdk/jdk/pull/7204#issuecomment-1032111596), when someone typed `/reviewers 2`, the bot replied `The number of required reviews for this PR is now set to 2 (with at least 1 of role reviewers)`. This patch puts similar information to the `Progress` of the PR body so that developers can read it earily instead of searching it in the comments.

I see - that makes sense. I'm not sure if that is what SKARA-282 was actually referring to though.

erikj79:
I don't think this is the right solution. After having looked through the various classes involved here, I think I would put the logic for generating a string describing the reviewer requirements in the ReviewersConfiguration class. Then I would keep PullRequestCheckIssueVisitor responsible for extracting the proper message for each kind of issue returned from the get*Checks methods, instead of modifying them in CheckRun. To do this the TooFewReviewersIssue would need to hold the requirements string from the configuration, but I think it makes enough sense for it to do that. ReviewersCheck has access to it and can provide it to the constructor.

lgxbslgx:
> To do this the TooFewReviewersIssue would need to hold the requirements string from the configuration, but I think it makes enough sense for it to do that. ReviewersCheck has access to it and can provide it to the constructor.

I tried this way at first when I solved this issue. But then I found the ReviewersCheck doesn’t always return the TooFewReviewersIssue which means the message may miss sometimes. And if we return the message in all the issues of the ReviewersCheck, the message may also miss when the ReviewersCheck passes without any issue.

erikj79:
> I tried this way at first when I solved this issue. But then I found the ReviewersCheck doesn’t always return the TooFewReviewersIssue which means the message may miss sometimes. And if we return the message in all the issues of the ReviewersCheck, the message may also miss when the ReviewersCheck passes without any issue.

Right, this would only work when the checkbox isn't checked.

My biggest issue here is that you are storing mutable state in `CheckablePullRequest`. That class is currently immutable.

Here is what I would suggest instead. Still move the logic for generating the reviewers description string to ReviewersConfiguration. Add a field to `PullRequestCheckIssueVisitor` for holding a Jcheck configuration. Set this field from `CheckablePullRequest::executeChecks`. The visitor is already a class that carries a mutable state, so this is kind of ok. Then in `PullRequestCheckIssueVisitor::get*Checks` explicitly modify the description for ReviewersCheck in the returned map. I would recommend a new private method that wraps the `Check::description` call and extracts the reviewer information from the conf field.

lgxbslgx:
The new way makes sense. I update the code just now.

bots/pr/src/test/java/org/openjdk/skara/bots/pr/ReviewersTests.java
```diff
@@ -42,6 +42,9 @@ public class ReviewersTests {
     private static final String reviewersCommandFinallyOutput = "The total number of required reviews for this PR " +
             "(including the jcheck configuration and the last /reviewers command) is now set to ";
 
+    private static final String reviewProgressTemplate = "Change must be properly reviewed (%d reviews required, with at least %s)";
+    private static final String zeroReviewProgress = "Change must be properly reviewed (no reviews required)";
```
erikj79, line 46: Please use uppercase underscore separated identifiers for constants.

```suggestion
    private static final String REVIEW_PROGRESS_TEMPLATE = "Change must be properly reviewed (%d reviews required, with at least %s)";
    private static final String ZERO_REVIEW_PROGRESS = "Change must be properly reviewed (no reviews required)";
```

jcheck/src/main/java/org/openjdk/skara/jcheck/ReviewersConfiguration.java
```diff
@@ -76,6 +99,10 @@ public boolean shouldCheckBackports() {
         return shouldCheckBackports;
     }
 
+    public String getReviewRequirements() {
+        return reviewRequirements;
```
erikj79, line 103: I would probably have put the logic for generating this string right here in the getter. Are we ever calling this more than once? Could also generate it lazily.

I realize this gets a lot of testing through the ReviewersTest above, but it would make sense to have cheap and quick tests of just this functionality here in the jcheck package.

lgxbslgx: > I would probably have put the logic for generating this string right here in the getter. 

Fixed. Please note the field `reviewRequirements` is not `final` now.

> I realize this gets a lot of testing through the ReviewersTest above, but it would make sense to have cheap and quick tests of just this functionality here in the jcheck package.

Added tests.

erikj79: That's fine, thanks!

I now realize that you put the space and parentheses in the string here. I think this method should just return the requirements and let the visitor add the surrounding formatting.

lgxbslgx: Fixed.

erikj79 (approved):

lgxbslgx:
@erikj79 Thanks for the review.

/integrate

jcheck/src/main/java/org/openjdk/skara/jcheck/ReviewersConfiguration.java
magicus: @lgxbslgx I missed reviewing this when it was open, and am just going through the backlog of changes that has happened now.

However, I think this was incorrect. The code here makes the text appears as: 
```
Change must be properly reviewed (1 review required, with at least 1 reviewer)
```

when it should have been:
```
Change must be properly reviewed (1 review required, with at least 1 Reviewer)
```

(Note the capitalization of Reviewer).

Traditionally, the (not very clear, I admit) distinction has been made in OpenJDK between "reviewer" (anyone who offers a review) and a "Reviewer" (a member with the official status as Reviewer for a project). This text is about the latter, so changing the "r" to a "R" in the message made a significant difference.

Can you please fix this in a follow-up PR?

lgxbslgx: Do the `contributor`, `author`, `committer`, `lead` also need to be changed to `Contributor`, `Author`, `Committer`, `Lead`?

lgxbslgx: Filed https://bugs.openjdk.java.net/browse/SKARA-1437 to follow up.

magicus: Yes, they should. See https://openjdk.java.net/bylaws#_7 for official roles.

However, to specify any requirement other than Reviewer seems to me to be mostly an academic exercise. In practice, no-one will ever set any limitation other than the project role as Reviewer. (I'm not even sure why that code is there in the first place. Just some completionist thinking, I believe.)

kevinrushforth: Actually, it's quite common to have as a requirement 2 reviewers, at least one of which must be a Reviewer and the other an Author (i.e., any role in the Project). Both `jdk` and `jfx` are configured this way. Here is the current Progress item that Skara adds in response to the command `/reviewers 2`:

```
- [ ] Change must be properly reviewed (2 reviews required, with at least 1 reviewer, 1 author)
```

And here is the comment:

```
The total number of required reviews for this PR (including the jcheck configuration and the last /reviewers command) is now set to 2 (with 1 of role reviewers, 1 of role authors).
```

Both "Reviewer" and "Author" should be capitalized in both messages. Separately, I see that the pluralization was fixed in the Progress item, but not the comment.

erikj79: There are also projects that do not have any members in the Reviewer role, so they configure their repository to only require Committers. The only role I doubt we will ever see required in reality is Lead, but that's ok.

lgxbslgx: What about changing the message `(with 1 of role reviewers, 1 of role authors)` to `(with at least 1 Reviewer, 1 Author)` and obeying the same pluralization rule. The message `(with 1 of role reviewers, 1 of role authors)` may be fixed as `(with 1 of role Reviewer, 1 of role Author)`. But if there is a message `(with 2 of role reviewers)`, we don't know whether it should fixed as `(with 2 of role Reviewer)` or `(with 2 of role Reviewers)` (pluralization).

lgxbslgx: I submit a PR https://github.com/openjdk/skara/pull/1319. We can continue the discussion there.

PR #1304, opened 2022-04-23
1407: GitLabMergeRequest#filesUrl returns wrong result

lgxbslgx:
Hi all,

The SKARA-744 [1][2] solved a problem about the link to the changes for just one commit. But it didn't solve the issue completely in GitLab. Please see the current issue [3] for more information.

This patch revises `GitLabMergeRequest#filesUrl` to return a `version url` if it exists which can represent the changed files **from the first commit to the provided commit HASH**. A corresponding test is added and a previous test is polished.

You can use the following config to run the new test `GitLabRestApiTest#testFilesUrl`:

```
gitlab.user=<your username in https://gitlab.com>
gitlab.pat=<your token in https://gitlab.com>

gitlab.uri=https://gitlab.com
gitlab.repository=35596381
gitlab.merge.request.id=1

gitlab.version.hash=60f39eb698e73cc2660f870667a54e7a53ee7018
gitlab.version.url=https://gitlab.com/lgxbslgx/test/-/merge_requests/1/diffs?diff_id=379601911
gitlab.nonversion.hash=c9b5a70790b13c3e7e182c9b83015f9f10d53df7
gitlab.nonversion.url=https://gitlab.com/lgxbslgx/test/-/merge_requests/1/diffs?commit_id=c9b5a70790b13c3e7e182c9b83015f9f10d53df7
```
Thanks for taking the time to review.

Best Regards,
-- Guoxiong

[1] https://bugs.openjdk.java.net/browse/SKARA-744
[2] https://git.openjdk.java.net/skara/commit/f16663b7
[3] https://bugs.openjdk.java.net/browse/SKARA-1407

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Change must be properly reviewed

### Issue
 * [SKARA-1407](https://bugs.openjdk.java.net/browse/SKARA-1407): GitLabMergeRequest#filesUrl returns wrong result

### Reviewers
 * [Erik Joelsson](https://openjdk.java.net/census#erikj) (@erikj79 - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1304/head:pull/1304` \
`$ git checkout pull/1304`

Update a local copy of the PR: \
`$ git checkout pull/1304` \
`$ git pull https://git.openjdk.java.net/skara pull/1304/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
[...]

PR #1306, opened 2022-04-25
1408: GitLabMergeRequest#reviews sometimes finds the wrong commit HASH

lgxbslgx:
Hi all,

Please consider the following steps:

- the author creates "commit 1" locally (with hash "95f2d613cf392275075d7c87285845531f5fe827" [1])
- the author pushes it and creates a merge request in the Gitlab, then the Gitlab creates "version 1" [2] (its hash is the hash of "commit 1") which has the changed files of "commit 1"
- the author creates "commit 2" locally (with hash "99ca51edfa0063113a2236fce548ba453aee7275" [3])
- a reviewer approves the MR with the "**version 1**" in the Gitlab. Note: the "commit 2" is not pushed now.
- the author creates "commit 3" locally (with hash "50b119c4c325053b8151ea6e761a60d1acfbb746" [4]) Note: this step seems optional
- the author pushes the commits and the gitlab create "version 2" [5] (its hash is the hash of "commit 3") which has the changed files of "commit 1-3"

When we use the method `GitLabMergeRequest#reviews` to get the reviews list, we can see the review hash is the hash of the "commit 2" ("99ca51edfa0063113a2236fce548ba453aee7275"). But actually, the reviewer only approved "version 1" with hash of the "commit 1" ("95f2d613cf392275075d7c87285845531f5fe827").

This patch uses the `versions` list instead of all the `commits` list to fix the issue. And when I wrote a test case to verify it, I got a NPE and filed SKARA-1409 to record it. This patch also solves SKARA-1409 so that the reviewer can run the test locally.

You can use the following config to run the test:
```
gitlab.user=<your username in https://gitlab.com>
gitlab.pat=<your token in https://gitlab.com>

gitlab.uri=https://gitlab.com
gitlab.repository=35596381
gitlab.merge.request.id=2
gitlab.review.hash=95f2d613cf392275075d7c87285845531f5fe827
```

Note: this PR seems have git conflict with [SKARA-1407-PATCH](https://github.com/openjdk/skara/pull/1304). I will solve the conflict when one of them is integrated.

Best Regards,
-- Guoxiong

[1] https://gitlab.com/lgxbslgx/test/-/merge_requests/2/diffs?commit_id=95f2d613cf392275075d7c87285845531f5fe827
[2] https://gitlab.com/lgxbslgx/test/-/merge_requests/2/diffs?diff_id=379883693
[3] https://gitlab.com/lgxbslgx/test/-/merge_requests/2/diffs?commit_id=99ca51edfa0063113a2236fce548ba453aee7275
[4] https://gitlab.com/lgxbslgx/test/-/merge_requests/2/diffs?commit_id=50b119c4c325053b8151ea6e761a60d1acfbb746
[5] https://gitlab.com/lgxbslgx/test/-/merge_requests/2/diffs?diff_id=379907777

---

/issue SKARA-1409

---------
### Progress

### Issues
 * [SKARA-1408](https://bugs.openjdk.java.net/browse/SKARA-1408): GitLabMergeRequest#reviews sometimes finds the wrong commit HASH
 * [SKARA-1409](https://bugs.openjdk.java.net/browse/SKARA-1409): The GitLab 'list-users' api doesn't return 'email' field for normal users

### Reviewers
 * [Erik Joelsson](https://openjdk.java.net/census#erikj) (@erikj79 - **Reviewer**) ⚠️ Review applies to 9217ff1e6922ccf05ce64e7c0b84539ac169bc6a

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1306/head:pull/1306` \
`$ git checkout pull/1306`

`$ git checkout pull/1306` \
`$ git pull https://git.openjdk.java.net/skara pull/1306/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1306`

`$ git pr show -t 1306`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1306.diff">https://git.openjdk.java.net/skara/pull/1306.diff</a>

</details>

### DIFF
--- a/forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabHost.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabHost.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2022, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -150,7 +150,7 @@ public class GitLabHost implements Forge {
         var id = o.get("id").asInt();
         var username = o.get("username").asString();
         var name = o.get("name").asString();
-        var email = o.get("email").asString();
+        var email = o.get("email") != null ? o.get("email").asString() : "";
         return HostUser.builder()
                        .id(id)
                        .username(username)
--- a/forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabMergeRequest.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabMergeRequest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2022, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -87,11 +87,11 @@ public class GitLabMergeRequest implements PullRequest {
             private ZonedDateTime date;
         }
 
-        var commits = request.get("commits").execute().stream()
+        var commits = request.get("versions").execute().stream()
                              .map(JSONValue::asObject)
                              .map(obj -> {
                                  var ret = new CommitDate();
-                                 ret.hash = new Hash(obj.get("id").asString());
+                                 ret.hash = new Hash(obj.get("head_commit_sha").asString());
                                  ret.date = ZonedDateTime.parse(obj.get("created_at").asString());
                                  return ret;
                              })
--- a/forge/src/test/java/org/openjdk/skara/forge/gitlab/GitLabRestApiTest.java
+++ b/forge/src/test/java/org/openjdk/skara/forge/gitlab/GitLabRestApiTest.java
@@ -37,7 +37,7 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 /**
  * To be able to run the tests, you need to remove or comment out the @Disabled annotation first.
  */
-//@Disabled("Manual")
+@Disabled("Manual")
 public class GitLabRestApiTest {
 
     @Test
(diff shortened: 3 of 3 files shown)


erikj79 (approved):

lgxbslgx:
@erikj79 Thanks for the review. Could I get your help to sponsor this patch?

/integrate

lgxbslgx:
Please don't sponsor this patch now. I found a strange point at SKARA-1400 which seems be related to this patch. I will provide the concrete info in next comment.

[1] https://bugs.openjdk.java.net/browse/SKARA-1400
[2] https://github.com/openjdk/skara/pull/1300

lgxbslgx:
From the description in issue SKARA-1400 [1]:

> This can happen if multiple commits have the same "created_at" date (the date the commit was pushed to gitlab).

It seems that the `created_at` date is not the date the commit was pushed to gitlab.
Please see the 3 commits of the PR body [2], the "commit 2" and "commit 3" were pushed at the same time but they have different `create_at` date which are their commit date instead of push date.

So I think SKARA-1400 may be a wrong analysis and different commit should have different `create_at` date (which means we don't need to handle the situation of the same `create_at` date).

What's your opinion? Did I miss anything? If you agree with me, I would like to revert SKARA-1400 in this patch.

[1] https://bugs.openjdk.java.net/browse/SKARA-1400
[2] https://gitlab.com/api/v4/projects/35596381/merge_requests/2/commits

erikj79:
> So I think SKARA-1400 may be a wrong analysis and different commit should have different `create_at` date (which means we don't need to handle the situation of the same `create_at` date).

I filed SKARA-1400 because I had a merge request in our internal Gitlab instance where the two commits were committed at different times, but pushed at the same time, and the `created_at` time was exactly the same. This caused the sorting to go the wrong way, just as I described in that issue.

It's possible that by working with "versions" instead of "commits", the issue in SKARA-1400 won't happen, because commits pushed at the same time will automatically be part of the same version. I would rather keep it there though, it won't cause any harm.

lgxbslgx:
> the two commits were committed at different times, but pushed at the same time, and the created_at time was exactly the same

This description is the same as the issue SKARA-1400, and it confuses me because the commits shouldn’t have the same create_at time if they were committed at different times.

> I would rather keep it there though, it won't cause any harm.

Ok, got it.

kevinrushforth:
>  I had a merge request in our internal Gitlab instance where the two commits were committed at different times, but pushed at the same time

The commits in question were in fact committed at the same time, as the result of a rebase prior to pushing to my local branch. So the CommitDate was the same (down to the second anyway, which is all I can see from `git log`). They were also pushed at the same time. I don't know whether commit time being the same or the fact that they were pushed at the same time was the problem.

lgxbslgx:
> The commits in question were in fact committed at the same time

This makes sense.

lgxbslgx:
I solved the git conflict just now.

/integrate

PR #1312, opened 2022-05-03
1424: CheckRun fails to evaluate jol#24

erikj79:
CheckRun, which is part of the PullRequestBot, is currently failing to evaluate https://github.com/openjdk/jol/pull/24. The failure happens in PullRequestUtils::containsForeignMerge where we are looking for merges in the PR branch which bring in commits that aren't descendants of the main merge parent. If this test is positive, the bot adds a comment with a warning.

The problem is that GitRepository::mergeBase doesn't differentiate between the git command failing with an error and when it just can't find a merge-base. For this check, a non existent merge-base should qualify for a positive test.

I'm refactoring Repository::mergeBase into two methods, one that returns an Optional, which will be empty when a merge-base can't be found. I've also gone through all callers to identify when the new method should be used to properly tell this situation apart.

---------
### Progress
- [x] Change must be properly reviewed (1 reviews required, with at least 1 reviewer)

### Issue
 * [SKARA-1424](https://bugs.openjdk.java.net/browse/SKARA-1424): CheckRun fails to evaluate jol#24

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1312/head:pull/1312` \
`$ git checkout pull/1312`

`$ git checkout pull/1312` \
`$ git pull https://git.openjdk.java.net/skara pull/1312/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1312`

`$ git pr show -t 1312`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1312.diff">https://git.openjdk.java.net/skara/pull/1312.diff</a>

</details>

### DIFF
--- a/forge/src/main/java/org/openjdk/skara/forge/PullRequestUtils.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/PullRequestUtils.java
@@ -128,11 +128,8 @@ public class PullRequestUtils {
         var sourceHead = fetchMergeSource(pr, localRepo);
 
         // Ensure that the source and the target are related
-        try {
-            localRepo.mergeBase(targetHash(localRepo), sourceHead);
-        } catch (IOException e) {
-            throw new CommitFailure("The target and the source branches do not share common history - cannot merge them.");
-        }
+        localRepo.mergeBaseOptional(targetHash(localRepo), sourceHead)
+                .orElseThrow(() -> new CommitFailure("The target and the source branches do not share common history - cannot merge them."));
 
         // Find the most recent commit from the merge source not present in the target
         var sourceHash = localRepo.mergeBase(pr.headHash(), sourceHead);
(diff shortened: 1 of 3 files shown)


forge/src/main/java/org/openjdk/skara/forge/PullRequestUtils.java
```diff
@@ -128,11 +128,8 @@ private static Hash findSourceHash(PullRequest pr, Repository localRepo, List<Co
         var sourceHead = fetchMergeSource(pr, localRepo);
 
         // Ensure that the source and the target are related
-        try {
-            localRepo.mergeBase(targetHash(localRepo), sourceHead);
-        } catch (IOException e) {
-            throw new CommitFailure("The target and the source branches do not share common history - cannot merge them.");
-        }
+        localRepo.mergeBaseOptional(targetHash(localRepo), sourceHead)
+                .orElseThrow(() -> new CommitFailure("The target and the source branches do not share common history - cannot merge them."));
```
erikj79, line 132: Here I think a real error should be let through so the whole WorkItem may be retried later.

vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
```diff
@@ -919,16 +919,28 @@ public void prune(Branch branch, String remote) throws IOException {
     }
 
     @Override
-    public Hash mergeBase(Hash first, Hash second) throws IOException {
+    public Optional<Hash> mergeBaseOptional(Hash first, Hash second) throws IOException {
         try (var p = capture("git", "merge-base", first.hex(), second.hex())) {
-            var res = await(p);
```
erikj79, line 924: await(p) basically does p.await() and then throws IOException if res.status() is non zero.

kevinrushforth (approved):
Looks good.

PR #1320, opened 2022-05-25
SKARA-1448

erikj79:
This patch makes it possible to define specialized "pool" versions with suffixes for 8u in JBS, similar to how it's already possible for 11u+. For example: we could create a JBS `8-pool-foo` version that got consumed by commits with fixVersion `8u331-foo`.

### DIFF
--- a/jbs/src/main/java/org/openjdk/skara/jbs/JdkVersion.java
+++ b/jbs/src/main/java/org/openjdk/skara/jbs/JdkVersion.java
@@ -32,7 +32,7 @@ public class JdkVersion implements Comparable<JdkVersion> {
     private final String opt;
     private final String build;
 
-    private final static Pattern jdkVersionPattern = Pattern.compile("(5\\.0|[1-9][0-9]?)(u([0-9]{1,3}))?$");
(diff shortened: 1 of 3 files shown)


jbs/src/main/java/org/openjdk/skara/jbs/JdkVersion.java
```diff
@@ -56,6 +56,10 @@ private static List<String> splitComponents(String raw) {
                 if (legacyMatcher.group(3) != null) {
                     finalComponents.add(legacyMatcher.group(3));
                 }
+                if (legacyMatcher.groupCount() > 3 && legacyMatcher.group(4) != null) {
```
kevinrushforth, line 59: Should that be `> 4`? if it was equal to 4 wouldn't `group(4)` be OOB?

erikj79: Group 0 is the complete expression and is not included in the count by groupCount(). Perhaps `>= 4` is more readable as the number 4 is the relevant one.

PR #1327, opened 2022-05-28
1433: The change of the CSR status doesn't force the CheckWorkItem to re-run the check

lgxbslgx:
Hi all,

Currently, the CheckWorkItem of the PR bot doesn't re-run the check in the following two situations so that the message of the PR body is not updated and then confuses the developers.

Situation 1 steps:
- The issue has no csr at first
- Use command `/csr needed` (now the PR has `csr` label)
- Create a csr for this issue
- Then the CheckWorkItem doesn't re-run the check (Because the `csr` label is kept in the PR and no new metadata found by CheckWorkItem)
- (optionally) use command `/csr needed` again
- (optionally) then the CheckWorkItem also doesn't re-run the check

Situation 2 steps:
- The issue has no csr at first (now the PR doesn't have `csr` label)
- Add the new fix version to the existing approved csr. (Now the issue has the csr issue. And because this csr issue had been approved, PR also doesn't have `csr` label.)
- then the CheckWorkItem doesn't re-run the check (No new metadata found by CheckWorkItem)
- (optionally) use command `/csr needed`
- (optionally) then the CheckWorkItem doesn't re-run the check

It is because these two situations don't add or remove the `csr` label so that the `CheckWorkItem#getMetadata` doesn't return new metadata and then `CheckWorkItem#currentCheckValid` returns true and then the check doesn't re-run.

This patch adds a new tag/marker named `<!-- csr:` so that `CheckWorkItem#getMetadata` can generate new metadata from it. When the developers use the `/csr` command or the csr issue status has been modified, the CSRBot or CSRCommand will add the `<!-- csr:` tag to the PR body or the comment. Then the CheckWorkItem will re-run the check because `CheckWorkItem#getMetadata` meets the newly added `<!-- csr:` tag.

Some test cases are added or adjusted.

Best Regards,
-- Guoxiong

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.java.net/bylaws#reviewer))

### Issue
 * [SKARA-1433](https://bugs.openjdk.java.net/browse/SKARA-1433): The change of the CSR status doesn't force the CheckWorkItem to re-run the check

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/skara pull/1327/head:pull/1327` \
`$ git checkout pull/1327`

`$ git checkout pull/1327` \
`$ git pull https://git.openjdk.java.net/skara pull/1327/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1327`

`$ git pr show -t 1327`

</details>

<a href="https://git.openjdk.java.net/skara/pull/1327.diff">https://git.openjdk.java.net/skara/pull/1327.diff</a>

</details>

### DIFF
--- a/bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
+++ b/bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
@@ -37,6 +37,7 @@ import java.util.logging.Logger;
 
 class CSRBot implements Bot, WorkItem {
     private final static String CSR_LABEL = "csr";
+    private final static String CSR_UPDATE_MARKER = "\n<!-- csr: 'update' -->\n";
     private final Logger log = Logger.getLogger("org.openjdk.skara.bots");;
     private final HostedRepository repo;
     private final IssueProject project;
@@ -62,7 +63,7 @@ class CSRBot implements Bot, WorkItem {
     /**
      * Get the fix version from the provided PR.
      */
-    public static Optional<JdkVersion> getVersion(PullRequest pullRequest) {
+    private static Optional<JdkVersion> getVersion(PullRequest pullRequest) {
         var confFile = pullRequest.repository().fileContents(".jcheck/conf", pullRequest.targetRef());
         var configuration = JCheckConfiguration.parse(confFile.lines().toList());
         var version = configuration.general().version().orElse(null);
@@ -70,6 +71,22 @@ class CSRBot implements Bot, WorkItem {
             return Optional.empty();
         }
         return JdkVersion.parse(version);
+    }
+
+    private boolean hasCsrIssueAndProgress(PullRequest pr, Issue csr) {
+        return hasCsrIssue(pr, csr) &&
+                (pr.body().contains("- [ ] Change requires a CSR request to be approved") ||
+                 pr.body().contains("- [x] Change requires a CSR request to be approved"));
+    }
+
+    private boolean hasCsrIssueAndProgressChecked(PullRequest pr, Issue csr) {
+        return hasCsrIssue(pr, csr) && pr.body().contains("- [x] Change requires a CSR request to be approved");
+    }
+
+    private boolean hasCsrIssue(PullRequest pr, Issue csr) {
+        return pr.body().contains(csr.id()) &&
+                pr.body().contains(csr.webUrl().toString()) &&
+                pr.body().contains(csr.title() + " (**CSR**)");
     }
 
     @Override
@@ -102,6 +119,13 @@ class CSRBot implements Bot, WorkItem {
             var csr = csrOptional.get();
 
             log.info("Found CSR for " + describe(pr) + ". It has id " + csr.id());
+            if (!hasCsrIssueAndProgress(pr, csr)) {
+                // If the PR body doesn't have the CSR issue or doesn't have the CSR progress,
+                // this bot need to add the csr update marker so that the PR bot can update the message of the PR body.
+                log.info("The PR body doesn't have the CSR issue or progress, adding the csr update marker for " + describe(pr));
+                pr.setBody(pr.body() + CSR_UPDATE_MARKER);
+            }
+
             var resolution = csr.properties().get("resolution");
             if (resolution == null || resolution.isNull()) {
                 if (!pr.labelNames().contains(CSR_LABEL)) {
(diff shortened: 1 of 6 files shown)


bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
```diff
@@ -72,6 +73,22 @@ public static Optional<JdkVersion> getVersion(PullRequest pullRequest) {
...
+
+    private boolean hasCsrIssueAndProgressChecked(PullRequest pr, Issue csr) {
+        return hasCsrIssue(pr, csr) && pr.body().contains("- [x] Change requires a CSR request to be approved");
+    }
+
+    private boolean hasCsrIssue(PullRequest pr, Issue csr) {
+        return pr.body().contains(csr.id()) &&
+                pr.body().contains(csr.webUrl().toString()) &&
+                pr.body().contains(csr.title() + " (**CSR**)");
+    }
```
erikj79, line 90: We could consider making these checks safer by only looking in the Skara generated part of the body. There is a delimiter line you could split on.

bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
```diff
@@ -102,6 +119,13 @@ public Collection<WorkItem> run(Path scratchPath) {
             var csr = csrOptional.get();
 
             log.info("Found CSR for " + describe(pr) + ". It has id " + csr.id());
```
erikj79, line 121: I reacted to this log message the other day. Could you shorten it like this while in the area?

```suggestion
            log.info("Found CSR " + csr.id() + " for " + describe(pr) + ");
```

bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
```diff
@@ -153,6 +177,12 @@ public Collection<WorkItem> run(Path scratchPath) {
                 log.info("CSR closed and approved for " + describe(pr) + ", removing CSR label");
                 pr.removeLabel(CSR_LABEL);
             }
+            if (!hasCsrIssueAndProgressChecked(pr, csr)) {
+                // If the PR body doesn't have the CSR issue or doesn't have the CSR progress or the CSR progress checkbox is not selected,
+                // this bot need to add the csr update marker so that the PR bot can update the message of the PR body.
+                log.info("CSR closed and approved for " + describe(pr) + ", adding the csr update marker");
+                pr.setBody(pr.body() + CSR_UPDATE_MARKER);
```
erikj79, line 184: We need to make sure we only add the update marker once. Otherwise we would end up adding a very large amount of these if the PR bot happens to be stuck for a while. I would recommend a method for adding the marker, which first checks if the body contains the marker already.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CSRCommand.java
```diff
@@ -33,6 +33,8 @@
 
 public class CSRCommand implements CommandHandler {
     private static final String CSR_LABEL = "csr";
+    private static final String CSR_NEEDED_MARKER = "<!-- csr: 'needed' -->";
+    private static final String CSR_UNNEEDED_MARKER = "<!-- csr: 'unneeded' -->";
```
erikj79, line 37: Did you intend to look for this in the CSRBot?

bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
lgxbslgx: Fixed.

bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBot.java
lgxbslgx: Fixed.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CSRCommand.java
erikj79: I don't understand what you mean. You set these markers, but I don't see anything looking for them.

lgxbslgx:
```java
// file CheckWorkItem.java
-    private final Pattern metadataComments = Pattern.compile("<!-- (?:(add|remove) (?:contributor|reviewer))|(?:summary: ')|(?:solves: ')|(?:additional required reviewers)|(?:jep: ')");
+    private final Pattern metadataComments = Pattern.compile("<!-- (?:(add|remove) (?:contributor|reviewer))|(?:summary: ')|(?:solves: ')|(?:additional required reviewers)|(?:jep: ')|(?:csr: ')");
```

The newly added `(?:csr: ')` can search it and update the metadata so that the check can be re-run.

erikj79: Ah of course, thanks.

erikj79 (approved):

lgxbslgx:
@erikj79 Thanks for the review. Integrating.

/integrate

PR #1330, opened 2022-06-06
SKARA-1465

erikj79:
The altfixversions feature introduced in [SKARA-1213](https://bugs.openjdk.java.net/browse/SKARA-1213) had the unfortunate side effect of also blocking additional commit comments from being added to existing backports. I think this needs to be fixed as those additional commit comments help integrators track where and when changes move between different repos.

This patch fixes this and adjusts the relevant test.


magicus (approved):
LGTM

PR #1335, opened 2022-06-13
SKARA-1475

erikj79:
This patch adds a configuration option for the IssueNotifier. This new option controls how tags are matched with fixVersion when updating the Resolved In Build field. Currently, any "opt" string in the fixVersion (e.g. 17.0.4-oracle) is ignored as tags are do not include the 'oracle' opt string. We now have the need to require tags to include such a string for some releases/versions. I believe this should be the default behavior and ignoring "oracle" should be the exception.

The new option goes into the issue part of the notifier block. By default, the opt string will be considered part of the version string and required in the tag for a match (this is the opposite of the current behavior). By using the new option, a set of opt strings that should not be required in a tag for a match may be configured. In the example below a tag `jdk8u341-b01` would match a fix version of `8u341-foo`.

```
"issue": {
  "tag": {
    "ignoreopt": [ "foo", "bar" ],
  }
}
```

### DIFF
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
@@ -70,6 +70,10 @@ class IssueNotifier implements Notifier, PullRequestListener, RepositoryListener
     // pronly is true.
     private final boolean resolve;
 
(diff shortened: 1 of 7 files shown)


erikj79:
Pinging to trigger new email.

magicus (approved):
It looks good to me, but my understanding about the entire fixVersion thingie is admittedly a bit hand-wavy. I think it would be good if @kevinrushforth could have a look at it as well.

PR #1341, opened 2022-07-09
1495: /backport should be allowed in PRs which have been integrated

lgxbslgx:
Hi all,

This patch mainly solves the issue described at SKARA-1495 which makes the `backport` command be valid in a pull request which has been integrated.

But in detail, this patch has the following feature or solves the following issues:

- Currently, the commands of the pull request and the commands of the commit are not related and can't be known from each other. So when we use the commit commands, such as `tag`, in the pull request, the bot replies `Unknown command tag`. Actually, the bot should replies the message like `The command tag can not be used in pull requests.  Please try to use this command on the commit`. The change is mainly at method `CommandExtractor::extractCommands`.
- Many places of the code use a duplicated method, named `resultingCommitHash`, this method could be put into class `PullRequest` to avoid the duplication. I refactor it in this patch.
- Now the `backport` command can not be used in pull requests. I revise the logic to let it available in a pull request which has been integrated, mainly removing the overrided method `allowedInPullRequest` of the class `BackportCommand`.
- The reason that these issues are not found before is that we hadn't tested all the branches of the code, especially the branches in methods `PullRequestCommandWorkItem::processCommand` and `CommitCommandWorkItem::processCommand`. So I add more test cases to test almost all the branches. And if the branch can't be tested, I add a comment to explain it.

Although it seems that many features are not related to SKARA-1495, I put them together as an enhancement of the PullRequestBot.

Best Regards,
-- Guoxiong

---------
### Progress

### Issue
 * [SKARA-1495](https://bugs.openjdk.org/browse/SKARA-1495): /backport should be allowed in PRs which have been integrated

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1341/head:pull/1341` \
`$ git checkout pull/1341`

`$ git checkout pull/1341` \
`$ git pull https://git.openjdk.org/skara pull/1341/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1341`

`$ git pr show -t 1341`

</details>

<a href="https://git.openjdk.org/skara/pull/1341.diff">https://git.openjdk.org/skara/pull/1341.diff</a>

</details>

### DIFF
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/PullRequestWorkItem.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/PullRequestWorkItem.java
@@ -184,7 +184,7 @@ public class PullRequestWorkItem implements WorkItem {
                 .materialize(historyPath);
 
         var issues = parseIssues();
-        var commit = pr.findIntegratedCommitHash(List.of(integratorId)).orElse(null);
+        var commit = integratorId != null ? pr.findIntegratedCommitHash(List.of(integratorId)).orElse(null) : null;
         var state = new PullRequestState(pr, issues, commit, pr.headHash(), pr.state());
         var stored = storage.current();
         if (stored.contains(state)) {
@@ -199,7 +199,7 @@ public class PullRequestWorkItem implements WorkItem {
         // The stored entry could be old and be missing commit information - if so, upgrade it
         if (storedState.isPresent()) {
             if (storedState.get().commitId().equals(Optional.of(Hash.zero()))) {
-                var hash = pr.findIntegratedCommitHash(List.of(integratorId)).orElse(null);
+                var hash = integratorId != null ? pr.findIntegratedCommitHash(List.of(integratorId)).orElse(null) : null;
                 storedState = Optional.of(new PullRequestState(pr, storedState.get().issueIds(), hash, pr.headHash(), pr.state()));
                 storage.put(storedState.get());
             }
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/PullRequestWorkItem.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/PullRequestWorkItem.java
@@ -50,6 +50,10 @@ public class PullRequestWorkItem implements WorkItem {
         this.integratorId = integratorId;
     }
 
+    private Hash resultingCommitHash() {
+        return integratorId != null ? pr.findIntegratedCommitHash(List.of(integratorId)).orElse(null) : null;
+    }
+
     private Set<PullRequestState> deserializePrState(String current) {
         if (current.isBlank()) {
             return Set.of();
@@ -184,7 +188,7 @@ public class PullRequestWorkItem implements WorkItem {
                 .materialize(historyPath);
 
         var issues = parseIssues();
-        var commit = integratorId != null ? pr.findIntegratedCommitHash(List.of(integratorId)).orElse(null) : null;
+        var commit = resultingCommitHash();
         var state = new PullRequestState(pr, issues, commit, pr.headHash(), pr.state());
         var stored = storage.current();
         if (stored.contains(state)) {
@@ -199,7 +203,7 @@ public class PullRequestWorkItem implements WorkItem {
         // The stored entry could be old and be missing commit information - if so, upgrade it
         if (storedState.isPresent()) {
             if (storedState.get().commitId().equals(Optional.of(Hash.zero()))) {
-                var hash = integratorId != null ? pr.findIntegratedCommitHash(List.of(integratorId)).orElse(null) : null;
+                var hash = resultingCommitHash();
                 storedState = Optional.of(new PullRequestState(pr, storedState.get().issueIds(), hash, pr.headHash(), pr.state()));
                 storage.put(storedState.get());
             }
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/PullRequestWorkItem.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/PullRequestWorkItem.java
@@ -24,7 +24,6 @@ package org.openjdk.skara.bots.notify;
 
 import org.openjdk.skara.bot.WorkItem;
 import org.openjdk.skara.forge.PullRequest;
-import org.openjdk.skara.issuetracker.Comment;
 import org.openjdk.skara.json.*;
 import org.openjdk.skara.storage.StorageBuilder;
 import org.openjdk.skara.vcs.Hash;
@@ -49,23 +48,6 @@ public class PullRequestWorkItem implements WorkItem {
         this.listeners = listeners;
         this.errorHandler = errorHandler;
         this.integratorId = integratorId;
-    }
-
-    private final static Pattern pushedPattern = Pattern.compile("Pushed as commit ([a-f0-9]{40})\\.");
-
-    private Hash resultingCommitHash() {
-        if (pr.labelNames().contains("integrated")) {
-            return pr.comments().stream()
-                     .filter(comment -> comment.author().id().equals(integratorId))
-                     .map(Comment::body)
-                     .map(pushedPattern::matcher)
-                     .filter(Matcher::find)
-                     .map(m -> m.group(1))
-                     .map(Hash::new)
-                     .findAny()
-                     .orElse(null);
-        }
-        return null;
     }
 
     private Set<PullRequestState> deserializePrState(String current) {
@@ -202,7 +184,7 @@ public class PullRequestWorkItem implements WorkItem {
                 .materialize(historyPath);
 
         var issues = parseIssues();
-        var commit = resultingCommitHash();
+        var commit = pr.findIntegratedCommitHash(List.of(integratorId)).orElse(null);
         var state = new PullRequestState(pr, issues, commit, pr.headHash(), pr.state());
         var stored = storage.current();
         if (stored.contains(state)) {
@@ -217,7 +199,7 @@ public class PullRequestWorkItem implements WorkItem {
         // The stored entry could be old and be missing commit information - if so, upgrade it
         if (storedState.isPresent()) {
             if (storedState.get().commitId().equals(Optional.of(Hash.zero()))) {
-                var hash = resultingCommitHash();
+                var hash = pr.findIntegratedCommitHash(List.of(integratorId)).orElse(null);
                 storedState = Optional.of(new PullRequestState(pr, storedState.get().issueIds(), hash, pr.headHash(), pr.state()));
                 storage.put(storedState.get());
             }
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/BackportCommand.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/BackportCommand.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2020, 2022, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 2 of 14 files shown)


magicus:
Thank you for tackling this PR. Both Erik and I are on vacation right now, and you'll have to wait to get a full review.

My main worry here is that there were some underlying reason for not going this route with `/backport` on the PRs directly, that is not obvious to me, and that may make this change unsuitable.

lgxbslgx:
> Thank you for tackling this PR. Both Erik and I are on vacation right now, and you'll have to wait to get a full review.
> 
> My main worry here is that there were some underlying reason for not going this route with `/backport` on the PRs directly, that is not obvious to me, and that may make this change unsuitable.

Thanks for the reply. Still waiting for comment or review.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/BackportCommand.java
```diff
@@ -52,8 +53,9 @@ public boolean allowedInCommit() {
     }
 
     @Override
-    public boolean allowedInPullRequest() {
-        return false;
+    public void handle(PullRequestBot bot, PullRequest pr, CensusInstance censusInstance, Path scratchPath,
+                CommandInvocation command, List<Comment> allComments, PrintWriter reply) {
+        reply.println("The command `backport` can not be used in a pull request that has not yet been integrated.");
```
erikj79, line 58: Shouldn't the message here be the one suggested in the Bug?

"(!) You cannot issue `/backport` on a pull request. Instead, give that command on the actual commit you want to backport."

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
```diff
@@ -184,11 +171,14 @@ private void processCommand(PullRequest pr, CensusInstance censusInstance, Path
         if (handler.isPresent()) {
             if (isCommit) {
                 if (handler.get().allowedInCommit()) {
-                    var hash = resultingCommitHash(allComments);
+                    var hash = pr.findIntegratedCommitHash();
```
erikj79, line 174: It's worth noting that this will trigger a new `comments()` call to the forge instead of reusing the existing list of comments.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
```diff
@@ -212,7 +202,7 @@ private void processCommand(PullRequest pr, CensusInstance censusInstance, Path
                 } else {
                     printer.print("The command `");
                     printer.print(command.name());
-                    printer.println("` can only be used in a pull request that has not yet been integrated.");
+                    printer.println("` can not be used in pull requests. Please try to use this command on the commit.");
```
erikj79, line 205: I don't think we need to suggest using the command on a commit in the general case.

```suggestion
                    printer.println("` cannot be used in pull requests.");
```

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
```diff
@@ -40,9 +39,8 @@ public class PullRequestCommandWorkItem extends PullRequestWorkItem {
 
     private static final String commandReplyMarker = "<!-- Jmerge command reply message (%s) -->";
     private static final Pattern commandReplyPattern = Pattern.compile("<!-- Jmerge command reply message \\((\\S+)\\) -->");
-    private final static Pattern pushedPattern = Pattern.compile("Pushed as commit ([a-f0-9]{40})\\.");
 
-    private static final Map<String, CommandHandler> commandHandlers = Map.ofEntries(
+    static final Map<String, CommandHandler> commandHandlers = Map.ofEntries(
```
erikj79, line 43: Instead of making these package private and adding the convoluted logic in extractCommands, can't we just add BackportCommand to the commandHandlers for PullRequestWorkItem?

It seems like the original author mixed two ways of separating commands between PRs and commits, both through these handler lists, and through the allowedIn* methods. If we really want to improve all the error messages, then all commands should be added to both lists, or maybe the lists should be unified into one.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CommandExtractor.java
```diff
@@ -52,7 +52,10 @@ static List<CommandInvocation> extractCommands(Map<String, CommandHandler> comma
                     multiLineHandler = null;
                 }
                 var command = commandMatcher.group(1).toLowerCase();
-                var handler = commandHandlers.get(command);
+                var handler = commitCommandFirst ? CommitCommandWorkItem.commandHandlers.get(command) : PullRequestCommandWorkItem.commandHandlers.get(command);
```
erikj79, line 55: What are you trying to achieve with the commitCommandFirst boolean? In what situation would it be wrong to just use one unified list?

forge/src/main/java/org/openjdk/skara/forge/PullRequest.java
```diff
@@ -176,4 +179,24 @@ default boolean isSame(PullRequest other) {
...
      */
     Optional<ZonedDateTime> lastForcePushTime();
+
+    /**
+     * Return the commit hash if the pull request was integrated.
+     */
+    Optional<Hash> findIntegratedCommitHash();
+
+    default Optional<Hash> findIntegratedCommitHash(List<String> userIds) {
+        Pattern pushedPattern = Pattern.compile("Pushed as commit ([a-f0-9]{40})\\.");
```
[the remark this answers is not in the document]
lgxbslgx: What about adding a parameter like `Pattern commitComment` to the method `findIntegratedCommitHash`?

bots/notify/src/main/java/org/openjdk/skara/bots/notify/PullRequestWorkItem.java
```diff
@@ -202,7 +184,7 @@ public Collection<WorkItem> run(Path scratchPath) {
                 .materialize(historyPath);
 
         var issues = parseIssues();
-        var commit = resultingCommitHash();
+        var commit = integratorId != null ? pr.findIntegratedCommitHash(List.of(integratorId)).orElse(null) : null;
```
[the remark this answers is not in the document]
lgxbslgx: Fixed.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/BackportCommand.java
lgxbslgx: I want the`/backport` command can be used in the PRs which hava been integrated. So the message `You cannot issue /backport on a pull request` seems not accurate. But the current message seems not accurate too. Looking for a better description and your idea.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
lgxbslgx: Yes, it is a tradeoff between code concision and the running time. If we find the code becomes slower in the future and want to optimize it, this is a good place to optimize. But now we can keep it because we want to refactor the method in this PR.

And the method `pr.findIntegratedCommitHash` always get the newest comments so it may avoid some issues that the `Pushed as ...` comment is added during processing command. But I am not really know whether this situation would happen.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
lgxbslgx: Fixed.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
lgxbslgx: I unified these two lists into one and put it in class `CommandExtractor`. All the tests in PR bot passed locally.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/BackportCommand.java
lgxbslgx: Additional information: When the PR has been integrated and the `backport` command is used in this integrated PR, the another `handle` method  instead of this method will be invoked so that the `backport` command can be processed.

forge/src/main/java/org/openjdk/skara/forge/PullRequest.java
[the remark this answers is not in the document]
erikj79: The drawback of that is that we then need to have a copy of the pattern everywhere we call the method. I think it's better to have something like a static method on PullRequest like `String commitHashMessage(Hash)`, which takes a hash and returns a String with the formatted message. IntegrateCommand can use that to format the relevant part of the comment.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/BackportCommand.java
lgxbslgx: In this patch, it can be used at PRs which had been integrated. This is a mian feature of this patch.

erikj79: Oh, I see now that PullRequestCommandWorkItem will call the handle method for commit if the PR has been integrated. That was rather surprising behavior. In that case, please align the reply with the one in PullRequestCommandWorkItem:

printer.print("The command `");
printer.print(command.name());
printer.println("` can only be used in a pull request that has been integrated.");

forge/src/main/java/org/openjdk/skara/forge/PullRequest.java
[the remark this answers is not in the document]
lgxbslgx: Maybe someone has another more elegant way to solve this issue. But now I can't find such way, so it is good to adopt your idea now.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/BackportCommand.java
lgxbslgx: > please align the reply with the one in PullRequestCommandWorkItem

Fixed.

> Also, if that's the case, please change the bug and PR title to make this clear. That's the main feature being added here.

I revise the title just now.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CommandExtractor.java
erikj79: Hm, now I realize what you mean about the HelpCommand. If you invoke `/help` on a commit now, you get all the PR commands listed. That's not good. We still need some way of separating which commands should be listed for help with a commit and help with a PR.

lgxbslgx: > The only one situation I concern about is the help command. But after second thought, it looks good. So I unified them.

In detail, the code logic of the master branch:

1. use `help` command on PR which has **not** been integrated

the bot prints the commands which are **allowed in PR** and **the extranal PR command**

2. use `help` command on PR which **has been integrated**

the bot prints the commands which are **allowed in commit** and **the extranal PR command**

3. use `help` command on a commit

the bot prints the commands which are **allowed in commit** and **the extranal commit command**

But the logic of this patch is:
1. use `help` command on PR which has **not** been integrated (unchanged)

the bot prints the commands which are **allowed in PR** and **the extranal PR command**

2. use `help` command on PR which **has been integrated** or use `help` **command on a commit**

the bot prints the commands which are **allowed in commit** and **the extranal commit command**

So there is a little diffrence between the master branch and my patch. But the logic in my patch seems good.

> Hm, now I realize what you mean about the HelpCommand. If you invoke `/help` on a commit now, you get all the PR commands listed. That's not good. We still need some way of separating which commands should be listed for help with a commit and help with a PR.

From your description, I think one among us misunderstood the code. Did I miss something?

erikj79: You are right, I missed that you added filtering on allowedIn*. That was an elegant solution, thanks!

erikj79 (approved):

PR #1348, opened 2022-08-05
1529: Optimize Census namespace instantiation

erikj79:
This patch adds a new factory method `Census::parseNamespace`, which enables initialization of a single Namespace directly from a HostedRepository (without having to clone the census repo to local disk first). 

The motivation behind this is bot performance. The Census is loaded in many WorkItems. Cloning/fetching the complete census repo to local disk is relatively expensive compared to a couple of REST calls to just read files directly from the remote repo. Especially as our REST client already leverages caching through ETag. This particular optimization can however only be applied to the IssueNotifier in the NotifyBot, where I doubt it will make that much of a difference. The main performance gain is realized in a different bot which currently isn't public (but these gains will impact all Skara users).

I chose to at least for now, only apply this for initializing a single Namespace as that can be done by reading only 2 files from the remote, so the performance gain is very clear. Other bots either need to read considerably more files, or have the need to update the census. This optimization idea may be expanded in the future if I see a need and noticeable gains from it. 

In order to make HostedRepository available to the Census class, I had to get rid of several existing module dependencies to avoid circular dependencies. Fortunately this wasn't too hard. It made no sense to me that `org.openjdk.skara.forge` would require `org.openjdk.skara.census`. There was just one static method `PullRequestUtils` in the forge module that used a reference to the census module, and this static method was only used from one module `org.openjdk.skara.bots.pr`, so I simply moved it to `CheckablePullRequest` which is kind of a utility class in that module. The rest of the offending dependencies were just unused, so I cleaned out a fair few of those.

In the `IssueNotifier`, in addition to using the new initialization method, I also added lazy initialization and reusing of the `CensusInstance` class, to avoid re-initializing the same data for every round in loops.

---------
### Progress

### Issue
 * [SKARA-1529](https://bugs.openjdk.org/browse/SKARA-1529): Optimize Census namespace instantiation

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1348/head:pull/1348` \
`$ git checkout pull/1348`

`$ git checkout pull/1348` \
`$ git pull https://git.openjdk.org/skara pull/1348/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1348`

`$ git pr show -t 1348`

</details>

<a href="https://git.openjdk.org/skara/pull/1348.diff">https://git.openjdk.org/skara/pull/1348.diff</a>

</details>

### DIFF
--- a/census/src/main/java/module-info.java
+++ b/census/src/main/java/module-info.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2022, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -22,6 +22,7 @@
  */
 module org.openjdk.skara.census {
     requires org.openjdk.skara.xml;
+    requires static org.openjdk.skara.forge;
     requires java.xml;
     requires java.net.http;
     requires java.logging;
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/CensusInstance.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/CensusInstance.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2022, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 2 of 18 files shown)


census/src/main/java/module-info.java
```diff
@@ -22,6 +22,7 @@
  */
 module org.openjdk.skara.census {
     requires org.openjdk.skara.xml;
+    requires static org.openjdk.skara.forge;
```
magicus, line 25: What does `requires static` mean in a `module-info.java` file?

magicus (approved):
Looks good to me.

census/src/main/java/module-info.java
erikj79: It's the java modules way of declaring a dependency optional. It's only needed at compile time, but not at runtime. 

At least in this case, the dependency is only used in a static method, which means any types from the forge module would only be used if you call that static method, and to be able to do that, you would already have had to load those classes.

I don't think it matters in Skara currently, as I think all modules that require census also require forge, but it feels a bit cleaner to me that pulling in forge is optional here.

PR #1349, opened 2022-08-05
1530: Race with CheckWorkItem in PR bot

erikj79:
This patch fixes a race in the PR bot. 

Currently, all WorkItems that process PRs get an instance of the PullRequest that was  fetched before the WorkItem instance was created. The data in that PullRequest instance (the parts that aren't dynamically fetched on method calls) may then be stale before the WorkItem even starts running.

With this patch, I'm changing PullRequestWorkItem (the super class of all WorkItems that handle PRs), to only be instantiated with the ID of the PR, and then fetches the actual PullRequest object at the start of the `run()` method. When the `run()` method is called, we are guaranteed to be the only WorkItem currently processing the particular PR, so fetching data there should guarantee a coherent and current view.

In many cases where PullRequestWorkItems were created, the creator first refreshed the PullRequest object to provide an updated view, reflecting any changes made by the creator. These refreshes are no longer necessary, so I've removed them all.

---------
### Progress

### Issue
 * [SKARA-1530](https://bugs.openjdk.org/browse/SKARA-1530): Race with CheckWorkItem in PR bot

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1349/head:pull/1349` \
`$ git checkout pull/1349`

`$ git checkout pull/1349` \
`$ git pull https://git.openjdk.org/skara pull/1349/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1349`

`$ git pr show -t 1349`

</details>

<a href="https://git.openjdk.org/skara/pull/1349.diff">https://git.openjdk.org/skara/pull/1349.diff</a>

</details>

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestWorkItem.java
@@ -22,6 +22,8 @@
  */
 package org.openjdk.skara.bots.pr;
 
+import java.nio.file.Path;
+import java.util.Collection;
 import org.openjdk.skara.bot.WorkItem;
 import org.openjdk.skara.forge.PullRequest;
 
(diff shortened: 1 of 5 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestWorkItem.java
```diff
@@ -30,26 +32,42 @@
...
 
     @Override
     public final boolean concurrentWith(WorkItem other) {
-        if (!(other instanceof PullRequestWorkItem)) {
+        if (!(other instanceof PullRequestWorkItem otherItem)) {
             return true;
         }
-        PullRequestWorkItem otherItem = (PullRequestWorkItem)other;
-        if (!pr.isSame(otherItem.pr)) {
+        if (!(prId.equals(otherItem.prId) && bot.repo().isSame(otherItem.bot.repo()))) {
```
kevinrushforth, line 49: This means that two work requests can run concurrently if either the `prId` is different or the `bot repo` is different. I presume this is because the `prId` is relative to the repo (which makes sense), right?

kevinrushforth (approved):
Looks like a good solution to me.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestWorkItem.java
erikj79: Correct. The prId is the number we see in PR links, e.g. 1349 for this one. The implementation of PullRequest::isSame is equivalent to the new code. The problem was that I can no longer guarantee to have the PullRequest object available when `concurrentWith` is called (rather, it's guaranteed to not be there), so I had to basically copy the code, and get the HostedRepository from the bot instead of the pr. This condition is supposed to be exactly the same as it was pre patch.

magicus (approved):
Looks good. Tricky one!

PR #1355, opened 2022-08-11
1539: Allow dots in merge PR branch name

erikj79:
This patch makes it possible to create merge style PRs based on branches with periods `.` in the branch name. We quite commonly put periods in branch names as we like to base branches on version strings.

Modified an existing test to cover this case.

---------
### Progress

### Issue
 * [SKARA-1539](https://bugs.openjdk.org/browse/SKARA-1539): Allow dots in merge PR branch name

### Reviewers
 * [Kevin Rushforth](https://openjdk.org/census#kcr) (@kevinrushforth - **Reviewer**) ⚠️ Review applies to [bb6825b9](https://git.openjdk.org/skara/pull/1355/files/bb6825b9bd580f7d10c39300795dfa13ace9707d)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1355/head:pull/1355` \
`$ git checkout pull/1355`

`$ git checkout pull/1355` \
`$ git pull https://git.openjdk.org/skara pull/1355/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1355`

`$ git pr show -t 1355`

</details>

<a href="https://git.openjdk.org/skara/pull/1355.diff">https://git.openjdk.org/skara/pull/1355.diff</a>

</details>

### DIFF
--- a/forge/src/main/java/org/openjdk/skara/forge/PullRequestUtils.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/PullRequestUtils.java
@@ -39,7 +39,7 @@ public class PullRequestUtils {
                                 committer.name(), committer.email(), ZonedDateTime.now(), List.of(targetHash(localRepo)), localRepo.tree(finalHead));
     }
 
-    private final static Pattern mergeSourcePattern = Pattern.compile("^Merge ([-/\\w:+]+)$");
+    private final static Pattern mergeSourcePattern = Pattern.compile("^Merge ([-/.\\w:+]+)$");
     private final static Pattern hashSourcePattern = Pattern.compile("[0-9a-fA-F]{6,40}");
 
     private static Optional<Hash> fetchRef(Repository localRepo, URI uri, String ref) {
--- a/bots/pr/src/test/java/org/openjdk/skara/bots/pr/MergeTests.java
+++ b/bots/pr/src/test/java/org/openjdk/skara/bots/pr/MergeTests.java
@@ -62,12 +62,12 @@ class MergeTests {
             localRepo.push(masterHash, author.url(), "master", true);
 
             // Make more changes in another branch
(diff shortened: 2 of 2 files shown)


forge/src/main/java/org/openjdk/skara/forge/PullRequestUtils.java
```diff
@@ -39,7 +39,7 @@ private static Hash commitSquashed(Repository localRepo, Hash finalHead, Author
                                 committer.name(), committer.email(), ZonedDateTime.now(), List.of(targetHash(localRepo)), localRepo.tree(finalHead));
     }
 
-    private final static Pattern mergeSourcePattern = Pattern.compile("^Merge ([-/\\w:+]+)$");
+    private final static Pattern mergeSourcePattern = Pattern.compile("^Merge ([-/.\\w:+]+)$");
```
kevinrushforth, line 42: This will match any character, which may not be what you intend. Shouldn't that be `\\.` instead of just `.`?

kevinrushforth (commented):
I left a question about the fix to the pattern string.

forge/src/main/java/org/openjdk/skara/forge/PullRequestUtils.java
erikj79: My understanding is that the only characters you need to escape inside a character class are ^, -, ] or \, everything else is a literal. Also Intellij gave me a warning when I tried to escape it saying it was unnecessary.

kevinrushforth: I stand corrected. I now see that `.` is a literal dot when it is inside `[]` (when not inside `[]` it matches any char).

kevinrushforth (approved):

PR #1356, opened 2022-08-11
1542: Census Namespace missing hashCode and equals

erikj79:
The Namespace class is missing hashCode and equals. This patch adds implementations for those two methods. The other data classes in the Census module already have them. 

This is preventing the change in [SKARA-1529](https://bugs.openjdk.org/browse/SKARA-1529) from actually taking effect, as we can't compare two instances of Namespace to see if they have the same contents.

---------
### Progress

### Issue
 * [SKARA-1542](https://bugs.openjdk.org/browse/SKARA-1542): Census Namespace missing hashCode and equals

### Reviewers
 * [Kevin Rushforth](https://openjdk.org/census#kcr) (@kevinrushforth - **Reviewer**) ⚠️ Review applies to [e0f5d6ea](https://git.openjdk.org/skara/pull/1356/files/e0f5d6ea32d8fb8b830660887799d201af19cd9f)
 * [Joe Darcy](https://openjdk.org/census#darcy) (@jddarcy - **Reviewer**) ⚠️ Review applies to [e0f5d6ea](https://git.openjdk.org/skara/pull/1356/files/e0f5d6ea32d8fb8b830660887799d201af19cd9f)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1356/head:pull/1356` \
`$ git checkout pull/1356`

`$ git checkout pull/1356` \
`$ git pull https://git.openjdk.org/skara pull/1356/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1356`

`$ git pr show -t 1356`

</details>

<a href="https://git.openjdk.org/skara/pull/1356.diff">https://git.openjdk.org/skara/pull/1356.diff</a>

</details>

### DIFF
--- a/census/src/main/java/org/openjdk/skara/census/Namespace.java
+++ b/census/src/main/java/org/openjdk/skara/census/Namespace.java
@@ -101,7 +101,9 @@ public class Namespace {
             return false;
         }
         Namespace namespace = (Namespace) o;
-        return name.equals(namespace.name) && mapping.equals(namespace.mapping) && reverse.equals(namespace.reverse);
+        return Objects.equals(name, namespace.name)
+                && Objects.equals(mapping, namespace.mapping)
+                && Objects.equals(reverse, namespace.reverse);
     }
 
     @Override
--- a/census/src/main/java/org/openjdk/skara/census/Namespace.java
+++ b/census/src/main/java/org/openjdk/skara/census/Namespace.java
@@ -91,4 +91,21 @@ public class Namespace {
 
         return new Namespace(name, mapping, reverse);
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        Namespace namespace = (Namespace) o;
+        return name.equals(namespace.name) && mapping.equals(namespace.mapping) && reverse.equals(namespace.reverse);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(name, mapping, reverse);
+    }
 }


kevinrushforth (approved):
This looks fine presuming that name, mapping, and reverse are always non-null.

census/src/main/java/org/openjdk/skara/census/Namespace.java
```diff
@@ -91,4 +91,21 @@ static Namespace parse(Element ele, Map<String, Contributor> contributors) throw
 
         return new Namespace(name, mapping, reverse);
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
```
jddarcy, line 100: The "same type" check could be phrased differently using instanceof, but what you have here should work okay too.

jddarcy (approved):

census/src/main/java/org/openjdk/skara/census/Namespace.java
erikj79: Hm, I let the IDE generate these for me so didn't think much of it. Aren't there subtle differences with instanceof, if there are subtypes involved? Not that it matters in this particular instance.

erikj79: So I experimented some more. In the intellij wizard for this, if I check "accept subclasses", it generates code with `instanceof` instead. I don't think this equals method should do that.

erikj79:
> This looks fine presuming that name, mapping, and reverse are always non-null.

Right, I checked the code again, and while the two maps will currently never be null, the name field could potentially end up as null if given bad input data. I think it's good to be defensive and just use the static Object.equals for all the fields. I did this by just rerunning the intellij wizard.

census/src/main/java/org/openjdk/skara/census/Namespace.java
jddarcy: To circle back on this, there is a school of thought that getClass should not be used in this fashion in an equals method since it violates the Liskov substitution principle -- instances of a subclass can be used where instances of the superclass can used. "Effective Java, 3rd edition" covers this topic in item 10. However, in a case like this where (I assume) subclasses are not expected, it doesn't matter very much in practice.

PR #1358, opened 2022-08-15
1355: Mirror bot needs branch selection based on patterns

erikj79:
This patch makes it possible to configure mirrored branches with regex patterns. For what I think is improved readability, I made it possible to supply the patterns as an array of strings, so we can have multiple separate patterns. It's still possible to define "branches" as a comma separated list, for backwards compatibility.

When deploying this change, I will need to delete the existing local repos used for mirroring on the bot runner machine, because the method by which repos are cloned is changing (for repos that do not just mirror "everything"). Going forward, I think it's good to always clone all repos the same way, regardless of configuration.

There is another behavioral change, where I'm adding `--force` when pushing branches with `includeTags`. It happens a bit too often that a repo maintainer changes tags, which currently causes error for any mirroring job because we aren't force pushing tags. I don't think we ever want to not force update when mirroring repos.

---------
### Progress

### Issue
 * [SKARA-1355](https://bugs.openjdk.org/browse/SKARA-1355): Mirror bot needs branch selection based on patterns

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1358/head:pull/1358` \
`$ git checkout pull/1358`

`$ git checkout pull/1358` \
`$ git pull https://git.openjdk.org/skara pull/1358/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1358`

`$ git pr show -t 1358`

</details>

<a href="https://git.openjdk.org/skara/pull/1358.diff">https://git.openjdk.org/skara/pull/1358.diff</a>

</details>

### DIFF
--- a/vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
+++ b/vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
@@ -594,6 +594,9 @@ public class GitRepository implements Repository {
 
         if (includeTags) {
             cmd.add("--tags");
+            if (force) {
+                cmd.add("--force");
+            }
         }
 
         cmd.add(uri.toString());
--- a/bots/mirror/src/main/java/org/openjdk/skara/bots/mirror/MirrorBot.java
+++ b/bots/mirror/src/main/java/org/openjdk/skara/bots/mirror/MirrorBot.java
@@ -22,6 +22,7 @@
  */
 package org.openjdk.skara.bots.mirror;
 
+import java.util.regex.Pattern;
 import org.openjdk.skara.bot.*;
 import org.openjdk.skara.forge.HostedRepository;
 import org.openjdk.skara.vcs.*;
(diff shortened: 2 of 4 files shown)


vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
```diff
@@ -594,6 +594,9 @@ public void push(Hash hash, URI uri, String ref, boolean force, boolean includeT
 
         if (includeTags) {
             cmd.add("--tags");
+            if (force) {
+                cmd.add("--force");
```
kevinrushforth, line 598: Is there a reason not to ignore the `force` flag when not pushing tags? It would seem more consistent to always look at the `force` flag and add `--force` is set. That way a repo configured not to push tags would still be able to handle a source branch that was force-pushed.

kevinrushforth (commented):
Looks good with one question.

vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
erikj79: I was torn about this. I went this way to change as little current behavior as possible, mainly because I don't fully understand the nuances of git behavior here. Note that further down, `force==true` is already handled for branches by prepending `+` to the refs. On the other hand, the current proposal may be inconsistent, which is also bad long term.

Perhaps the best solution would be to get rid of the `+` below and just add `--force` when `force==true`?

kevinrushforth: Using `--force` consistently sounds cleanest, but if you want to limit the scope of this PR and do that in a follow-up PR, that would be good.

kevinrushforth (approved):

magicus (approved):
LGTM

PR #1359, opened 2022-08-19
1551: Improve logging and metrics for BotRunner

erikj79:
I'm trying to better understand bot performance. I'm sure there are lots of things we could improve performance wise, but first we need to figure out how. We already have some logging and metrics gathering, but it's still not providing me with the information I need.

With this change, I'm aiming to gather better metrics and logging around the getPeriodicItems() call, as well as refining the metrics for running WorkItems.

The thread user time measurements from MX beans is useless to me, so I'm removing it. The cpu time is probably also useless, but I'm keeping it for now. What I really want is wall clock time for WorkItems, so I'm adding that. The gauges for measuring any kind of running time aren't useful at all. At least with counters, we can get some kind of idea how much time is spent on each kind of WorkItem over some time frame (using a rate function in the presentation layer).

To complement the timing counters, I'm adding simple WorkItem counters as well, to keep track how many are scheduled, added as pending, submitted and finished.

For the BotRunner.checkPeriodicItems() method, I'm adding two metrics. Since the whole method is running only one at a time, a gauge for tracking how long each round takes does make sense. Then for each sub call to Bot.getPeriodicItems there is a counter that basically gives the breakdown on how much of the time is spent in each kind of bot.

For logging, I'm adding some more log messages to better track which Bot.getPeriodic item is called when, and a LogContext, so that all log messages are tagged with the bot toString, just like WorkItems are already. To support this I had to add toString() methods to all Bot implementations that were missing it.

---------
### Progress

### Issue
 * [SKARA-1551](https://bugs.openjdk.org/browse/SKARA-1551): Improve logging and metrics for BotRunner

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1359/head:pull/1359` \
`$ git checkout pull/1359`

`$ git checkout pull/1359` \
`$ git pull https://git.openjdk.org/skara pull/1359/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1359`

`$ git pr show -t 1359`

</details>

<a href="https://git.openjdk.org/skara/pull/1359.diff">https://git.openjdk.org/skara/pull/1359.diff</a>

</details>

### DIFF
--- a/bot/src/main/java/org/openjdk/skara/bot/BotRunner.java
+++ b/bot/src/main/java/org/openjdk/skara/bot/BotRunner.java
@@ -99,6 +99,8 @@ public class BotRunner {
                 Gauge.name("skara_runner_submitted_time").labels("bot", "work_item").register();
         private static final Counter.WithTwoLabels TIME_COUNTER =
                 Counter.name("skara_runner_run_time_total").labels("bot", "work_item").register();
+        private static final Counter.WithTwoLabels ITEM_FINISHED_COUNTER =
+                Counter.name("skara_runner_finished").labels("bot", "work_item").register();
         private static final Counter.WithTwoLabels CPU_TIME_COUNTER =
                 Counter.name("skara_runner_cpu_time_total").labels("bot", "work_item").register();
         private static final Counter.WithTwoLabels ALLOCATED_BYTES_COUNTER =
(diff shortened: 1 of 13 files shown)


kevinrushforth (approved):
Changes looks reasonable.

PR #1360, opened 2022-08-20
1431: The method TestPullRequest#diff sometimes returns wrong information

lgxbslgx:
Hi all,

The method `TestPullRequest#diff` always passes the head hash of the target branch to the method `Repository#diff` as the base hash. But it may be wrong when the target branch was submitted some new commits after creating the pull request. `TestPullRequest#diff` should find the last common hash of the source and target branch as the base hash.

This patch fixes the bug and adds the test case. I don't know whether it is good to place the test case at the class `CSRBotTests`, but the method `CSRBotTests#testBackportCsr` indeed contains the situation of this bug.

Best Regards,
-- Guoxiong

---------
### Progress

### Issue
 * [SKARA-1431](https://bugs.openjdk.org/browse/SKARA-1431): The method TestPullRequest#diff sometimes returns wrong information

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1360/head:pull/1360` \
`$ git checkout pull/1360`

`$ git checkout pull/1360` \
`$ git pull https://git.openjdk.org/skara pull/1360/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1360`

`$ git pr show -t 1360`

</details>

<a href="https://git.openjdk.org/skara/pull/1360.diff">https://git.openjdk.org/skara/pull/1360.diff</a>

</details>

### DIFF
--- a/test/src/main/java/org/openjdk/skara/test/TestPullRequest.java
+++ b/test/src/main/java/org/openjdk/skara/test/TestPullRequest.java
@@ -252,14 +252,7 @@ public class TestPullRequest extends TestIssue implements PullRequest {
                 sourceHash = targetLocalRepository.fetch(sourceUri, sourceRef);
             }
             // Find the base hash of the source and target branches.
-            var baseHash = sourceHash;
-            var targetBranch = new Branch(targetRef());
-            while (!"0".repeat(40).equals(baseHash.hex())) {
-                if (targetLocalRepository.contains(targetBranch, baseHash)) {
-                    break;
-                }
-                baseHash = targetRepository.commit(baseHash).get().parents().get(0);
-            }
+            var baseHash = targetLocalRepository.mergeBase(sourceHash, targetRepository.branchHash(targetRef()));
             return targetLocalRepository.diff(baseHash, sourceHash);
         } catch (IOException e) {
             throw new UncheckedIOException(e);
--- a/test/src/main/java/org/openjdk/skara/test/TestPullRequest.java
+++ b/test/src/main/java/org/openjdk/skara/test/TestPullRequest.java
@@ -26,6 +26,7 @@ import org.openjdk.skara.forge.*;
 import org.openjdk.skara.host.*;
 import org.openjdk.skara.issuetracker.IssueProject;
 import org.openjdk.skara.network.URIBuilder;
+import org.openjdk.skara.vcs.Branch;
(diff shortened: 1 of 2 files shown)


test/src/main/java/org/openjdk/skara/test/TestPullRequest.java
```diff
@@ -242,17 +243,24 @@ public URI headUrl() {
...
+            var sourceHash = headHash();
+            if (!targetLocalRepository.root().equals(sourceLocalRepository.root())) {
+                // The target and source repo are not same, fetch the source branch
+                var sourceUri = URI.create("file://" + sourceLocalRepository.root().toString());
+                sourceHash = targetLocalRepository.fetch(sourceUri, sourceRef);
             }
+            // Find the base hash of the source and target branches.
+            var baseHash = sourceHash;
+            var targetBranch = new Branch(targetRef());
+            while (!"0".repeat(40).equals(baseHash.hex())) {
```
magicus, line 257: So if the common commit is more than 40 commits away, return "no diff"? :-)

This is only used in testing, right? If so, this you could get away with it, but I think a comment describing the limitations would be needed.

erikj79: I think you are misunderstanding the code. `"0".repeat(40)` just generates a string with 40 zeros, basically a null hash.

Looping to find the common ancestor is one way of doing it, but I would recommend checking out `git merge-base`. There is a method for it already on `GitRepository`.

magicus: I see. Yes, I misunderstood the code as looping backwards for a maximum of 40. Don't know how I got to that conclusion. My brain is apparently still partially stuck in vacation mode. I don't think I'm used to methods being called on a string literal.

lgxbslgx: > Looping to find the common ancestor is one way of doing it, but I would recommend checking out `git merge-base`. There is a method for it already on `GitRepository`.

Thanks for reminding me. I change the code to use the method `mergeBase` just now.

magicus (approved):
Much better.

erikj79 (approved):

lgxbslgx:
Thanks for the review. Integrating.

/integrate

PR #1361, opened 2022-08-23
1552: Log relevant timer numbers

erikj79:
This is a pretty big patch, but it shouldn't be too scary as it's only supposed to change logging, mainly by adding more logging (and also removing some now redundant log calls).

The main focus here is to add structured timing data to some log messages. To do this, I'm using the optional `parameters` `Object[]` that you can supply to the logger from `java.util.logging`. If the LogstashHandler finds a `Duration` instance in that array, it will add it as a "duration" field in the logstash log document. In the BotRunner, we were already using parameters for a different purpose, so it got a little bit messy, but I think it's ok.

There are two kinds of timings I'm adding logging for. The obvious one is timing external HTTP calls and running processes, as well as total running/pending/scheduled times for WorkItems and getPeriodicItems in the BotRunner. The other kind are more complex as I'm trying to measure the user perceived latency from when a PR or bug is updated to when the bot actually reacts with a response to that update. This is of course tricky to get right and needs to be carefully implemented for each instance, but I believe I have at least added some reasonably relevant approximations that will make sense to measure and track. To generate these, I had to introduce a new field in various WorkItems that stores the timestamp from the external change that triggered it (e.g. `prUpdatedAt`).

In addition to this, there are a ton more regular log messages, mostly on info level, that detail actions performed by various WorkItems. This is something I have been missing a lot when browsing logs.

---------
### Progress

### Issue
 * [SKARA-1552](https://bugs.openjdk.org/browse/SKARA-1552): Log relevant timer numbers

### Reviewers
 * [Magnus Ihse Bursie](https://openjdk.org/census#ihse) (@magicus - **Reviewer**) ⚠️ Review applies to [2cb0aee7](https://git.openjdk.org/skara/pull/1361/files/2cb0aee7088042572d1c4c764d93b544ad1a6d8d)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1361/head:pull/1361` \
`$ git checkout pull/1361`

`$ git checkout pull/1361` \
`$ git pull https://git.openjdk.org/skara pull/1361/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1361`

`$ git pr show -t 1361`

</details>

<a href="https://git.openjdk.org/skara/pull/1361.diff">https://git.openjdk.org/skara/pull/1361.diff</a>

</details>

### DIFF
--- a/bot/src/main/java/org/openjdk/skara/bot/BotRunner.java
+++ b/bot/src/main/java/org/openjdk/skara/bot/BotRunner.java
@@ -233,7 +233,7 @@ public class BotRunner {
                     item.handleRuntimeException(e);
                 } finally {
                     var duration = Duration.between(start, Instant.now());
-                    log.log(Level.FINE, "Item " + item + " is now done " + duration,
+                    log.log(Level.FINE, "Item " + item + " is now done after " + duration,
                             new Object[]{TaskPhases.END, duration});
                 }
                 if (followUpItems != null) {
--- a/bot/src/main/java/org/openjdk/skara/bot/BotRunner.java
+++ b/bot/src/main/java/org/openjdk/skara/bot/BotRunner.java
@@ -205,12 +205,14 @@ public class BotRunner {
             }
 
             Collection<WorkItem> followUpItems = null;
+            var start = Instant.now();
             try (var __ = new LogContext(Map.of("work_item", item.toString(),
                     "work_id", String.valueOf(workId)))) {
-                var submittedDuration = Duration.between(createTime, Instant.now());
+                var submittedDuration = Duration.between(createTime, start);
                 SUBMITTED_TIME_GAUGE.labels(item.botName(), item.workItemName()).set(submittedDuration.toMillis() / 1_000.0);
                 log.log(Level.FINE, "Executing item " + item + " on repository " + scratchPath
-                        + " after being submitted for " + submittedDuration, TaskPhases.BEGIN);
+                        + " after being submitted for " + submittedDuration,
+                        new Object[]{TaskPhases.BEGIN, submittedDuration});
                 try {
                     followUpItems = item.run(scratchPath);
                 } catch (UncheckedRestException e) {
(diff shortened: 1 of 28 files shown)


bot/src/main/java/org/openjdk/skara/bot/BotRunner.java
```diff
@@ -230,43 +232,45 @@ private void runMeasured() {
                     }
                     item.handleRuntimeException(e);
                 } finally {
-                    log.log(Level.FINE, "Item " + item + " is now done", TaskPhases.END);
+                    var duration = Duration.between(start, Instant.now());
+                    log.log(Level.FINE, "Item " + item + " is now done " + duration,
```
magicus, line 236:
```suggestion
                    log.log(Level.FINE, "Item " + item + " is now done after " + duration,
```

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -337,20 +338,27 @@ public Collection<WorkItem> prRun(Path scratchPath) {
...
+                return List.of(new CheckWorkItem(bot, prId, errorHandler, prUpdatedAt));
             }
 
             try {
                 Repository localRepo = materializeLocalRepo(scratchPath, hostedRepositoryPool);
 
                 var expiresAt = CheckRun.execute(this, pr, localRepo, comments, allReviews, activeReviews, labels, census, bot.ignoreStaleReviews(), bot.integrators());
+                if (log.isLoggable(Level.INFO)) {
+                    var updatedPr = bot.repo().pullRequest(prId);
+                    logLatency("Time from PR updated to CheckRun done ", updatedPr.updatedAt(), log);
```
magicus, line 360: logLatency only logs at INFO level anyway. I'm not sure why you put just this one instance inside the log.isLoggable guard.

magicus: Actually, I'm trying to understand why this is correct. `logLatency` looks like this:
```
    protected void logLatency(String message, ZonedDateTime endTime, Logger log) {
        var latency = Duration.between(prUpdatedAt, endTime);
...
```

I would have expected the endTime to be the current time. But maybe I'm just getting lost here on the difference between `updatedPr.updatedAt()` and `prUpdatedAt`.

magicus (approved):
Overall looks good to me.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
erikj79: We basically always log INFO so the guard isn't strictly necessary. Even so I tried to put guards whenever I needed to fetch data from a remote to generate the log message. In this case, it's refetching the PullRequest from the forge.

I should add a comment about this to make it clearer. Whenever possible, I want the latency numbers to be calculated between timestamps from the same clock. In this case the clock is on the forge (e.g. GitHub). The start time is the updatedAt timestamp read from the PR when we triggered this WorkItem. The endTime is the updatedAt from the PR just now, after we updated it from this WorkItem (CheckRun.execute() will always update the `check` in the PR).

PR #1362, opened 2022-08-24
1560: Backport command fails to load census

erikj79:
This patch attempts to fix the problem where a commit command (in our case a /backport command) fails to instantiate a CensusInstance, due to the configured jcheck "project", at the particular commit, not existing in the census in question. What happens then is that CensusInstance.create fails, which triggers an endless series of retries in the bot.

Since none of the current commit commands have a need for the "project" part of the census, I decided to create a separate CensusInstance type `LimitedCensusInstance` without the "project", and explicitly use this in the commit commands.

Tests are passing, and I could manually verify that this solves the problem we currently have with a bad /backport command in a private GitLab instance.

---------
### Progress

### Issue
 * [SKARA-1560](https://bugs.openjdk.org/browse/SKARA-1560): Backport command fails to load census

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1362/head:pull/1362` \
`$ git checkout pull/1362`

`$ git checkout pull/1362` \
`$ git pull https://git.openjdk.org/skara pull/1362/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1362`

`$ git pr show -t 1362`

</details>

<a href="https://git.openjdk.org/skara/pull/1362.diff">https://git.openjdk.org/skara/pull/1362.diff</a>

</details>

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CensusInstance.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CensusInstance.java
@@ -33,17 +33,12 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.util.Optional;
 
-class CensusInstance {
-    private final Census census;
-    private final JCheckConfiguration configuration;
+class CensusInstance extends LimitedCensusInstance {
     private final Project project;
-    private final Namespace namespace;
 
     private CensusInstance(Census census, JCheckConfiguration configuration, Project project, Namespace namespace) {
-        this.census = census;
-        this.configuration = configuration;
+        super(census, configuration, namespace);
         this.project = project;
-        this.namespace = namespace;
     }
 
     private static Project project(JCheckConfiguration configuration, Census census) {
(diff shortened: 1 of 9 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/CensusInstance.java
```diff
@@ -158,3 +161,33 @@ boolean isReviewer(HostUser hostUser) {
         return project.isReviewer(contributor.username(), version);
     }
 }
+
+class LimitedCensusInstance {
```
kevinrushforth, line 165: Maybe move this to its own file? I try to avoid having more than one top level class in the same file, but do whatever makes the most sense (maybe Skara uses this in other places?).

kevinrushforth (approved):
Looks good.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CensusInstance.java
erikj79: I will think about it. This is a rather common pattern in Skara, which is why I went for it.

kevinrushforth (approved):
I think this is a cleaner separation. Looks good.

lgxbslgx (approved):
LGTM

erikj79:
Thanks!
/integrate

PR #1365, opened 2022-09-01
1568: Dependency PR Resolution Detection Does Not Check Means of Resolution

lgxbslgx:
Hi all,

Currently, the closed but not integrated PR causes the dependent PRs to be retargeted. It is not right. This patch fixes it and changes the test case.

Best Regards,
-- Guoxiong

---------
### Progress

### Issue
 * [SKARA-1568](https://bugs.openjdk.org/browse/SKARA-1568): Dependency PR Resolution Detection Does Not Check Means of Resolution

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1365/head:pull/1365` \
`$ git checkout pull/1365`

`$ git checkout pull/1365` \
`$ git pull https://git.openjdk.org/skara pull/1365/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1365`

`$ git pr show -t 1365`

</details>

<a href="https://git.openjdk.org/skara/pull/1365.diff">https://git.openjdk.org/skara/pull/1365.diff</a>

</details>

### DIFF
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/prbranch/PullRequestBranchNotifier.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/prbranch/PullRequestBranchNotifier.java
@@ -114,11 +114,11 @@ public class PullRequestBranchNotifier implements Notifier, PullRequestListener 
                 for (var retargeted : retargetedDependencies) {
                     log.info("Posting retargeted comment on PR " + pr.id());
                     retargeted.addComment("""
-                            The parent pull request that this pull request depends on has been closed without being integrated. \
-                            This has caused the pull request branch pr/XX to be deleted and the target branch of this pull \
-                            request to be updated, which means that changes from the parent pull request will start to show up \
-                            in this pull request. If closing the parent pull request was done in error, it will need to be \
-                            re-opened and this pull request will need to be retargeted again manually.
+                            The parent pull request that this pull request depends on has been closed without being \
+                            integrated and the target branch of this pull request has been updated as the previous \
+                            branch was deleted. This means that changes from the parent pull request will start to \
+                            show up in this pull request. If closing the parent pull request was done in error, it will \
+                            need to be re-opened and this pull request will need to manually be retargeted again.
                             """);
                 }
             }
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/prbranch/PullRequestBranchNotifier.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/prbranch/PullRequestBranchNotifier.java
@@ -86,16 +86,17 @@ public class PullRequestBranchNotifier implements Notifier, PullRequestListener 
     @Override
     public void onStateChange(PullRequest pr, Path scratchPath, Issue.State oldState) {
         if (pr.state() == Issue.State.CLOSED) {
+            var retargetedDependencies = PreIntegrations.retargetDependencies(pr);
             deleteBranch(pr);
             if (pr.labelNames().contains("integrated")) {
-                var retargetedDependencies = PreIntegrations.retargetDependencies(pr);
                 for (var retargeted : retargetedDependencies) {
                     log.info("Posting retargeted comment on PR " + pr.id());
                     retargeted.addComment("""
-                            The dependent pull request has now been integrated, and the target branch of this pull request \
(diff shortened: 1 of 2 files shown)


erikj79 (commented):
I don't think this really solves the issue. The pr branch for the closed PR is still deleted, so the dependent PR will still get retargeted, but without any notification to the user. I think we need to post a different message in the dependent PRs, explaining that the PR this PR depends on was closed without being integrated.

Reading the current message, the word 'dependent' isn't used correctly.

lgxbslgx:
> I don't think this really solves the issue. The pr branch for the closed PR is still deleted, so the dependent PR will still get retargeted, but without any notification to the user.

I agree that this patch needs to be adjusted. At least the branch can't be deleted before the successor pull request retargeting and notifing.

> I think we need to post a different message in the dependent PRs, explaining that the PR this PR depends on was closed without being integrated.

Posting another message is an available way. But it can't actually solve all the problem. There are several situations we may meet (some but not all):

- The dependent pull request is closed because of timeout.

At this situation, we may need the author to re-open the pull request and continue the work. And target branch of the successor pull request should be adjusted to this dependent pull request again.

- The dependent pull request is closed intendedly because the author want the change to be implemented in the successor pull request (two patch reduce to one patch).

At this situation, the bot only need to retarget the branch, just like your suggestion.

-  The dependent pull request is closed because the another better pull request is submitted.

At this situation, the bot now don't know which branch the successor pull request should retarget to.

- The dependent pull request is closed because its feature is no longer needed.

At this situation, we don't know the succesor pull request how to change.

The bot can't handle all the situations even though we can list all the situations. There may be two root reasons:
1. The bot or we (the developers of the bot) don't know why the pull request is closed in some situations. It is complicated.
2. The bot currently doesn't have the ability to retarget the branch to the original branch (But this can be implemented).

erikj79:
I agree that all of those are possible situations, and I also agree that the Skara bot can't reasonably figure out which one is correct. Because of that, I think the best we can reasonably do is post a comment explaining what happened and tell the user what to do. I also think we have to still delete the branch and retarget the dependent pull request. 

Also please note that "dependent" should refer to the PR that depends on another PR. Example PR A has a PR B depending on it. That means the target of PR B is the branch pr/A. In this situation, PR B is the dependent of PR A. The current message has this mixed up. If we need to refer to PR A, we could perhaps use the word "parent".

Here is my take on a message:

The parent pull request that this pull request depend on has been closed without being integrated. This has caused the pull request branch `pr/XX` to be deleted and the target branch of this pull request to be updated, which means that changes from the parent pull request will start to show up in this pull request. If closing the parent pull request was done in error, it will need to be re-opened and this pull request will need to be retargeted again manually.

lgxbslgx:
It is good to need more manual work in the related situations. I updated the code just as you suggested.

bots/notify/src/main/java/org/openjdk/skara/bots/notify/prbranch/PullRequestBranchNotifier.java
```diff
@@ -88,24 +88,39 @@ public void onStateChange(PullRequest pr, Path scratchPath, Issue.State oldState
...
+                }
+            } else {
+                for (var retargeted : retargetedDependencies) {
+                    log.info("Posting retargeted comment on PR " + pr.id());
+                    retargeted.addComment("""
+                            The parent pull request that this pull request depends on has been closed without being integrated. \
+                            This has caused the pull request branch pr/XX to be deleted and the target branch of this pull \
+                            request to be updated, which means that changes from the parent pull request will start to show up \
+                            in this pull request. If closing the parent pull request was done in error, it will need to be \
+                            re-opened and this pull request will need to be retargeted again manually.
```
erikj79, line 121: Reading what I wrote again, I want to revise it a bit.
```suggestion
                            The parent pull request that this pull request depends on has been closed without being \ 
                            integrated and the target branch of this pull request has been updated as the previous \
                            branch was deleted. This means that changes from the parent pull request will start to \
                            show up in this pull request. If closing the parent pull request was done in error, it will \
                            need to be re-opened and this pull request will need to manually be retargeted again.
```

lgxbslgx: Fixed just now.

erikj79 (approved):

PR #1370, opened 2022-09-08
1589: Improve TestIssue and TestPullRequest

erikj79:
While working on [SKARA-1565](https://bugs.openjdk.org/browse/SKARA-1565) I've realized that the current test implementations of `Issue` and `PullRequest` (TestIssue and TestPullRequest) do not represent the behavior of their real implementation counterparts very well. For [SKARA-1565](https://bugs.openjdk.org/browse/SKARA-1565), this discrepancy in behavior makes it impossible to write new tests for the poller functionality and at the same time keeping existing tests for any specific bot working when switching the bot to using the poller. Fixing the tests for this turned out to be a pretty large change, so I'm moving it to a separate bug to make reviews easier.

The main issue here is the difference in lifecycle between instances of the test classes and the real classes. In production, when querying an `IssueProject` for an Issue, what you get is basically an object instance which contains a snapshot of the external server side data for the issue. When you do another query, you get a new instance with potentially more up to date data. Some data is not included in the snapshot and will be dynamically fetched from the remote server on request, e.g. comments. Any method on the production implementation of Issue that adds or modifies data calls through to the external server and does not update the current snapshot.

In contrast to the above behavior, the `TestIssue`, when returned from the `TestIssueProject`, will be a new instance each time, but that instance shares most of the underlying data directly with all other instances of the "same" issue. So most additions or updates are immediately reflected in all instances of the "same" issue. For the poller tests, this makes it impossible to simulate having different snapshots returned at different times, and comparing the contents of them to figure out if something has changed. I'm also worried that we risk inadvertently building in assumptions in our bot code that relies on this behavior.

I'm fixing this by changing `TestIssue` and `TestPullRequest` to better mimic the behavior of the production implementations. To achieve this, I'm introducing two (new) classes TestIssueStore and `TestPullRequestStore`. These classes represent the server side state of these objects. The `TestIssue` and `TestPullRequest` are changed to take similar snapshots as the production classes, and any method that modifies the state will just call through and modify the `*Store` instance. For test code to have an easier time inspecting "server side" data, the `*Store` objects are exposed to test code.

The bulk of the changes are updates to existing test code, mostly to use the new `TestPullRequest::store` method to verify state in a pull request. Another notable change is that `HostCredentials` class (which is where test code generally create new `PullRequest` and `Issue` instances, now returns `TestPullRequest` and `TestIssue` which makes it easy for tests to get hold of the backing `*Store` instances. I'm also removing all instances of this construct:

```
            // Make sure that the push registered
            var lastHeadHash = pr.headHash();
            var refreshCount = 0;
            do {
                pr = author.pullRequest(pr.id());
                if (refreshCount++ > 100) {
                    fail("The PR did not update after the new push");
                }
            } while (pr.headHash().equals(lastHeadHash));
```
I don't think it was needed before this patch, and I'm fairly certain it's not needed now. The `headHash` field of a TestPullRequest is fetched from the underlying repository on instance creation. 

This is a test only change not touching any product code.

---------
### Progress

### Issue
 * [SKARA-1589](https://bugs.openjdk.org/browse/SKARA-1589): Improve TestIssue and TestPullRequest

### Reviewers
 * [Magnus Ihse Bursie](https://openjdk.org/census#ihse) (@magicus - **Reviewer**) ⚠️ Review applies to [4387c8c0](https://git.openjdk.org/skara/pull/1370/files/4387c8c0d31f3056c3fa90ace4c0e273c29e4f8e)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1370/head:pull/1370` \
`$ git checkout pull/1370`

`$ git checkout pull/1370` \
`$ git pull https://git.openjdk.org/skara pull/1370/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1370`

`$ git pr show -t 1370`

</details>

<a href="https://git.openjdk.org/skara/pull/1370.diff">https://git.openjdk.org/skara/pull/1370.diff</a>

</details>


bots/pr/src/test/java/org/openjdk/skara/bots/pr/CheckTests.java
```diff
@@ -203,63 +192,49 @@ void multipleReviews(TestInfo testInfo) throws IOException {
...
-            do {
-                authorPr = author.pullRequest(authorPr.id());
-                if (refreshCount++ > 100) {
-                    fail("The PR did not update after the new push");
-                }
-            } while (authorPr.headHash().equals(lastHeadHash));
-
             // Check that the review is flagged as stale
             TestBotRunner.runPeriodicItems(checkBot);
-            authorPr = author.pullRequest(authorPr.id());
```
magicus, line 233: What was this line even supposed to accomplish? Some kind of reload?

issuetracker/src/test/java/org/openjdk/skara/issuetracker/IssueTrackerTests.java
```diff
@@ -127,7 +127,8 @@ void addIssueLink(TestInfo info) throws IOException {
 
             var links = issue2.links();
             assertEquals(1, links.size());
-            assertEquals(link, links.get(0));
+            assertEquals(link.relationship(), links.get(0).relationship());
```
magicus, line 130: Here's an obvious drawback of the "var" syntax. What is this `links`? Why did not an ordinary equals suffice? Does this really relate to the rest of the changes, or is it an independent fix that happened to tag along?

magicus (approved):
This is quite a massive PR. It took some time to look thought. I lost track of some changes but I think I got to the bottom of most of these.  I can definitely see a drawback with using the "var" syntax when reviewing diffs on Github -- when I'm in the IDE I can get tooltips telling  me about the type, but now I had to jump back and forth to look for definitions to understand how the code tied together. 

So in the end, I think it looks good. The fact that it is only test changes makes it definitely less risky.

bots/pr/src/test/java/org/openjdk/skara/bots/pr/CheckTests.java
erikj79: I'm happy to try to clarify. With the old `TestPullRequest` implementation, some data was part of a snapshot and some wasn't, so fetching the PR again caused a reload and was necessary to see some kinds of changes that a bot would make. This is even more true with the new implementation, but since we have an explicit object representing the server side state of the PR, I think it's better if test code uses that explicitly to inspect state.

issuetracker/src/test/java/org/openjdk/skara/issuetracker/IssueTrackerTests.java
erikj79: Yes, in modern intellij, I get grey type names for every `var` variable, which makes it really convenient. I too have to open most PRs I review in the IDE to properly read the code.

Regarding why equals won't suffice. Before this patch, `TestIssue::links` would just return a list of all `Link` objects that had been added to the Issue. In this patch, I had to change it to always refresh the target `Issue` of each link with a fresh snapshot, because that's what happens when you query for any real links. Another way to solve this would be to implement `Link::equals` to take this into account. That would probably be a good thing to do, but then this patch will touch product code. Long term, that's probably better and I can do that if you want.

erikj79: Looking closer at this, the `Link::equals` already does the right thing. The reason it's not sufficient is that `issue1` is being modified by setting a body. `TestIssue::setBody` just calls through to the `TestIssueStore` object. The snapshot in `issue1` doesn't change. When the link is created it points to this snapshot. When `issue2.links()` is called, the new logic will return a new `Link` object pointing to a fresh snapshot of `issue1`. If we just ran `equals` to compare `link` with `links.get(0)`, then it would return false, because `link` points to `issue1` without a body, while `links.get(0)` points to a fresh snapshot with a body.

Checking that `Issue::id` is equal is the correct thing to do here. Links point to an issue identified by an ID, not to a specific snapshot of the issue.

erikj79:
> This is quite a massive PR. It took some time to look thought. I lost track of some changes but I think I got to the bottom of most of these. I can definitely see a drawback with using the "var" syntax when reviewing diffs on Github -- when I'm in the IDE I can get tooltips telling me about the type, but now I had to jump back and forth to look for definitions to understand how the code tied together.
> 
> So in the end, I think it looks good. The fact that it is only test changes makes it definitely less risky.

Thank you for taking the time!

erikj79:
I seem to have an intermittent failure in BackportsTest on Windows. Still trying to figure it out, so will hold off on this for now.

erikj79:
> I seem to have an intermittent failure in BackportsTest on Windows. Still trying to figure it out, so will hold off on this for now.

I believe I found the issue. In `TestIssue::equals` (and hashCode) I forgot to include the `id` field, so instances with different IDs but the same contents were considered the same, which would randomly confuse tests. If all GHA checks shows green, I think we are good to go.

PR #1372, opened 2022-09-12
1595: Introduce an IssuePoller similar to the PullRequestPoller

erikj79:
Inspired by the `PullRequestPoller`, I have now also created an `IssuePoller`. The requirements are a bit different on this poller, which makes it a bit easier to implement. 

Like the `PullRequestPoller`, it needs to handle retries, something I had overlooked in the original `CSRIssueBot`. Compared to that implementation, I'm also moving away from using `IssueProject::lastUpdatedIssue` for the initial query. The problem with relying on such a query for the very first round is if it, or something else in that round, fails. At least for the `CSRIssueBot`, once we are in the second round, after a failed first round, it's no longer safe to only query for the very last updated Issue, there could be multiple updates that we would miss. Instead there is a configurable `Duration` that defines how far back we go from the time the bot was created for the first query. Depending on what the bot needs, this Duration can be expected to cover any downtime for the bot (typically in the order of days), which would be the case for the `SyncLabelBot` bot, or it could just be needed to cover any time sync differences between the bot and the server (typically in the order of minutes to an hour), which would apply to the `CSRIssueBot`.

Like with my revised PR for SKARA-1565, I'm not changing any bot to use this poller yet. I'm going to followup with that change for the CSRBot in SKARA-1594, and for other bots later.

---------
### Progress

### Issue
 * [SKARA-1595](https://bugs.openjdk.org/browse/SKARA-1595): Introduce an IssuePoller similar to the PullRequestPoller

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1372/head:pull/1372` \
`$ git checkout pull/1372`

`$ git checkout pull/1372` \
`$ git pull https://git.openjdk.org/skara pull/1372/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1372`

`$ git pr show -t 1372`

</details>

<a href="https://git.openjdk.org/skara/pull/1372.diff">https://git.openjdk.org/skara/pull/1372.diff</a>

</details>

### DIFF
--- a/issuetracker/build.gradle
+++ b/issuetracker/build.gradle
@@ -26,6 +26,7 @@ module {
     test {
         requires 'org.openjdk.skara.test'
         requires 'org.junit.jupiter.api'
+        requires 'org.openjdk.skara.forge'
         requires 'jdk.httpserver'
         opens 'org.openjdk.skara.issuetracker' to 'org.junit.platform.commons'
         opens 'org.openjdk.skara.issuetracker.jira' to 'org.junit.platform.commons'
@@ -42,6 +43,7 @@ dependencies {
     implementation project(':host')
 
     testImplementation project(':test')
+    testImplementation project(':forge')
 }
 
 publishing {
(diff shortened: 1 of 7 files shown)


magicus (approved):
LGTM

erikj79:
I integrated this, but I remember now that this change does depend on https://github.com/skara/pull/1369 for correct JSON equals. This is mainly a note to not try to start using the IssuePoller before PR is integrated.

PR #1376, opened 2022-09-15
1602: Fix labels handling in GitLabMergeRequest

erikj79:
A PullRequest exports two methods for getting all labels: `labels()` and `labelNames()`. The first returns `List<Label>` and the latter returns `List<String>`. The labelNames method has a default implementation that calls labels() and maps them to the name field. This is ok for most implementations, but not for GitLabMergeRequest. 

In GitLabMergeRequest, only the names of the labels are available from the merge_request REST object itself. In order to convert these to Label objects, it needs to make another REST call on the "project" (repository) to get the full label data. This extra call can add up to quite significant amounts of time when processing multiple PullRequest instances.

To make matters worse, the labels in GitLabMergeRequest are initialized in the constructor, so for every GitLabMergeRequest instance created, an extra REST call is made. There is also just a single use of PullRequest::labels() in the whole code base, and that is in the default implementation for labelNames(), so caller is ever interested in the full Label objects. This means that every instance of GitLabMergeRequest is making this extra REST call to initialize a cached set of data that is never used.

I'm solving this by only caching the label names instead of full Label objects. I've tried to retain the semantics of each method that deal with labels. The add/remove/set methods have been improved to update the cache instead of just clearing it.

---------
### Progress

### Issue
 * [SKARA-1602](https://bugs.openjdk.org/browse/SKARA-1602): Fix labels handling in GitLabMergeRequest

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1376/head:pull/1376` \
`$ git checkout pull/1376`

`$ git checkout pull/1376` \
`$ git pull https://git.openjdk.org/skara pull/1376/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1376`

`$ git pr show -t 1376`

</details>

<a href="https://git.openjdk.org/skara/pull/1376.diff">https://git.openjdk.org/skara/pull/1376.diff</a>

</details>

### DIFF
--- a/forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabMergeRequest.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabMergeRequest.java
@@ -44,7 +44,9 @@ public class GitLabMergeRequest implements PullRequest {
     private final GitLabRepository repository;
     private final GitLabHost host;
 
-    private List<Label> labels;
+    // Only cache the label names as those are most commonly used and converting to
+    // Label objects is expensive.
+    private List<String> labels;
 
     GitLabMergeRequest(GitLabRepository repository, GitLabHost host, JSONValue jsonValue, RestRequest request) {
         this.repository = repository;
@@ -52,9 +54,8 @@ public class GitLabMergeRequest implements PullRequest {
         this.json = jsonValue;
         this.request = request.restrict("merge_requests/" + json.get("iid").toString() + "/");
 
-        labels = json.get("labels")
-                     .stream()
-                     .map(s -> labelNameToLabel(s.asString()))
+        labels = json.get("labels").stream()
+                     .map(JSONValue::asString)
                      .sorted()
                      .collect(Collectors.toList());
     }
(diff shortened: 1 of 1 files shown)


forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabMergeRequest.java
```diff
@@ -646,28 +646,28 @@ public void removeLabel(String label) {
...
-                                // Avoid throwing NPE for unknown labels
-                                .filter(Objects::nonNull)
-                                .sorted()
-                                .collect(Collectors.toList());
-        }
+        return labels.stream()
+                .map(this::labelNameToLabel)
+                // Avoid throwing NPE for unknown labels
+                .filter(Objects::nonNull)
+                .toList();
```
magicus, line 666: You're losing `.sorted()`. Maybe it doesn't matter?

magicus (approved):
Looks good.

forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabMergeRequest.java
erikj79: Hm, probably shouldn't matter, but I retained the sorted property on the `labels` field in the constructor, so should probably make sure all assignments to that field are made with sorted lists for consistency.

magicus: Sounds good. In general I think it’s best to keep things sorted, in order to minimize spurious changes in places that might end up being user visible.

magicus (approved):
Looks good

PR #1380, opened 2022-09-22
1606: PullRequestPoller always processes the last updated MR for GitLab

erikj79:
The new PullRequestPoller from [SKARA-1565](https://bugs.openjdk.org/browse/SKARA-1565) is inefficient in combination with GitLab. In the most common case, when no MRs have been updated, you would expect it to only do one GET query and be done. Instead, that initial query will always return the last updated MR, and then continue to fetch all the metadata for it, which adds up to a lot of time. 

This is made even worse by having to fetch comments and reviews for every MR to be able to compare them for potential updates. In GitLab, both comments and reviews are stored in the same set of "notes" for a merge request. Since we are only fetching them for comparing between different snapshots, we can make this faster by just fetching all notes once (which will also increase accuracy as there are other kinds of notes too, which we probably should check for updates too).

The repeated refetching of the last updated MR is basically the same problem as the IssuePoller is already solving. The resolution for timestamp based queries is limited (1s on GitLab) and the query is inclusive. In contrast, the current implementation for GitLabRepository treats the timestamp as exclusive.

This patch does the following:

1. Change all timestamp based pull request and issue queries (including for test implementations) to be inclusive, for consistency.
2. Implement the same kind of "positive" padding to the updatedAfter parameter in PullRequestPoller as is already present in IssuePoller. I'm also modifying tests to actually verify this behavior by exposing some of the internal data of the poller to the test.
3. Add `PullRequest::comparisonSnapshot` which returns an `Object` that represents all data that needs to be considered when evaluating if a snapshot is equal to another. In GitLabMergeRequest, this is cached lazily to avoid repeated remote calls.

While working on this, I've realized that IssuePoller and PullRequestPoller are basically doing the exact same thing now, with pretty small variations. Ideally we should combine these into a common super class, but I'm leaving that for a separate change to make the current changes easier to follow.

I'm going to run this change in "staging" for a bit to see how it behaves.

---------
### Progress

### Issue
 * [SKARA-1606](https://bugs.openjdk.org/browse/SKARA-1606): PullRequestPoller always processes the last updated MR for GitLab

### Reviewers
 * [Erik Helin](https://openjdk.org/census#ehelin) (@edvbld - **Reviewer**) ⚠️ Review applies to [b9092aee](https://git.openjdk.org/skara/pull/1380/files/b9092aee8618ff2b5b08010396d3fb34da250337)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1380/head:pull/1380` \
`$ git checkout pull/1380`

`$ git checkout pull/1380` \
`$ git pull https://git.openjdk.org/skara pull/1380/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1380`

`$ git pr show -t 1380`

</details>

<a href="https://git.openjdk.org/skara/pull/1380.diff">https://git.openjdk.org/skara/pull/1380.diff</a>

</details>

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestBot.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestBot.java
@@ -102,7 +102,7 @@ class PullRequestBot implements Bot {
         this.enableJep = enableJep;
 
         autoLabelled = new HashSet<>();
-        poller = new PullRequestPoller(repo, true, true, true);
+        poller = new PullRequestPoller(repo, true);
 
         // Only check recently updated when starting up to avoid congestion
         lastFullUpdate = Instant.now();
--- a/bots/tester/src/test/java/org/openjdk/skara/bots/tester/InMemoryPullRequest.java
+++ b/bots/tester/src/test/java/org/openjdk/skara/bots/tester/InMemoryPullRequest.java
@@ -339,7 +339,7 @@ class InMemoryPullRequest implements PullRequest {
     }
 
     @Override
-    public Object comparisonSnapshot() {
+    public Object snapshot() {
         return this;
     }
 }
--- a/bots/tester/src/test/java/org/openjdk/skara/bots/tester/InMemoryPullRequest.java
+++ b/bots/tester/src/test/java/org/openjdk/skara/bots/tester/InMemoryPullRequest.java
@@ -337,4 +337,9 @@ class InMemoryPullRequest implements PullRequest {
     public Optional<Hash> findIntegratedCommitHash() {
         return Optional.empty();
     }
+
+    @Override
+    public Object comparisonSnapshot() {
+        return this;
+    }
 }
--- a/forge/src/main/java/org/openjdk/skara/forge/HostedRepository.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/HostedRepository.java
@@ -53,14 +53,16 @@ public interface HostedRepository {
     List<PullRequest> openPullRequests();
 
     /**
-     * Returns a list of all pull requests (both open and closed) that have been updated after the
-     * provided time, ordered by latest updated first. If there are many pull requests that
-     * match, the list may have been truncated.
+     * Returns a list of all pull requests (both open and closed) that have
+     * been updated after or on the given time, with a resolution given by
+     * Host::timeStampQueryPrecision, ordered by latest updated first. If there
+     * are many pull requests that match, the list may have been truncated.
      */
     List<PullRequest> pullRequestsAfter(ZonedDateTime updatedAfter);
 
     /**
-     * Returns a list of all open pull requests that have been updated after the given time.
+     * Returns a list of all open pull requests that have been updated after or on
+     * the given time, with a resolution given by Host::timeStampQueryPrecision.
      */
     List<PullRequest> openPullRequestsAfter(ZonedDateTime updatedAfter);
     List<PullRequest> findPullRequestsWithComment(String author, String body);
--- a/forge/src/main/java/org/openjdk/skara/forge/PullRequest.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/PullRequest.java
@@ -211,5 +211,5 @@ public interface PullRequest extends Issue {
      * Returns an object that represents a complete snapshot of this pull request.
      * Used for detecting if anything has changed between two snapshots.
      */
-    Object comparisonSnapshot();
+    Object snapshot();
 }
--- a/forge/src/main/java/org/openjdk/skara/forge/PullRequest.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/PullRequest.java
@@ -206,4 +206,10 @@ public interface PullRequest extends Issue {
     static String commitHashMessage(Hash hash) {
         return hash != null ? "Pushed as commit " + hash.hex() + "." : "";
     }
+
+    /**
+     * Returns an object that represents a complete snapshot of this pull request.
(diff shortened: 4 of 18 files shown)


erikj79:
Looking at logs, this is now working as intended.

erikj79:
This change is currently blocking deployment of new changes to Skara bots, unless we back out [SKARA-1590](https://bugs.openjdk.org/browse/SKARA-1590). It would certainly be better to get this fix in however.

edvbld:
Hey @erikj79, nice work 🎉 

First of all, I think your changes are fine, the reasoning and the code checks out. I have a bunch of suggestions, but those you can choose whether you want to pursue or not, they do not have to be made within the scope of this PR. Smaller things to consider:

- in `GitLabMergeRequest`, since you are already caching the `comparisonSnapshot`, should we instead cache the `notes` JSON result? There are several methods in `GitLabMergeRequest` that makes use of the `notes` JSON, then they would be able to utilise this cached result as well (saving a number of requests). This thought can of course be extended to other JSON objects that `GitLabMergeRequest` fetches as well.
- very minor nit, but I would have named `comparisonSnapshot` just `snapshot`, but that is up to you 😄 
- more of a question than a comment: the snapshot functionality will mean that we won't even run the bots if a MR hasn't changed, but what is the main cost of running the bots on a MR that hasn't changed? My guess would be that the expensive part are all the requests that the bots have to make in order to figure out that they don't have to do anything, but if we cache more aggressively, would that cost go away? Anyways, just a thought

Now for a bigger experiment 🧑‍🔬 One annoying thing is that the GitLab REST API doesn't return the comments for a MR when fetching a single MR (unlike GitHub which does). So I played around a bit with the GitLab GraphQL API and think I managed to achieve this:

```
$ GRAPHQL_TOKEN=<a personal access token with read-api scope>
$ curl -v 'https://<gitlab-domain>/api/graphql' \
     --header "Authorization: Bearer $GRAPHQL_TOKEN" \
     --header "Content-Type: application/json"
     --request POST \
     --data '{"query": "{ project(fullPath: \"<group-name>/<project-name>\") { mergeRequests(state: opened) { nodes { title notes { nodes { id body  } } } }  } }" }'
```

GraphQL queries can a look a bit messy when typing the on the command-line for CURL, below is the query a bit nicer formatted:

```
{
  "query": 
  "{ project(fullPath: \"<group-name>/<project-name>\") {
        mergeRequests(state: opened) {
          nodes {
            title
              notes {
                nodes { 
                  id 
                  body
                }
            }
          }
        }
      }
   }"
}
```

The query would have to be expanded to get include all the fields we use from the two `GET` calls we do today fetch the main MR information and the MR "notes" (comments). I think that can be done and if it can be, then `GitLabRepository::pullRequests` could be updated to utilise this (a new constructor for `GitLabMergeRequest` would also be needed to store the notes, that ties into my previous suggestion of caching the notes).

With the above, and with your great work on the `PullRequestPoller` (and drilling down into why GitLab doesn't honor the "updatedAfter" query parameter) I _think_ we would get away with doing a single POST to GitLab's GraphQL endpoint when checking for updated MRs (since the notes would be included). So if nothing has changed, then all that is going to happen is a single HTTP request, and we can't optimise that away 😆 

Feel free to ping me offline if you want to chat about GraphQL and these suggestions. Anyhow, this PR is good to go as it is, but there is also room to make further improvements 👍

edvbld (approved):

erikj79:
> First of all, I think your changes are fine, the reasoning and the code checks out. I have a bunch of suggestions, but those you can choose whether you want to pursue or not, they do not have to be made within the scope of this PR. Smaller things to consider:

I also have a lot of ideas for how to further optimize. I'm trying to work iteratively on this so that each change can be easily proven correct (to make reviews easier and to hopefully avoid some mistakes). I also want to evaluate the metrics I have added between changes to make sure we attack the right things. 

> * in `GitLabMergeRequest`, since you are already caching the `comparisonSnapshot`, should we instead cache the `notes` JSON result? There are several methods in `GitLabMergeRequest` that makes use of the `notes` JSON, then they would be able to utilise this cached result as well (saving a number of requests). This thought can of course be extended to other JSON objects that `GitLabMergeRequest` fetches as well.

Maybe, but I'm not sure if every fetch of a GitLabMergeRequest will eventually result in fetching notes. I thin I have been leaning towards a lazy fetch of notes instead. That would still be two queries, or rather an extra query per MR, but still a lot less than we have today.

> * very minor nit, but I would have named `comparisonSnapshot` just `snapshot`, but that is up to you 😄

I picked the name to make it clear that the intended purpose of this snapshot object is only for comparisons. Different implementations will include different sets of data, so no guarantees are given on anything besides that. Not sure I succeeded with that name though, so may change as suggested.

> * more of a question than a comment: the snapshot functionality will mean that we won't even run the bots if a MR hasn't changed, but what is the main cost of running the bots on a MR that hasn't changed? My guess would be that the expensive part are all the requests that the bots have to make in order to figure out that they don't have to do anything, but if we cache more aggressively, would that cost go away? Anyways, just a thought

There are plenty of other things we do in most WorkItems. We often instantiate a local workspace based on the PR source branch. This is quite expensive (5-10s in some repos). We instantiate some kind of Census instance, also in the order of seconds. In the mailinglist bridge, instantiating the marks workspace takes over 1 minute(!) but that is something I plan to address separately.

> Now for a bigger experiment 🧑‍🔬 One annoying thing is that the GitLab REST API doesn't return the comments for a MR when fetching a single MR (unlike GitHub which does). So I played around a bit with the GitLab GraphQL API and think I managed to achieve this:

Playing around with GraphQL could certainly help reducing the number of queries needed. I haven't tried that yet.

PR #1382, opened 2022-09-23
1364: Jira Issues should be resolved as "Fixed"

zhaosongzs:
Jira Issues will be resolved as "Fixed" regardless of the current state.

---------
### Progress

### Issue
 * [SKARA-1364](https://bugs.openjdk.org/browse/SKARA-1364): Jira Issues should be resolved as "Fixed"

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1382/head:pull/1382` \
`$ git checkout pull/1382`

`$ git checkout pull/1382` \
`$ git pull https://git.openjdk.org/skara pull/1382/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1382`

`$ git pr show -t 1382`

</details>

<a href="https://git.openjdk.org/skara/pull/1382.diff">https://git.openjdk.org/skara/pull/1382.diff</a>

</details>

### DIFF
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
@@ -196,14 +196,18 @@ class IssueNotifier implements Notifier, PullRequestListener, RepositoryListener
                     log.info("Resolving issue which status is not open, the status is " + issue.state());
                 } else {
                     log.info("Resolving issue " + issue.id());
-                }
-                issue.setState(Issue.State.RESOLVED);
-                if (issue.assignees().isEmpty()) {
-                    var username = findIssueUsername(commit, scratchPath);
-                    if (username.isPresent()) {
-                        var assignee = issueProject.issueTracker().user(username.get());
-                        assignee.ifPresent(hostUser -> issue.setAssignees(List.of(hostUser)));
+                    if (issue.assignees().isEmpty()) {
+                        var username = findIssueUsername(commit, scratchPath);
+                        if (username.isPresent()) {
+                            var assignee = issueProject.issueTracker().user(username.get());
+                            assignee.ifPresent(hostUser -> issue.setAssignees(List.of(hostUser)));
+                        }
                     }
+                }
+                if (!issue.isFixed()) {
+                    issue.setState(Issue.State.RESOLVED);
+                } else {
+                    log.info("The issue has been already resolved");
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


erikj79 (commented):
Since last we talked, I have been thinking about this some more. I think we actually do need to check the `resolution` of the issue before re-opening and resolving it. If it's already "resolved" "fixed" then we shouldn't touch the state. 

The reason for this is to try to keep the bots idempotent as much as possible. If we need to rerun the notifier on commits, then if nothing needs to happen, nothing should happen.

The challenge here is that Skara doesn't currently know how to read the resolution, so we would need to implement support for that.

erikj79:
> Since last we talked, I have been thinking about this some more. I think we actually do need to check the `resolution` of the issue before re-opening and resolving it. If it's already "resolved" "fixed" then we shouldn't touch the state.
> 
> The reason for this is to try to keep the bots idempotent as much as possible. If we need to rerun the notifier on commits, then if nothing needs to happen, nothing should happen.
> 
> The challenge here is that Skara doesn't currently know how to read the resolution, so we would need to implement support for that.

I just checked the source for `JiraIssue` and there is already a method `isFixed` which you can use. It should do exactly what we want here.

zhaosongzs:
> > Since last we talked, I have been thinking about this some more. I think we actually do need to check the `resolution` of the issue before re-opening and resolving it. If it's already "resolved" "fixed" then we shouldn't touch the state.
> > The reason for this is to try to keep the bots idempotent as much as possible. If we need to rerun the notifier on commits, then if nothing needs to happen, nothing should happen.
> > The challenge here is that Skara doesn't currently know how to read the resolution, so we would need to implement support for that.
> 
> I just checked the source for `JiraIssue` and there is already a method `isFixed` which you can use. It should do exactly what we want here.

Got it! I will use it.

bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
```diff
@@ -192,9 +192,10 @@ public void onIntegratedPullRequest(PullRequest pr, Path scratchPath, Hash hash)
 
             // If prOnly is false, this is instead done when processing commits
             if (prOnly && resolve) {
-                if (issue.state() == Issue.State.OPEN) {
+                if (issue.state() != Issue.State.OPEN) {
+                    log.info("Resolving issue which status is not open, the status is " + issue.state());
+                } else {
                     log.info("Resolving issue " + issue.id());
```
erikj79, line 198: No need for conditional here, we can use the same log message in regardless of current state. Also, updating assignees should also be done (as long as it's currently empty).
```suggestion
                log.info("Resolving issue " + issue.id() + " from state " + issue.state());
```

bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
```diff
@@ -203,6 +204,11 @@ public void onIntegratedPullRequest(PullRequest pr, Path scratchPath, Hash hash)
                         }
                     }
                 }
+                if (!issue.isFixed()) {
+                    issue.setState(Issue.State.RESOLVED);
+                } else {
+                    log.info("The issue has been already resolved");
```
erikj79, line 210:
```suggestion
                    log.info("The issue was already 'fixed'");
```
Same below.

bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
```diff
@@ -311,9 +317,10 @@ public void onNewCommits(HostedRepository repository, Repository localRepository
                 if (!alreadyPostedComment) {
                     issue.addComment(commitNotification);
                 }
-                if (issue.state() == Issue.State.OPEN) {
+                if (issue.state() != Issue.State.OPEN) {
```
zhaosongzs, line 320: Hi @erikj79 , I am wondering whether we need condition here. Do we only update the assignee when the issue is open?

zhaosongzs: If it it not, I think I should update one assertion in IssueNotifierTests#testAltFixVersionsMatch

erikj79: oh, forgot to say, same thing here as in the other method. We don't need this conditional.

bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
```diff
@@ -192,17 +192,20 @@ public void onIntegratedPullRequest(PullRequest pr, Path scratchPath, Hash hash)
...
-                if (issue.state() == Issue.State.OPEN) {
-                    log.info("Resolving issue " + issue.id());
-                    issue.setState(Issue.State.RESOLVED);
-                    if (issue.assignees().isEmpty()) {
-                        var username = findIssueUsername(commit, scratchPath);
-                        if (username.isPresent()) {
-                            var assignee = issueProject.issueTracker().user(username.get());
-                            assignee.ifPresent(hostUser -> issue.setAssignees(List.of(hostUser)));
-                        }
+                log.info("Resolving issue " + issue.id() + " from state " + issue.state());
```
erikj79, line 195: We should keep the log message and the issue.setState call together. I think moving lines 203-207 up here would be best to keep the flow as similar to before as we can. Same applies in the other method.

zhaosongzs: Sure

erikj79 (approved):

PR #1387, opened 2022-09-30
1603: Make labels handling consistent in all Issue implementations

zhaosongzs:
Make TestIssue and JiraIssue update the cache when updating labels.

---------
### Progress

### Issue
 * [SKARA-1603](https://bugs.openjdk.org/browse/SKARA-1603): Make labels handling consistent in all Issue implementations

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1387/head:pull/1387` \
`$ git checkout pull/1387`

`$ git checkout pull/1387` \
`$ git pull https://git.openjdk.org/skara pull/1387/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1387`

`$ git pr show -t 1387`

</details>

<a href="https://git.openjdk.org/skara/pull/1387.diff">https://git.openjdk.org/skara/pull/1387.diff</a>

</details>

### DIFF
--- a/issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraIssue.java
+++ b/issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraIssue.java
@@ -38,18 +38,22 @@ import java.util.stream.*;
 public class JiraIssue implements Issue {
     private final JiraProject jiraProject;
     private final RestRequest request;
-    private final JSONValue json;
+    private JSONValue json;
     private final boolean needSecurity;
 
     private final Logger log = Logger.getLogger("org.openjdk.skara.issuetracker.jira");
 
     private static final DateTimeFormatter dateFormat = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
 
+    private List<Label> labels;
+
     JiraIssue(JiraProject jiraProject, RestRequest request, JSONValue json) {
         this.jiraProject = jiraProject;
         this.request = request;
         this.json = json;
-
+        this.labels = json.get("fields").get("labels").stream()
+                .map(s -> new Label(s.asString()))
+                .collect(Collectors.toList());
         needSecurity = jiraProject.jiraHost().securityLevel().isPresent();
     }
 
@@ -260,6 +264,7 @@ public class JiraIssue implements Issue {
 
     @Override
     public void addLabel(String label) {
+        labels = null;
         var query = JSON.object()
                         .put("update", JSON.object()
                                            .put("labels", JSON.array().add(JSON.object()
@@ -269,6 +274,7 @@ public class JiraIssue implements Issue {
 
     @Override
     public void removeLabel(String label) {
+        labels = null;
         var query = JSON.object()
                         .put("update", JSON.object()
                                            .put("labels", JSON.array().add(JSON.object()
@@ -278,6 +284,7 @@ public class JiraIssue implements Issue {
 
     @Override
     public void setLabels(List<String> labels) {
+        this.labels = null;
         var labelsArray = JSON.array();
         for (var label : labels) {
             labelsArray.add(label);
@@ -291,9 +298,13 @@ public class JiraIssue implements Issue {
 
     @Override
     public List<Label> labels() {
-        return json.get("fields").get("labels").stream()
-                   .map(s -> new Label(s.asString()))
-                   .collect(Collectors.toList());
+        if(labels == null){
+            json = request.get("").execute();
+            labels = json.get("fields").get("labels").stream()
+                    .map(s -> new Label(s.asString()))
+                    .collect(Collectors.toList());
+        }
+        return labels;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraIssue.java
```diff
@@ -291,9 +298,13 @@ public void setLabels(List<String> labels) {
...
     public List<Label> labels() {
-        return json.get("fields").get("labels").stream()
-                   .map(s -> new Label(s.asString()))
-                   .collect(Collectors.toList());
+        if(labels == null){
+            json = request.get("").execute();
+            labels = json.get("fields").get("labels").stream()
+                    .map(s -> new Label(s.asString()))
+                    .collect(Collectors.toList());
+        }
```
zhaosongzs, line 306: I tried to just update the labels but I could not make it. So I update the json first and then update the labels.

issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraIssue.java
```diff
@@ -278,6 +284,7 @@ public void removeLabel(String label) {
 
     @Override
     public void setLabels(List<String> labels) {
+        this.labels = null;
```
erikj79, line 287: Both `GitHubPullRequest` and `GitHubMergeRequest` update the internal `labels` field in this method instead of making it null.

issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraIssue.java
```diff
@@ -291,9 +298,13 @@ public void setLabels(List<String> labels) {
 
     @Override
     public List<Label> labels() {
-        return json.get("fields").get("labels").stream()
-                   .map(s -> new Label(s.asString()))
-                   .collect(Collectors.toList());
+        if(labels == null){
```
erikj79, line 301:
```suggestion
        if (labels == null) {
```

issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraIssue.java
```diff
@@ -291,9 +298,13 @@ public void setLabels(List<String> labels) {
...
+            json = request.get("").execute();
```
erikj79, line 302: I don't think it's a good idea to have `JiraIssue::labels` sometimes update the whole json object. Just fetch it to a local variable and keep the `json` field final.

test/src/main/java/org/openjdk/skara/test/TestIssue.java
```diff
@@ -182,19 +182,22 @@ public boolean isFixed() {
...
     @Override
     public void removeLabel(String label) {
+        labels = null;
         store.labels().remove(label);
         store.setLastUpdate(ZonedDateTime.now());
     }
 
     @Override
     public void setLabels(List<String> labels) {
+        this.labels = null;
```
erikj79, line 200: This method should update the `labels` field instead of setting it to null.

test/src/main/java/org/openjdk/skara/test/TestIssue.java
```diff
@@ -48,7 +48,7 @@ public class TestIssue implements Issue {
     protected final String title;
     protected final State state;
     // Mimic JiraIssue where labels are part of the main JSON object
```
erikj79, line 50: This comment is now outdated.
```suggestion
    // Labels are cached but still kept up to date
```

issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraIssue.java
zhaosongzs: I think we update the internal labels in GitHubPullRequest and GitHubMergeRequest because we could get the return value of the update request. However, for Jira Issue, I read the document of Jira REST API, it shows that it wouldn't have return value of the update request, it will just return Status 204.
Reference:
https://developer.atlassian.com/server/jira/platform/jira-rest-api-examples/

zhaosongzs: So we update the internal labels in GitHubPullRequest and GitLabMerge request could help us reduce one request. But update the internal labels here, I think we wouldn't get any benefit.

test/src/main/java/org/openjdk/skara/test/TestIssue.java
zhaosongzs: The same with JiraIssue

issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraIssue.java
zhaosongzs: NVM, my bad, update here would also reduce one request, I will change it

erikj79 (approved):
This looks good. Hopefully tests are still passing.

It would be good to walk through all instances of label handling with `JiraIssue` to make sure this change in behavior won't be a problem. I don't think there will be problems, but some kind of manual inspection would be appreciated.

PR #1388, opened 2022-10-03
1215: Clean label not updating correctly

zhaosongzs:
Originally, If the user mistakenly creates a Backport PR with the wrong hash and then later corrects it, the clean status is never correctly re-evaluated. 

Currently, If the user correct the hash, the bot will be forced to do the check again and could correctly add clean label.

---------
### Progress

### Issue
 * [SKARA-1215](https://bugs.openjdk.org/browse/SKARA-1215): Clean label not updating correctly

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1388/head:pull/1388` \
`$ git checkout pull/1388`

`$ git checkout pull/1388` \
`$ git pull https://git.openjdk.org/skara pull/1388/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1388`

`$ git pr show -t 1388`

</details>

<a href="https://git.openjdk.org/skara/pull/1388.diff">https://git.openjdk.org/skara/pull/1388.diff</a>

</details>

### DIFF
--- a/bots/pr/src/test/java/org/openjdk/skara/bots/pr/BackportTests.java
+++ b/bots/pr/src/test/java/org/openjdk/skara/bots/pr/BackportTests.java
@@ -1576,4 +1576,123 @@ class BackportTests {
             assertEquals(List.of(), message.additional());
         }
     }
+
+    @Test
+    void updateOriginalHashFromWrongToCorrect(TestInfo testInfo) throws IOException {
+        try (var credentials = new HostCredentials(testInfo);
+             var tempFolder = new TemporaryDirectory()) {
+
+            var author = credentials.getHostedRepository();
+            var integrator = credentials.getHostedRepository();
+            var reviewer = credentials.getHostedRepository();
+            var issues = credentials.getIssueProject();
+            var censusBuilder = credentials.getCensusBuilder()
+                    .addCommitter(author.forge().currentUser().id())
+                    .addReviewer(integrator.forge().currentUser().id())
+                    .addReviewer(reviewer.forge().currentUser().id());
+            var bot = PullRequestBot.newBuilder()
+                    .repo(integrator)
+                    .censusRepo(censusBuilder.build())
+                    .issueProject(issues)
+                    .build();
+
+            // Populate the projects repository
+            var localRepo = CheckableRepository.init(tempFolder.path(), author.repositoryType());
+            var masterHash = localRepo.resolve("master").orElseThrow();
+            localRepo.push(masterHash, author.url(), "master", true);
+
+            var releaseBranch = localRepo.branch(masterHash, "release");
+            localRepo.checkout(releaseBranch);
+            var newFile = localRepo.root().resolve("a_new_file.txt");
+            Files.writeString(newFile, "hello");
+            localRepo.add(newFile);
+            var issue1 = credentials.createIssue(issues, "An issue");
+            var issue1Number = issue1.id().split("-")[1];
+            var originalMessage = issue1Number + ": An issue\n" +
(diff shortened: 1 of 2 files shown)


erikj79:
I've closed this issue as a duplicate of SKARA-1449. I don't think the fix proposed here is needed.

bots/pr/src/test/java/org/openjdk/skara/bots/pr/BackportTests.java
```diff
@@ -1576,4 +1576,123 @@ void updateOriginal(TestInfo testInfo) throws IOException {
...
+                    "\n" +
+                    "Reviewed-by: integrationreviewer2";
+            var releaseHash = localRepo.commit(originalMessage, "integrationcommitter1", "integrationcommitter1@openjdk.org");
+            localRepo.push(releaseHash, author.url(), "refs/heads/release", true);
+
+            var newFileSZ = localRepo.root().resolve("a_new_file2.txt");
+            Files.writeString(newFileSZ, "hello2");
+            localRepo.add(newFileSZ);
+            var releaseHash2 =localRepo.commit(originalMessage, "integrationcommitter1", "integrationcommitter1@openjdk.org");
+            localRepo.push(releaseHash2, author.url(), "refs/heads/release", true);
```
zhaosongzs, line 1621: Hi @erikj79 , is this situation possible? I mean if we have a branch named 'release', in the first commit, it added a new file 'a_new_file.txt', in the second commit it added a new file 'a_new_file2.txt'. And we have an edit branch, in its commit, it added a new file 'a_new_file.txt'. Then we create the backport incorrectly with the hash of the second commit of release branch. So the backport is not clean. After that, if we correct the hash to the first commit of release, the check will not run again and the clean label will not be added to the backport.

erikj79: Changing the title will cause the check metadata to change, so `currentCheckValid` will be false.

erikj79: I think the new comment should trigger a metadata change.

zhaosongzs: > I think the new comment should trigger a metadata change.

Do you mean the commit message? Yeah, if we change the commit message it will trigger a metadata change. So in this situation, the commit message will always be changed, right?

erikj79: I see now that it doesn't. I need to think about this.

erikj79: Ah, the problem is that `<!-- backport` is not matched as a metadata comment in `CheckWorkItem::getMetadata`. See the regexp `metadataComments`. If you add that, it should consider the backport hash comments as updates that need rechecking.

erikj79: The PR user here is the bot user. That filter is making sure we are only considering comments made by the bot itself. The idea here is that new check runs should always be triggered when a special "metadata" comment is added or changed. These are basically book keeping comments made by the bot to record various kinds of state in the PR. The data in these comments are hidden from view by being written inside html comments `<!-- -->`. Not including the backport hash metadata comment here was an oversight.

erikj79 (approved):

lgxbslgx (approved):
LGTM

PR #1390, opened 2022-10-05
1616: Improve error message when an author isn't found in GitLab

zhaosongzs:
Improved the error message in GitLabMergeRequest:author

---------
### Progress

### Issue
 * [SKARA-1616](https://bugs.openjdk.org/browse/SKARA-1616): Improve error message when an author isn't found in GitLab

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1390/head:pull/1390` \
`$ git checkout pull/1390`

`$ git checkout pull/1390` \
`$ git pull https://git.openjdk.org/skara pull/1390/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1390`

`$ git pr show -t 1390`

</details>

<a href="https://git.openjdk.org/skara/pull/1390.diff">https://git.openjdk.org/skara/pull/1390.diff</a>

</details>

### DIFF
--- a/forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabMergeRequest.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabMergeRequest.java
@@ -77,7 +77,13 @@ public class GitLabMergeRequest implements PullRequest {
 
     @Override
     public HostUser author() {
-        return repository.forge().user(json.get("author").get("username").asString()).get();
+        var username = json.get("author").get("username").asString();
+        var author = repository.forge().user(username);
+        if (author.isPresent()) {
+            return author.get();
+        } else {
+            throw new IllegalArgumentException("Unknown username: " + username);
+        }
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabMergeRequest.java
```diff
@@ -77,7 +77,13 @@ public String id() {
 
     @Override
     public HostUser author() {
-        return repository.forge().user(json.get("author").get("username").asString()).get();
+        var username = json.get("author").get("username").asString();
+        var author = repository.forge().user(username);
+        if (author.isPresent()) {
+            return author.get();
+        } else {
+            throw new IllegalArgumentException("Unknown username: " + username);
```
erikj79, line 85: I don't think IllegalArgumentException is correct here. It's meant to be thrown when a method receives a bad argument, but this method doesn't take arguments. We don't really have a reason to define a new exception type either, so a plain `RuntimeException` is probably better, at least for now.

I'm usually for short and concise messages, but in this case I think we can add some more details. Something like `"Author of merge request unknown: " + username`

zhaosongzs: Ok, I also thought about the type of exception yesterday. I was thinking that the username is an argument of `user(username)`, but throw exception in method `user` is a little complex. 

I will change it to RuntimeException and make the error message have more details.

erikj79 (approved):

lgxbslgx (approved):

PR #1391, opened 2022-10-05
1618: CSR bot failed to update PR when CSR was approved

erikj79:
This patch fixes a mistake in the CSRBotFactory. The intention was to create one CSRIssueBot for each unique IssueProject, but we currently get one per repository instead. This is because JiraProject doesn't implement hashcode, so we can't trust a HashSet to sort out uniqueness. I'm changing this to a HashMap and using the configuration string as key instead.

For details on how this was causing the reported problem, see comments in the bug.

---------
### Progress

### Issue
 * [SKARA-1618](https://bugs.openjdk.org/browse/SKARA-1618): CSR bot failed to update PR when CSR was approved

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1391/head:pull/1391` \
`$ git checkout pull/1391`

`$ git checkout pull/1391` \
`$ git pull https://git.openjdk.org/skara pull/1391/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1391`

`$ git pr show -t 1391`

</details>

<a href="https://git.openjdk.org/skara/pull/1391.diff">https://git.openjdk.org/skara/pull/1391.diff</a>

</details>

### DIFF
--- a/bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBotFactory.java
+++ b/bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRBotFactory.java
@@ -23,7 +23,6 @@
 package org.openjdk.skara.bots.csr;
 
 import java.util.HashMap;
-import java.util.HashSet;
 import org.openjdk.skara.bot.*;
 
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


lgxbslgx (commented):
This patch reduces the duplicated bots and maps one issue bot to one IssueProject. It also solves the SKARA-1618 as a side effect.

But I don't think it is the root cause of SKARA-1618. The `IssueWorkItem#concurrentWith` currently only checks the issue project's name and doesn't check the issue id and repos' names so that these work items can't be run concurrently and some of them will be eliminated (Because the method `WorkItem#replaces` returns `true`).

So I suggest to file another issue to record this patch and solve SKARA-1618 by revising the method `IssueWorkItem#concurrentWith` in another patch. What is your opinion?

erikj79:
> But I don't think it is the root cause of SKARA-1618. The `IssueWorkItem#concurrentWith` currently only checks the issue project's name and doesn't check the issue id and repos' names so that these work items can't be run concurrently and some of them will be eliminated (Because the method `WorkItem#replaces` returns `true`).
> 
> So I suggest to file another issue to record this patch and solve SKARA-1618 by revising the method `IssueWorkItem#concurrentWith` in another patch. What is your opinion?

Good catch that the concurrentWith method is also wrong. I don't think there is any point in splitting the fixes though, I will fix that method in this change.

lgxbslgx (approved):
Looks good.

edvbld (approved):
Looks good

PR #1393, opened 2022-10-10
SKARA-1627

zhaosongzs:
Add some tests for BotFactories.


zhaosongzs:
Hi @erikj79 ,this commit contains test for CSRBotFactory.  If you think this is ok, I would add tests for other bot factors as well.

erikj79:
> Hi @erikj79 ,this commit contains test for CSRBotFactory. If you think this is ok, I would add tests for other bot factories as well.

I started to look around the source and I found TestBotFactory which seems to already be doing most of what I was envisioning here. Only two existing bots are using it for their tests, which is why I've missed it before. With that class you can supply just the JSON for the bot factory under test, and any other entities (i.e. IssueTracker or forge) are added as already constructed objects (using the test implementations).

Unless it's really needed, I would advice having the JSON defined inline in the test using the new triple quote multiline string literals (https://blogs.oracle.com/javamagazine/post/text-blocks-come-to-java). The size of these snippets should be fairly small as they only need to contain test configuration for a single bot.

zhaosongzs:
> > Hi @erikj79 ,this commit contains test for CSRBotFactory. If you think this is ok, I would add tests for other bot factories as well.
> 
> I started to look around the source and I found TestBotFactory which seems to already be doing most of what I was envisioning here. Only two existing bots are using it for their tests, which is why I've missed it before. With that class you can supply just the JSON for the bot factory under test, and any other entities (i.e. IssueTracker or forge) are added as already constructed objects (using the test implementations).
> 
> Unless it's really needed, I would advice having the JSON defined inline in the test using the new triple quote multiline string literals (https://blogs.oracle.com/javamagazine/post/text-blocks-come-to-java). The size of these snippets should be fairly small as they only need to contain test configuration for a single bot.

Hi Erik, I looked at the TestBotFactory and I thought that if we want to let the IssueTracker or Forge public for all botFactory tests, we should test all botFactories in one test. I don't know if it is a good idea. I prefer to create a test for each botFactory. 

And I tried to use TestBotFactory, in the test, there will be some trivial problems like the name of Hosts and IssueHosts, it will make us need to create Host or IssueHost manually and make our test more complex. So I prefer to retain the config of forges and issueTrackers and still use TestBotRunnerConfiguration to create the hosts. What is your idea?

In the case below, I used the credentials to create HostedRepository and issueTrackers, the default name for the HostedRepository and issueTrackers are all test, so after that, all the bots we got are having the same name. If we want to create HostedRepository or IssueTrackers with specific name, I think use TestBotRunnerConfiguration is a good idea.

    @Test
    public void testCreate(TestInfo testInfo) throws IOException, ConfigurationError {
        try (var credentials = new HostCredentials(testInfo);
             var tempFolder = new TemporaryDirectory()) {

            var repo = credentials.getHostedRepository();
            var repoFolder = tempFolder.path().resolve("repo");
            var localRepo = CheckableRepository.init(repoFolder, repo.repositoryType());
            var issueTracker1 = credentials.getIssueProject();
            var issueTracker2 = credentials.getIssueProject();
            credentials.commitLock(localRepo);
            localRepo.pushAll(repo.url());


            String jsonString = """
                     {
                       "projects": [
                         {
                           "repository": "test@github:jdk1",
                           "issues": "test_bugs/TEST"
                         },
                         {
                           "repository": "test@github:jdk2",
                           "issues": "test_bugs/TEST"
                         },
                         {
                           "repository": "test@github:jdk3",
                           "issues": "test_bugs/TEST"
                         },
                         {
                           "repository": "test@github:jdk4",
                           "issues": "test_bugs/TEST"
                         },
                         {
                           "repository": "test@github:jdk5",
                           "issues": "test_bugs/TEST"
                         },
                         {
                           "repository": "test@github:jdk6",
                           "issues": "test_bugs/TEST2"
                         }
                       ]
                     }
                    """;
            var jsonConfig = JWCC.parse(jsonString).asObject();

            var testBotFactory = TestBotFactory.newBuilder().addHostedRepository("test@github", repo)
                    .addIssueProject("test_bugs/TEST", issueTracker1)
                    .addIssueProject("test_bugs/TEST2", issueTracker2)
                    .build();

            var bots = testBotFactory.createBots("csr", jsonConfig);
            assertEquals(8, bots.size());

            var csrPullRequestBots = bots.stream().filter(e -> e.getClass().equals(CSRPullRequestBot.class)).collect(Collectors.toList());
            var csrIssueBots = bots.stream().filter(e -> e.getClass().equals(CSRIssueBot.class)).collect(Collectors.toList());

            // a CSRPullRequestBot for every configured repository
            assertEquals(6, csrPullRequestBots.size());
            // a CSRIssueBot for each unique IssueProject
            assertEquals(2, csrIssueBots.size());

            for (var bot : csrIssueBots) {
                CSRIssueBot csrIssueBot = (CSRIssueBot) bot;
                if (csrIssueBot.toString().equals("CSRIssueBot@TEST")) {
                    assertEquals(5, csrIssueBot.repositories().size());
                } else if (csrIssueBot.toString().equals("CSRIssueBot@TEST2")) {
                    assertEquals(1, csrIssueBot.repositories().size());
                } else {
                    throw new RuntimeException("This issue bot is not expected");
                }
            }
        }
    }
}
`


For the JSON, I would use the new triple quote multiline string literals. Thanks!

erikj79:
I actually don't think your example looks that bad. The annoying part is having to create a real `TestHostedRepository` through the `HostedCredentials`, because that will create a real git repository on disk for each test, which gets expensive. But, since we aren't actually going to use that `TestHostedRepository` for anything other than supplying it to the factory, we can probably cheat and just create it with its public constructor, with mostly null values. (If we had been using a mock framework in Skara tests, this would have been so much easier, but as it stands we need to use semi real objects.) Same goes for the TestIssueProject. I think we can get away with not having a HostCredentials instance at all.

bots/csr/src/test/java/org/openjdk/skara/bots/csr/CSRBotFactoryTest.java
```diff
@@ -0,0 +1,178 @@
...
+            if (csrIssueBot.toString().equals("CSRIssueBot@TEST")) {
+                assertEquals(5, csrIssueBot.repositories().size());
+            } else if (csrIssueBot.toString().equals("CSRIssueBot@TEST2")) {
+                assertEquals(1, csrIssueBot.repositories().size());
+            } else {
+                throw new RuntimeException("This issue bot is not expected");
+            }
+        }
+    }
+}
```
zhaosongzs, line 178: I tried to write the test with `TestBotFactory`, but I still couldn't get the advantage of using `TestBotFactory`.

**Using `TestBotRunnerConfiguration`:**
pros:
1. We just override two parsers in `BotRunnerConfiguration`, so we could go through most of code in real `BotRunnerConfiguration` and `BotConfiguration`.
2. We don't need to create RepositoryHost or IssueHost manually.

cons:
1. we need the config of forges and issueTrackers
--------------------------------------------------------------------------------------------------------------

**Using `TestBotFactory`**:
pros:
1. we don't need the config of forges and issueTrackers

cons:
1. we need to create all HostedRepositories and issueProjects manually because the fields in TestBotFactory is `private final Map<String, HostedRepository> hostedRepositories;` & `private final Map<String, IssueProject> issueProjects;`, otherwise, the contents in the bots will be wrong.
2.  `BotConfiguration` in `TestBotFactory` is not the real one we use.
--------------------------------------------------------------------------------------------------------------

All in all, I think use the `TestBotRunnerConfiguration` will be more realistic.

zhaosongzs: Hi @erikj79 , if you have any time, please have a look at this comment.

erikj79: I still think that using the already existing TestBotFactory is the way to go. Here are my reasons:

1. Adding another test implementation doing something very similar to what we already have is generally bad from a future maintenance perspective.
2. Having to open up the production class `BotRunnerConfiguration` for subclassing, just to be able to create a test class is sometimes a necessary pattern, but should generally be avoided if possible.
3. There is no need to create a `RepositoryHost` and `IssueHost` to instantiate `TestRepository` or `TestIssue`, you can just null those fields using the public constructor. We also don't need a local `Repository` instance (if we fix the constructor in `TestRepository` to stop calling `::url`).
4. I don't want these unit tests to need JSON definitions of forges and issue trackers. Especially if those need to fake actual github and jira configurations. That doesn't add any value to the tests, but makes these tests depend on how those entities are configured. The scope of the tests in question is to verify the behavior of the specific `*BotFactory` class, not parsing any other parts of a configuration file. 
5. Needing to create the `HostedRepository` and IssueProject` instances programmatically is actually a feature here. As an example, allows us to verify that the factory picks up the correct `RemoteRepository`, as the JSON does not allow us to define specific repositories, just a host.
6. Having a test implementation of `BotConfiguration` being used in these tests is a good thing. It makes it easier to customize specific behavior in a test to check corner cases. This is what unit tests are for. Granted, the current implementation is an inline class in TestBotFactory, but that can be refactored if needed. As I mentioned in an earlier comment, if we had access to a mock framework here, I would prefer mocking the `BotConfiguration` completely and just call the `*BotFactory::create` directly, but as we don't have such a dependency present in this project, using the existing test implementation classes is the next best thing.

zhaosongzs: > I still think that using the already existing TestBotFactory is the way to go. Here are my reasons:
> 
> 1. Adding another test implementation doing something very similar to what we already have is generally bad from a future maintenance perspective.
> 2. Having to open up the production class `BotRunnerConfiguration` for subclassing, just to be able to create a test class is sometimes a necessary pattern, but should generally be avoided if possible.
> 3. There is no need to create a `RepositoryHost` and `IssueHost` to instantiate `TestRepository` or `TestIssue`, you can just null those fields using the public constructor. We also don't need a local `Repository` instance (if we fix the constructor in `TestRepository` to stop calling `::url`).
> 4. I don't want these unit tests to need JSON definitions of forges and issue trackers. Especially if those need to fake actual github and jira configurations. That doesn't add any value to the tests, but make these tests depend on how those entities are configured. The scope of the tests in question is to verify the behavior of the specific `*BotFactory` class, not parsing any other parts of a configuration file.
> 5. Needing to create the `HostedRepository` and `IssueProject` instances programmatically is actually a feature here. As an example, it allows us to verify that the factory picks up the correct `RemoteRepository`, as the JSON does not allow us to define specific repositories, just a host.
> 6. Having a test implementation of `BotConfiguration` being used in these tests is a good thing. It makes it easier to customize specific behavior in a test to check corner cases. This is what unit tests are for. Granted, the current implementation is an inline class in TestBotFactory, but that can be refactored if needed. As I mentioned in an earlier comment, if we had access to a mock framework here, I would prefer mocking the `BotConfiguration` completely and just call the `*BotFactory::create` directly, but as we don't have such a dependency present in this project, using the existing test implementation classes is the next best thing.

Thanks for the detailed explanation, I think it makes sense now and  will use `TestBotFactory`.

bots/hgbridge/src/test/java/org/openjdk/skara/bots/hgbridge/JBridgeBotFactoryTest.java
```diff
@@ -0,0 +1,72 @@
...
+                    .build();
+
+            var bots = testBotFactory.createBots(JBridgeBotFactory.NAME, jsonConfig);
+            // A JBridgeBot for every configured repo
+            assertEquals(2, bots.size());
+
+            JBridgeBot jBridgeBot1 = (JBridgeBot) bots.get(0);
+            assertEquals("JBridgeBot@https://test.org/source1", jBridgeBot1.toString());
+            JBridgeBot jBridgeBot2 = (JBridgeBot) bots.get(1);
+            assertEquals("JBridgeBot@https://test.org/source2", jBridgeBot2.toString());
```
erikj79, line 69: I think this bot is a good example of where we should probably add some accessors to be able to verify the configuration in more detail.

bots/mlbridge/src/test/java/org/openjdk/skara/bots/mlbridge/MailingListBridgeBotFactoryTest.java
```diff
@@ -0,0 +1,155 @@
...
+            List<Bot> mailingListBridgeBots = bots.stream().filter(e -> e.getClass().equals(MailingListBridgeBot.class)).toList();
+
+            assertEquals(2, mailingListArchiveReaderBots.size());
+            assertEquals(3, mailingListBridgeBots.size());
+
+            assertEquals("MailingListArchiveReaderBot@repo3", mailingListArchiveReaderBots.get(0).toString());
+            assertEquals("MailingListArchiveReaderBot@repo5", mailingListArchiveReaderBots.get(1).toString());
+            assertEquals("MailingListBridgeBot@repo3", mailingListBridgeBots.get(0).toString());
+            assertEquals("MailingListBridgeBot@repo4", mailingListBridgeBots.get(1).toString());
+            assertEquals("MailingListBridgeBot@repo5", mailingListBridgeBots.get(2).toString());
```
erikj79, line 152: It looks like there are more fields we would want to verify here too.

bots/submit/src/test/java/org/openjdk/skara/bots/submit/SubmitBotFactoryTest.java
```diff
@@ -0,0 +1,49 @@
+package org.openjdk.skara.bots.submit;
+
+import org.junit.jupiter.api.Test;
+import org.openjdk.skara.json.JWCC;
+import org.openjdk.skara.test.TestBotFactory;
+import org.openjdk.skara.test.TestHostedRepository;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class SubmitBotFactoryTest {
```
erikj79, line 10: This bot isn't actually used, so no need to add more to this test.

bots/tester/src/test/java/org/openjdk/skara/bots/tester/TestBotFactoryTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+import org.junit.jupiter.api.Test;
+import org.openjdk.skara.json.JWCC;
+import org.openjdk.skara.test.TemporaryDirectory;
+import org.openjdk.skara.test.TestBotFactory;
+import org.openjdk.skara.test.TestHostedRepository;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class TestBotFactoryTest {
```
erikj79, line 11: This bot isn't actually used, so no need to add more to this test.

test/src/main/java/org/openjdk/skara/test/TestBotFactory.java
```diff
@@ -156,4 +156,77 @@ public JSONObject specific() {
         }
         throw new RuntimeException("Failed to find bot factory with name: " + name);
     }
+
+    public List<Bot> createBots(String name, JSONObject configuration) {
```
erikj79, line 160: This is a copy of the method above. They should share implementation. You can make `create` call `createBots` and keep lines 151 to 154 to convert the list.

test/src/main/java/org/openjdk/skara/test/TestHostedRepository.java
```diff
@@ -55,6 +55,15 @@ public TestHostedRepository(TestHost host, String projectName, Repository localR
         commitComments = new HashMap<>();
     }
 
+    public TestHostedRepository(TestHost host, String projectName) {
```
erikj79, line 58: Could you add a comment explaining the limitations of creating an instance with this constructor? Something like:
"Creates an instance without a backing local repository that will not support any actual repository interaction."

Also since we are always calling this with host=null, we might as well remove that parameter too.

erikj79 (commented):
This is starting to look pretty good. In several bots, there is more configuration data that needs verification, and to do that we need to add accessors to the bot class. I didn't comment individually on all those tests as it turned out to be quite a few. There are also some bots that aren't really used, so we shouldn't spend more time making the tests better for them, but these I did mark.

test/src/main/java/org/openjdk/skara/test/TestBotFactory.java
zhaosongzs: > This is a copy of the method above. They should share implementation. You can make `create` call `createBots` and keep lines 151 to 154 to convert the list.

Ok, but the implementations of `continuousIntegration` and `repository` in `BotConfiguration` are a little different between `create` and `createBots`. Hope it will not affect other tests, althoughI don't think it's going to have any bad effects.

erikj79: Ah I didn't realize you changed them. Ideally we should be able to have a unified implementation.

test/src/main/java/org/openjdk/skara/test/TestHostedRepository.java
```diff
@@ -55,6 +55,30 @@ public TestHostedRepository(TestHost host, String projectName, Repository localR
...
+        this.projectName = projectName;
+        this.localRepository = null;
+        pullRequestPattern = null;
+        commitComments = new HashMap<>();
+    }
+
+    /**
+     * Creates an instance without a backing local repository that will not support any actual repository interaction
+     */
+    public TestHostedRepository(TestHost host, String projectName) {
```
zhaosongzs, line 73: Still need this constructor because in a test, we need a `TestHostedRepository` with `TestHost`

erikj79 (approved):
Looks good! Thanks for writing all these tests.

PR #1394, opened 2022-10-11
SKARA-1633

erikj79:
A combination of changes in SKARA-1584 and SKARA-1606 is causing a test failure. The problem is in `TestPullRequest::review`, where we will sometimes (most often) return the internal ArrayList<Review> from the underlying store object. This is causing TestPullRequest::snapshot to not return a pure snapshot, but instead something that gets updated when the store is updated, defeating the whole idea of the snapshot.

The fix is to wrap the return value with a `List::copyOf`. I'm also applying this to `Issue::comments` as that's also part of the snapshot and using unmodifiable collections is usually preferable.


edvbld (approved):
Looks good, thanks for fixing!

PR #1396, opened 2022-10-12
1611: Improve adding and removing of labels in GitLab

zhaosongzs:
Improved adding and removing of labels in GitLabMergeRequest.

And make the label handling consistent with GitHubPullRequest.

---------
### Progress

### Issue
 * [SKARA-1611](https://bugs.openjdk.org/browse/SKARA-1611): Improve adding and removing of labels in GitLab

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1396/head:pull/1396` \
`$ git checkout pull/1396`

`$ git checkout pull/1396` \
`$ git pull https://git.openjdk.org/skara pull/1396/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1396`

`$ git pr show -t 1396`

</details>

<a href="https://git.openjdk.org/skara/pull/1396.diff">https://git.openjdk.org/skara/pull/1396.diff</a>

</details>

### DIFF
--- a/forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabMergeRequest.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabMergeRequest.java
@@ -658,15 +658,9 @@ public class GitLabMergeRequest implements PullRequest {
 
     @Override
     public void addLabel(String label) {
-        // GitLab does not allow adding/removing single labels, only setting the full list
-        // We retrieve the list again here to try to minimize the race condition window
-        var currentJson = request.get("").execute().asObject();
-        var labels = Stream.concat(currentJson.get("labels").stream()
-                                .map(JSONValue::asString),
-                        Stream.of(label))
-                .collect(Collectors.toSet());
+        labels = null;
         request.put("")
-                .body("labels", String.join(",", labels))
+                .body("add_labels", label)
                 // Temporary workaround for GitLab returning 500 when changing labels.
                 // The labels are still modified.
                 .onError(response -> {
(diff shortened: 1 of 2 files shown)


forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabMergeRequest.java
```diff
@@ -710,6 +698,12 @@ public void setLabels(List<String> labels) {
 
     @Override
     public List<Label> labels() {
+        if (labels == null) {
+            labels = request.get("").execute().get("labels").stream()
+                    .map(JSONValue::asString)
+                    .sorted()
+                    .collect(Collectors.toList());
+        }
```
zhaosongzs, line 706: I don't know if we need to make it consistent with GitHubPullRequest but I did.

erikj79: Please note that GitLabMergeRequest handles labels a bit differently to JiraIssue and GitHubPullRequest. The internal cache is a list of String, and we override the labelNames() method. This is all because of an optimization I made recently.

This means that the lazy fetch needs to happen in labelNames() instead, and the labels() method should call labelNames() to get the list to convert to Label objects.

erikj79 (commented):
When changing API calls like this, they need to be tested. The normal automatic tests do not cover it. See GitLabRestApiTest and add a test there to verify that labels can be added/removed/set with the new implementation. You will need to provide credentials in a special properties file (that should NOT be committed). You can generate an access token for your own user to test this out against a personal repository.

zhaosongzs:
> When changing API calls like this, they need to be tested. The normal automatic tests do not cover it. See GitLabRestApiTest and add a test there to verify that labels can be added/removed/set with the new implementation. You will need to provide credentials in a special properties file (that should NOT be committed). You can generate an access token for your own user to test this out against a personal repository.

Before creating this MR, I tested it with my own repo and I think it works well. I will add tests in GitLabRestApiTest.

erikj79 (approved):

PR #1399, opened 2022-10-19
1587: Adding review comments should not mean approval status change

zhaosongzs:
The issue here is ' if a user leaves additional comments using the "Review changes" dialog, and selects the "Comments", then their approval is considered revoked.'

I believe this happens because if we use "Review changes" to add a comment, it will add a review with `Review.Verdict.NONE` to our pr.  According to our current logic, this new review will cover old reviews, so the approval will be revoked.

I made changes to `CheckablePullRequest#filterActiveReviews`. Now If latest review is just a comment, the verdict will inherit from the previous review.

---------
### Progress

### Issue
 * [SKARA-1587](https://bugs.openjdk.org/browse/SKARA-1587): Adding review comments should not mean approval status change

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1399/head:pull/1399` \
`$ git checkout pull/1399`

`$ git checkout pull/1399` \
`$ git pull https://git.openjdk.org/skara pull/1399/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1399`

`$ git pr show -t 1399`

</details>

<a href="https://git.openjdk.org/skara/pull/1399.diff">https://git.openjdk.org/skara/pull/1399.diff</a>

</details>

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckablePullRequest.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckablePullRequest.java
@@ -140,9 +140,9 @@ public class CheckablePullRequest {
                 var reviewCorrectTarget = review.targetRef().equals(targetRef);
                 var reviewNewer = prevReview.createdAt().isBefore(review.createdAt());
 
-                if ((prevReviewCorrectTarget && reviewCorrectTarget && reviewNewer)
+                if ((!review.verdict().equals(Review.Verdict.NONE)) && ((prevReviewCorrectTarget && reviewCorrectTarget && reviewNewer)
                         || (!prevReviewCorrectTarget && !reviewCorrectTarget && reviewNewer)
-                        || (!prevReviewCorrectTarget && reviewCorrectTarget)) {
+                        || (!prevReviewCorrectTarget && reviewCorrectTarget))) {
                     reviewPerUser.put(review.reviewer(), review);
                 }
             } else {
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckablePullRequest.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckablePullRequest.java
@@ -140,7 +140,8 @@ public class CheckablePullRequest {
                 var reviewCorrectTarget = review.targetRef().equals(targetRef);
                 var reviewNewer = prevReview.createdAt().isBefore(review.createdAt());
 
-                if ((!review.verdict().equals(Review.Verdict.NONE)) && ((prevReviewCorrectTarget && reviewCorrectTarget && reviewNewer)
+                if ((!review.verdict().equals(Review.Verdict.NONE))
+                        && ((prevReviewCorrectTarget && reviewCorrectTarget && reviewNewer)
                         || (!prevReviewCorrectTarget && !reviewCorrectTarget && reviewNewer)
                         || (!prevReviewCorrectTarget && reviewCorrectTarget))) {
                     reviewPerUser.put(review.reviewer(), review);
--- a/bots/pr/src/test/java/org/openjdk/skara/bots/pr/ReviewersTests.java
+++ b/bots/pr/src/test/java/org/openjdk/skara/bots/pr/ReviewersTests.java
@@ -646,40 +646,35 @@ public class ReviewersTests {
             TestBotRunner.runPeriodicItems(prBot);
             TestBotRunner.runPeriodicItems(prBot);
             // The pr should contain 'Ready' label
-            var updatedPr = author.pullRequest(pr.id());
-            assertTrue(updatedPr.labelNames().contains("ready"));
+            assertTrue(pr.store().labelNames().contains("ready"));
 
             // Add a review comment
             reviewerPr.addReview(Review.Verdict.NONE, "Just a comment1");
             TestBotRunner.runPeriodicItems(prBot);
             TestBotRunner.runPeriodicItems(prBot);
             // The pr should still contain 'Ready' label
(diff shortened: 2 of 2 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckablePullRequest.java
```diff
@@ -143,6 +143,16 @@ static List<Review> filterActiveReviews(List<Review> allReviews, String targetRe
...
+                    // If current review is just a comment, the verdict should inherit from the previous review
+                    if (review.verdict().equals(Review.Verdict.NONE)) {
+                        review = new Review(review.createdAt(),
+                                review.reviewer(),
+                                prevReview.verdict(),
+                                review.hash().isPresent() ? review.hash().get() : null,
+                                review.id(),
+                                review.body().isPresent() ? review.body().get() : null,
+                                review.targetRef());
+                    }
```
zhaosongzs, line 155: I don't whether it is a good idea to create a new review here.

If we don't want to create a new review, we should make `verdict` not final and add a set method in Review.

bots/pr/src/test/java/org/openjdk/skara/bots/pr/ReviewersTests.java
```diff
@@ -613,4 +613,73 @@ void testZeroReviewer(TestInfo testInfo) throws IOException {
...
+            var updatedPr = author.pullRequest(pr.id());
+            assertTrue(updatedPr.labelNames().contains("ready"));
+
+            // Add a review comment
+            reviewerPr.addReview(Review.Verdict.NONE, "Just a comment1");
+            TestBotRunner.runPeriodicItems(prBot);
+            TestBotRunner.runPeriodicItems(prBot);
+            // The pr should still contain 'Ready' label
+            updatedPr = author.pullRequest(pr.id());
+            assertTrue(updatedPr.labelNames().contains("ready"));
```
zhaosongzs, line 658: I am still wondering why we should update the pr manually.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckablePullRequest.java
erikj79: I don't think we should create a new `Review` here. I _think_ that the way to achieve the desired behavior is to ignore reviews where the verdict isn't `APPROVED` or `DISAPPROVED`. A `Review` with verdict `NONE` is essentially pointless when considering active reviews.

zhaosongzs: > I don't think we should create a new `Review` here. I _think_ that the way to achieve the desired behavior is to ignore reviews where the verdict isn't `APPROVED` or `DISAPPROVED`. A `Review` with verdict `NONE` is essentially pointless when considering active reviews.

Ah, exactly! Thanks for the suggestion! I will change it.

bots/pr/src/test/java/org/openjdk/skara/bots/pr/ReviewersTests.java
erikj79: An instance of `PullRequest` is not guaranteed to reflect the current server state of the pr, it should always be considered a snapshot. This property of a `PullRequest` wasn't properly reflected in the `TestPullRequest` implementation until recently, with the explicit `TestPullRequestStore` object that represents the server side state.

In test code, instead of explicitly fetching an updated snapshot of a PR, you can access the store directly to inspect the state through `pr.store()`. Most old test code has not been updated to use this. Note that the `TestPullRequestStore` doesn't have as many convenience methods for all the data as `TestPullRequest` has, so it may be a bit clunky, at least for labels. You could always add a better accessor (in `TestIssueStore` in this case) to help with that.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckablePullRequest.java
```diff
@@ -140,9 +140,9 @@ static List<Review> filterActiveReviews(List<Review> allReviews, String targetRe
                 var reviewCorrectTarget = review.targetRef().equals(targetRef);
                 var reviewNewer = prevReview.createdAt().isBefore(review.createdAt());
 
-                if ((prevReviewCorrectTarget && reviewCorrectTarget && reviewNewer)
+                if ((!review.verdict().equals(Review.Verdict.NONE)) && ((prevReviewCorrectTarget && reviewCorrectTarget && reviewNewer)
```
erikj79, line 143: Maybe break the line for better readability before the first `&&`.

bots/pr/src/test/java/org/openjdk/skara/bots/pr/ReviewersTests.java
zhaosongzs: > An instance of `PullRequest` is not guaranteed to reflect the current server state of the pr, it should always be considered a snapshot. This property of a `PullRequest` wasn't properly reflected in the `TestPullRequest` implementation until recently, with the explicit `TestPullRequestStore` object that represents the server side state.
> 
> In test code, instead of explicitly fetching an updated snapshot of a PR, you can access the store directly to inspect the state through `pr.store()`. Most old test code has not been updated to use this. Note that the `TestPullRequestStore` doesn't have as many convenience methods for all the data as `TestPullRequest` has, so it may be a bit clunky, at least for labels. You could always add a better accessor (in `TestIssueStore` in this case) to help with that.

Got it! Thanks for explaining this to me. I will use `pr.store()` to do the test.

erikj79 (approved):

PR #1401, opened 2022-10-21
1625: Validate input from /summary command

zhaosongzs:
1.  Added validation to SummaryCommand.
2. If exceptions are thrown when the bot is evaluating the current status of the pr, the exceptions will be treated as integration blocker.

---------
### Progress

### Issue
 * [SKARA-1625](https://bugs.openjdk.org/browse/SKARA-1625): Validate input from /summary command

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1401/head:pull/1401` \
`$ git checkout pull/1401`

`$ git checkout pull/1401` \
`$ git pull https://git.openjdk.org/skara pull/1401/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1401`

`$ git pr show -t 1401`

</details>

<a href="https://git.openjdk.org/skara/pull/1401.diff">https://git.openjdk.org/skara/pull/1401.diff</a>

</details>

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
@@ -236,7 +236,8 @@ class PullRequestCheckIssueVisitor implements IssueVisitor {
     @Override
     public void visit(MessageIssue issue) {
         var message = String.join("\n", issue.commit().message());
-        addFailureMessage(issue.check(), "Incorrectly formatted commit message: " + message);
+        log.info("Incorrectly formatted commit message: " + message);
+        addFailureMessage(issue.check(), "Incorrectly formatted commit message");
     }
 
     @Override
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
@@ -236,7 +236,7 @@ class PullRequestCheckIssueVisitor implements IssueVisitor {
     @Override
     public void visit(MessageIssue issue) {
         var message = String.join("\n", issue.commit().message());
-        log.info("Incorrectly formatted commit message: " + message);
+        log.warning("Incorrectly formatted commit message: " + message);
         addFailureMessage(issue.check(), "Incorrectly formatted commit message");
     }
 
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
@@ -149,7 +149,7 @@ class PullRequestCheckIssueVisitor implements IssueVisitor {
     @Override
     public void visit(InvalidReviewersIssue e) {
         var invalid = String.join(", ", e.invalid());
-        throw new IllegalStateException("Invalid reviewers " + invalid);
+        addFailureMessage(e.check(), "Invalid reviewers " + invalid);
     }
 
     @Override
@@ -236,7 +236,7 @@ class PullRequestCheckIssueVisitor implements IssueVisitor {
     @Override
     public void visit(MessageIssue issue) {
         var message = String.join("\n", issue.commit().message());
-        throw new IllegalStateException("Incorrectly formatted commit message: " + message);
+        addFailureMessage(issue.check(), "Incorrectly formatted commit message: " + message);
     }
 
     @Override
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/SummaryCommand.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/SummaryCommand.java
@@ -28,9 +28,11 @@ import org.openjdk.skara.issuetracker.Comment;
 import java.io.PrintWriter;
 import java.nio.file.Path;
 import java.util.*;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 public class SummaryCommand implements CommandHandler {
+    private static final Pattern INVALID_SUMMARY_PATTERN = Pattern.compile("(^(Co-authored-by:)(.*))|(^(Reviewed-by:)(.*))|(^(Backport-of:)(.*))|(^[0-9]+:(.*))");
     @Override
     public void handle(PullRequestBot bot, PullRequest pr, CensusInstance censusInstance, Path scratchPath, CommandInvocation command, List<Comment> allComments, PrintWriter reply) {
         if (!command.user().equals(pr.author())) {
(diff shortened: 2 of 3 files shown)


erikj79:
> If exceptions are thrown when the bot is evaluating the current status of the pr, the exceptions will be treated as integration blocker.

I didn't mean that any exception should be an integration blocker. I just wanted the very specific case when parsing the commit message fails to be one. For everything else, we should still let the exception go through.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/SummaryCommand.java
```diff
@@ -50,17 +52,23 @@ public void handle(PullRequestBot bot, PullRequest pr, CensusInstance censusInst
...
-            var action = currentSummary.isPresent() ? "Updating existing" : "Setting";
-            if (summary.contains("\n")) {
-                reply.println(action + " summary to:\n" +
-                                      "\n" +
-                                      "```\n" +
-                                      summary +
-                                      "\n```");
+            if (!checkSummary(summary)) {
+                reply.println("\"" + summary + "\" is invalid, summary should not contain \"<BugId>:\", \"Co-authored-by:\", \"Reviewed-by:\" and \"Backport-of:\".\n" +
+                        "This Summary will be removed.");
```
erikj79, line 57: When quoting the given invalid summary please quote it in the same way as the accepting message below (multiline if needed, use back ticks). Prefix it with "Invalid summary: ". Then on the next line add the information about what cannot be included in a summary. I don't think we need the "This Summary will be removed." message on top of that.

Example:

Invalid summary:
```
foobar
12345: foo
```
A summary line cannot start with any of the following: `<issue-id>:`, `Co-authored-by:`, `Reviewed-by:`, `Backport-of:`. See [JEP 357](https://openjdk.org/jeps/357) for details.

bots/pr/src/test/java/org/openjdk/skara/bots/pr/SummaryTests.java
```diff
@@ -386,4 +386,46 @@ void singleAndMultiLineSummary(TestInfo testInfo) throws IOException {
...
+            assertLastCommentContains(pr, "summary should not contain");
+
+            pr.addComment("/summary first line\n Reviewed-by: user1");
+            TestBotRunner.runPeriodicItems(prBot);
+            assertLastCommentContains(pr, "summary should not contain");
+
+            pr.addComment("/summary Backport-of: 12434");
+            TestBotRunner.runPeriodicItems(prBot);
+            assertLastCommentContains(pr, "summary should not contain");
+
```
erikj79, line 425: Maybe add a test case for `<bugid>:`?

zhaosongzs:
> > If exceptions are thrown when the bot is evaluating the current status of the pr, the exceptions will be treated as integration blocker.
> 
> I didn't mean that any exception should be an integration blocker. I just wanted the very specific case when parsing the commit message fails to be one. For everything else, we should still let the exception go through.

Ok, my bad. I was thinking it is a little difficult to catch the specific exception. I look through the code again and think if the commit message is bad, `PullRequestCheckIssueVisitor` will throw `IllegalStateException`, so we just need to catch `IllegalStateException`, right?

bots/pr/src/main/java/org/openjdk/skara/bots/pr/SummaryCommand.java
zhaosongzs: > When quoting the given invalid summary please quote it in the same way as the accepting message below (multiline if needed, use back ticks). Prefix it with "Invalid summary: ". Then on the next line add the information about what cannot be included in a summary. I don't think we need the "This Summary will be removed." message on top of that.
> 
> Example:
> 
> Invalid summary:
> 
> ```
> foobar
> 12345: foo
> ```
> 
> A summary line cannot start with any of the following: `<issue-id>:`, `Co-authored-by:`, `Reviewed-by:`, `Backport-of:`. See [JEP 357](https://openjdk.org/jeps/357) for details.

Thanks for the suggestion! I will change it.

erikj79:
> Ok, my bad. I was thinking it is a little difficult to catch the specific exception. I look through the code again and find if the commit message is bad, `PullRequestCheckIssueVisitor` will throw `IllegalStateException`, so we just need to catch `IllegalStateException`, right?

No, that's a too generic exception, and also IMO not used correctly here. If we are to treat this specific case in a unique way we need to introduce a new exception type just for this. I would suggest `BadCommitMessageException` (extends `RuntimeException`). This becomes a contract between `PullRequestCheckIssueVisitor` and `CheckRun`, which can then catch just this exception and make it an integration blocker.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/SummaryCommand.java
```diff
@@ -50,17 +52,26 @@ public void handle(PullRequestBot bot, PullRequest pr, CensusInstance censusInst
...
-                                      "```\n" +
-                                      summary +
-                                      "\n```");
+            if (!checkSummary(summary)) {
+                reply.println("Invalid summary:\n" +
+                        "\n" +
+                        "```\n" +
+                        summary +
+                        "\n```\n" +
+                        "A summary line cannot start with any of the following: `<issue-id>:`, `Co-authored-by:`, `Reviewed-by:`, `Backport-of:`. See [JEP 357](https://openjdk.org/jeps/357) for details.");
```
erikj79, line 61: Please break this long line.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
```diff
@@ -149,7 +155,7 @@ public void visit(TooFewReviewersIssue e) {
     @Override
     public void visit(InvalidReviewersIssue e) {
         var invalid = String.join(", ", e.invalid());
-        throw new IllegalStateException("Invalid reviewers " + invalid);
+        throw new BadCommitMessageException("Invalid reviewers " + invalid);
```
erikj79, line 158: I didn't mean for this new exception to be used everywhere. We were specifically trying to avoid just issues with the commit message here.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
```diff
@@ -236,7 +242,7 @@ public void visit(WhitespaceIssue e) {
     @Override
     public void visit(MessageIssue issue) {
         var message = String.join("\n", issue.commit().message());
-        throw new IllegalStateException("Incorrectly formatted commit message: " + message);
+        throw new BadCommitMessageException("Incorrectly formatted commit message: " + message);
```
erikj79, line 245: Looking more at this, communicating through an exception seems wrong. The visitor already has a way to communicate problems through the addFailureMessage method. We should just stop throwing exception and treat this as any other "normal" failure.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
zhaosongzs: > I didn't mean for this new exception to be used everywhere. We were specifically trying to avoid just issues with the commit message here.

I was thinking about whether to change the exception type here. In my opinion, if one of the reviewers is invalid, it is meaningless for the bot to retry, we also should make the bot end.

erikj79: Hm, you are probably right there, it was just that the specific exception name didn't fit here. But, if we report failures in the normal way instead, it's all good.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
zhaosongzs: > Looking more at this, communicating through an exception seems wrong. The visitor already has a way to communicate problems through the addFailureMessage method. We should just stop throwing exception and treat this as any other "normal" failure.

Yes, this way is better. I was just trying to make bad commit message as an `integrationBlocker`. If we choose the way you mentioned, we also no longer need the `BadCommitMessageException`, right?

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
```diff
@@ -236,7 +236,7 @@ public void visit(WhitespaceIssue e) {
     @Override
     public void visit(MessageIssue issue) {
         var message = String.join("\n", issue.commit().message());
-        throw new IllegalStateException("Incorrectly formatted commit message: " + message);
+        addFailureMessage(issue.check(), "Incorrectly formatted commit message: " + message);
```
erikj79, line 239: I don't think a multiline value will format well as a failure message. The commit message is also printed in the status comment, so probably don't need to include it here. Maybe it would be a good idea to log the full message just in case.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
```diff
@@ -236,7 +236,8 @@ public void visit(WhitespaceIssue e) {
     @Override
     public void visit(MessageIssue issue) {
         var message = String.join("\n", issue.commit().message());
-        throw new IllegalStateException("Incorrectly formatted commit message: " + message);
+        log.info("Incorrectly formatted commit message: " + message);
```
erikj79, line 239: This warrants a warning level message I think. We really don't expect this to happen.

erikj79 (approved):

PR #1403, opened 2022-10-26
1450: Jcheck status message can be too big for Gitlab

zhaosongzs:
SKARA-1450 describes the problem that the comment adding to a GitLab merge request can be too big to add and leads to fail. 

To fix it, I add the function `limitBodySize()` to `addComment()` and `updateComment()` in `GitLabMergeRequest`, the function will check the size of the comment and  the comment will be cropped if it is too big.

I also found that GitHub pull request might have the same problem, so I did the same change to `GitHubPullRequest`.

Currently, I set the comment max size to 20000, and I don't know whether it is fine. I think it could not be too small because the user wants to see more information and it should not be big because if it is too big, the page loading process will be slow when the user tries to open the pr.

[Maximum length for the comment body in GitHub PR](https://github.com/orgs/community/discussions/27190)
[My test in Playground](https://github.com/openjdk/playground/pull/99)

---------
### Progress

### Issue
 * [SKARA-1450](https://bugs.openjdk.org/browse/SKARA-1450): Jcheck status message can be too big for Gitlab

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1403/head:pull/1403` \
`$ git checkout pull/1403`

`$ git checkout pull/1403` \
`$ git pull https://git.openjdk.org/skara pull/1403/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1403`

`$ git pr show -t 1403`

</details>

<a href="https://git.openjdk.org/skara/pull/1403.diff">https://git.openjdk.org/skara/pull/1403.diff</a>

</details>

### DIFF
--- a/forge/src/main/java/org/openjdk/skara/forge/github/GitHubPullRequest.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/github/GitHubPullRequest.java
@@ -46,6 +46,8 @@ public class GitHubPullRequest implements PullRequest {
 
     private List<Label> labels = null;
 
+    private static final int GITHUB_PR_COMMENT_BODY_MAX_SIZE = 20000;
+
     GitHubPullRequest(GitHubRepository repository, JSONValue jsonValue, RestRequest request) {
         this.host = (GitHubHost)repository.forge();
         this.repository = repository;
@@ -351,6 +353,7 @@ public class GitHubPullRequest implements PullRequest {
 
     @Override
     public Comment addComment(String body) {
+        body = limitBodySize(body);
         var comment = request.post("issues/" + json.get("number").toString() + "/comments")
                 .body("body", body)
                 .execute();
@@ -364,6 +367,7 @@ public class GitHubPullRequest implements PullRequest {
 
     @Override
     public Comment updateComment(String id, String body) {
+        body = limitBodySize(body);
         var comment = request.patch("issues/comments/" + id)
                              .body("body", body)
                              .onError(r -> {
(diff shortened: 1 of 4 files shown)


forge/src/main/java/org/openjdk/skara/forge/github/GitHubPullRequest.java
```diff
@@ -46,6 +46,8 @@ public class GitHubPullRequest implements PullRequest {
 
     private List<Label> labels = null;
 
+    private static final int GITHUB_PR_COMMENT_BODY_MAX_SIZE = 20000;
```
erikj79, line 49: I think we can bump this up a bit more. Maybe 64k is a reasonable number? Also consider using `_` to group long number constants into triplets for easier reading.

forge/src/main/java/org/openjdk/skara/forge/github/GitHubPullRequest.java
```diff
@@ -783,4 +787,15 @@ public Optional<Hash> findIntegratedCommitHash() {
     public Object snapshot() {
         return json;
     }
+
+    public String limitBodySize(String body) {
+        if (body.length() > GITHUB_PR_COMMENT_BODY_MAX_SIZE) {
+            return "⚠️This comment is too long, only the first "
```
erikj79, line 793: I think ending with `...` is enough to indicate that the comment has been truncated. This will only apply to comments generated by the bot. If we start such a comment with a warning like this, it will risk confusing users.

erikj79 (commented):
Comments on GitHubPullRequest also apply to GitLabMergeRequest.

forge/src/main/java/org/openjdk/skara/forge/github/GitHubPullRequest.java
zhaosongzs: > I think we can bump this up a bit more. Maybe 64k is a reasonable number? Also consider using `_` to group long number constants into triplets for easier reading.

64K would work well for both GitHub and GitLab

erikj79 (approved):

PR #1406, opened 2022-10-31
1608: Handle shenandoah tags in IssueNotifier

erikj79:
This patch aims to add support for notifying JBS when tags are added in the shenandoah-jdk8u repository. This repository is special because it's a jdk8u 'project' repo that is never going to be merged upstream. Instead it's a downstream repo that produces its own releases. Because of this it has its own fixVersion configured in .jcheck/conf and its own tag format, to keep them apart from the main JDK jdk8u tags.

The logic for matching fixVersions and tags is already a bit convoluted due to conventions that have been added over time. This repo will now add another matching rule. A typical fixVersion would be `shenandoah8u332` and a corresponding tag `shenandoah8u332-b01`. This is actually a pretty neat pattern to match for as the tag prefix is exactly the fixVersion, which is why I agreed to implement support for this. The complicating factors are that for mainline OpenJDK jdk8u we have fixVersion `openjdk8u332` and tag `jdk8u332-b01` (and for Oracle jdk8u the fixVersion is `8u332` but with the same tag format). This special handling of a 'jdk' prefix for a tag, and ignoring any prefix in the fixVersion is what complicates things.

So currently we ignore any prefix before the main version number in both a tag and fixVersion when matching them. For shenandoah-jdk8u we want to exactly match this prefix. To handle this, I've added a new configuration option for the notifier:
```
"issue": {  // existing config element
  "tag": {  // existing config element
    "matchprefix": true  // new
  }
}
```
The default value for this option is `false` to reflect the current behavior. When set true, a fixVersion and tag version must match exactly. Note though that the special handling of the tag prefix 'jdk' will still be at play due to the way OpenJDKTag currently parses a JdkVersion from a tag, and I don't dare changing it. This means that if this option is set, then a fixVersion `8u332` will match tags `jdk8u332-b01` and `8u332-b01`, but a fixVersion `jdk8u332` would not match the tag `jdk8u332-b01`. We don't currently use the 'jdk' prefix in any fixVersions in JBS, so this should be fine. I just want to make it clear.

---------
### Progress

### Issue
 * [SKARA-1608](https://bugs.openjdk.org/browse/SKARA-1608): Handle shenandoah tags in IssueNotifier

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1406/head:pull/1406` \
`$ git checkout pull/1406`

`$ git checkout pull/1406` \
`$ git pull https://git.openjdk.org/skara pull/1406/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1406`

`$ git pr show -t 1406`

</details>

<a href="https://git.openjdk.org/skara/pull/1406.diff">https://git.openjdk.org/skara/pull/1406.diff</a>

</details>

### DIFF
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
@@ -73,6 +73,11 @@ class IssueNotifier implements Notifier, PullRequestListener, RepositoryListener
     // do not need to be part of a tag to be considered a match.
     private final Set<String> tagIgnoreOpt;
 
+    // Should the prefix of a tag match the prefix of a fix version to be considered
+    // a match (except for the special tag prefix 'jdk' which will always be ignored
+    // when parsing a version from a tag).
+    private final boolean tagMatchPrefix;
+
     private final Logger log = Logger.getLogger("org.openjdk.skara.bots.notify");
 
     // Lazy loaded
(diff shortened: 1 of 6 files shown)


edvbld (approved):
This looks good, I know this is a can (swimming pool?) of worms. A future enhancement might be to break apart `IssueNotifier` into multiple cases for different repos, e.g. `Shenandoah8uIssueNotifier`, `JDKIssueNotifier`, `UpdatesIssueNotifier` just to reduce the "branchyness" of the code. This code lead to some code duplication, but it might also maker it the logic for each special case much clearer.

Thanks,
Erik

PR #1407, opened 2022-10-31
1393: PR bot fails on missing .jcheck/conf

zhaosongzs:
SKARA-1393 describes a problem related with `.jcheck/conf` missing.

In this patch, when the problem happens, the author of this pull request will get instructions about how to solve this problem.

---------
### Progress

### Issue
 * [SKARA-1393](https://bugs.openjdk.org/browse/SKARA-1393): PR bot fails on missing .jcheck/conf

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1407/head:pull/1407` \
`$ git checkout pull/1407`

`$ git checkout pull/1407` \
`$ git pull https://git.openjdk.org/skara pull/1407/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1407`

`$ git pr show -t 1407`

</details>

<a href="https://git.openjdk.org/skara/pull/1407.diff">https://git.openjdk.org/skara/pull/1407.diff</a>

</details>

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
@@ -248,7 +248,7 @@ class CheckWorkItem extends PullRequestWorkItem {
                         + "Until that is resolved, this pull request cannot be processed. Please notify the repository owner.";
                 addErrorComment(text, comments);
             } else {
-                var text = " ⚠️ @" + pr.author().username() + " The external jcheck configuration for this repository could not be resolved. "
+                var text = " ⚠️ @" + pr.author().username() + " The external jcheck configuration for this repository is invalid. "
                         + "Until that is resolved, this pull request cannot be processed. Please notify a Skara admin.";
                 addErrorComment(text, comments);
             }
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
@@ -233,10 +233,12 @@ class CheckWorkItem extends PullRequestWorkItem {
                     bot.confOverrideRepository().orElse(null), bot.confOverrideName(), bot.confOverrideRef());
         } catch (MissingJCheckConfException e) {
             if (bot.confOverrideRepository().isEmpty()) {
+                log.info("No .jcheck/conf found in repo " + bot.repo().name() + ": " + e);
                 var text = " ⚠️ @" + pr.author().username() + " No `.jcheck/conf` found in the target branch of this pull request. "
                         + "Until that is resolved, this pull request cannot be processed. Please notify the repository owner.";
                 addErrorComment(text, comments);
             } else {
+                log.info("No .jcheck/conf found in external repo " + bot.confOverrideRepository().get().name() + ": " + e);
                 var text = " ⚠️ @" + pr.author().username() + " The external jcheck configuration for this repository could not be found. "
                         + "Until that is resolved, this pull request cannot be processed. Please notify a Skara admin.";
                 addErrorComment(text, comments);
@@ -244,10 +246,12 @@ class CheckWorkItem extends PullRequestWorkItem {
             return List.of();
         } catch (InvalidJCheckConfException e) {
             if (bot.confOverrideRepository().isEmpty()) {
+                log.info("Invalid .jcheck/conf found in repo " + bot.repo().name() + ": " + e);
                 var text = " ⚠️ @" + pr.author().username() + " The `.jcheck/conf` in the target branch of this pull request is invalid. "
                         + "Until that is resolved, this pull request cannot be processed. Please notify the repository owner.";
                 addErrorComment(text, comments);
             } else {
+                log.info("Invalid .jcheck/conf found in external repo " + bot.confOverrideRepository().get().name() + ": " + e);
                 var text = " ⚠️ @" + pr.author().username() + " The external jcheck configuration for this repository is invalid. "
                         + "Until that is resolved, this pull request cannot be processed. Please notify a Skara admin.";
                 addErrorComment(text, comments);
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
@@ -230,7 +230,7 @@ class CheckWorkItem extends PullRequestWorkItem {
         var labels = new HashSet<>(pr.labelNames());
         try {
             census = CensusInstance.createCensusInstance(hostedRepositoryPool, bot.censusRepo(), bot.censusRef(), scratchPath.resolve("census"), pr,
-                    bot.confOverrideRepository().orElse(null), bot.confOverrideName(), bot.confOverrideRef()).orElseThrow();
+                    bot.confOverrideRepository().orElse(null), bot.confOverrideName(), bot.confOverrideRef());
         } catch (MissingJCheckConfException e) {
             if (bot.confOverrideRepository().isEmpty()) {
                 var text = " ⚠️ @" + pr.author().username() + " No `.jcheck/conf` found in the target branch of this pull request. "
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
@@ -233,7 +233,7 @@ class CheckWorkItem extends PullRequestWorkItem {
                     bot.confOverrideRepository().orElse(null), bot.confOverrideName(), bot.confOverrideRef()).orElseThrow();
         } catch (MissingJCheckConfException e) {
             if (bot.confOverrideRepository().isEmpty()) {
-                var text = " ⚠️ @" + pr.author().username() + " The `.jcheck/conf` in the target branch of this pull request is missing completely. "
+                var text = " ⚠️ @" + pr.author().username() + " No `.jcheck/conf` found in the target branch of this pull request. "
                         + "Until that is resolved, this pull request cannot be processed. Please notify the repository owner.";
                 addErrorComment(text, comments);
             } else {
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
@@ -224,7 +224,7 @@ class CheckWorkItem extends PullRequestWorkItem {
         // First determine if the current state of the PR has already been checked
         var seedPath = bot.seedStorage().orElse(scratchPath.resolve("seeds"));
         var hostedRepositoryPool = new HostedRepositoryPool(seedPath);
-        CensusInstance census = null;
+        CensusInstance census;
         var comments = pr.comments();
         var allReviews = pr.reviews();
         var labels = new HashSet<>(pr.labelNames());
@@ -238,7 +238,8 @@ class CheckWorkItem extends PullRequestWorkItem {
                         + "Until that is resolved, this pull request cannot be processed. Please notify the repository owner.";
                 addErrorComment(text, comments);
             } else {
-                log.info("No .jcheck/conf found in external repo " + bot.confOverrideRepository().get().name() + ": " + e);
+                log.log(Level.SEVERE, "Jcheck configuration file " + bot.confOverrideName()
+                        + " not found in external repo " + bot.confOverrideRepository().get().name(), e);
                 var text = " ⚠️ @" + pr.author().username() + " The external jcheck configuration for this repository could not be found. "
                         + "Until that is resolved, this pull request cannot be processed. Please notify a Skara admin.";
                 addErrorComment(text, comments);
@@ -251,7 +252,8 @@ class CheckWorkItem extends PullRequestWorkItem {
                         + "Until that is resolved, this pull request cannot be processed. Please notify the repository owner.";
                 addErrorComment(text, comments);
             } else {
-                log.info("Invalid .jcheck/conf found in external repo " + bot.confOverrideRepository().get().name() + ": " + e);
+                log.log(Level.SEVERE, "Invalid Jcheck configuration file " + bot.confOverrideName()
+                        + " in external repo " + bot.confOverrideRepository().get().name(), e);
                 var text = " ⚠️ @" + pr.author().username() + " The external jcheck configuration for this repository is invalid. "
                         + "Until that is resolved, this pull request cannot be processed. Please notify a Skara admin.";
                 addErrorComment(text, comments);
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
@@ -224,12 +224,36 @@ class CheckWorkItem extends PullRequestWorkItem {
         // First determine if the current state of the PR has already been checked
         var seedPath = bot.seedStorage().orElse(scratchPath.resolve("seeds"));
         var hostedRepositoryPool = new HostedRepositoryPool(seedPath);
-
-        var census = CensusInstance.createCensusInstance(hostedRepositoryPool, bot.censusRepo(), bot.censusRef(), scratchPath.resolve("census"), pr,
-                                           bot.confOverrideRepository().orElse(null), bot.confOverrideName(), bot.confOverrideRef()).orElseThrow();
+        CensusInstance census = null;
         var comments = pr.comments();
         var allReviews = pr.reviews();
         var labels = new HashSet<>(pr.labelNames());
+        try {
+            census = CensusInstance.createCensusInstance(hostedRepositoryPool, bot.censusRepo(), bot.censusRef(), scratchPath.resolve("census"), pr,
+                    bot.confOverrideRepository().orElse(null), bot.confOverrideName(), bot.confOverrideRef()).orElseThrow();
+        } catch (MissingJCheckConfException e) {
+            if (bot.confOverrideRepository().isEmpty()) {
+                var text = " ⚠️ @" + pr.author().username() + " The `.jcheck/conf` in the target branch of this pull request is missing completely. "
+                        + "Until that is resolved, this pull request cannot be processed. Please notify the repository owner.";
+                addErrorComment(text, comments);
+            } else {
+                var text = " ⚠️ @" + pr.author().username() + " The external jcheck configuration for this repository could not be resolved. "
+                        + "Until that is resolved, this pull request cannot be processed. Please notify a Skara admin.";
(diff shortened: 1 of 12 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -225,6 +225,22 @@ public Collection<WorkItem> prRun(Path scratchPath) {
         var seedPath = bot.seedStorage().orElse(scratchPath.resolve("seeds"));
         var hostedRepositoryPool = new HostedRepositoryPool(seedPath);
 
+        if (bot.confOverrideRepository().isEmpty()) {
+            //check whether this repo contains .jcheck/conf, if not, let the pr author to notify the repo owner
+            try {
+                pr.repository().fileContents(".jcheck/conf", pr.targetRef());
```
erikj79, line 231: Instead of explicitly checking for `.jcheck/conf` here through a remote call, I think it would be better to react to if the `Optional<CensusInstance>` below comes back empty or not. If it's empty and `bot.confOverrideRepository().isEmpty()` then you can print a message.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -225,6 +225,22 @@ public Collection<WorkItem> prRun(Path scratchPath) {
...
+            } catch (NoSuchElementException e) {
+                String message = "@" + pr.author().username() + " This repo doesn't contain `.jcheck/conf`,"
+                        + " please close this pr and notify the repo owner."
+                        + " After uploading `.jcheck/conf` to the target branch, you can create a new pull request.";
+                boolean messageSent = pr.comments().stream().anyMatch(comment -> comment.body().equals(message));
+                if (!messageSent) {
+                    pr.addComment(message);
+                }
```
erikj79, line 239: I think you can reuse the method `addBackportErrorComment` for posting the comment. There is nothing specific to backport about that method, so it should just be renamed to `addErrorComment`. 

For the message, I don't think we need to tell the user to close the PR. I think we should add the :warning: emoji first in the message. I would suggest something like this:

":warning: at user The target branch of this pull request is either missing `.jcheck/conf` or it is invalid. Until that is resolved, this pull request cannot be processed. Please notify the repository owner."

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
zhaosongzs: > Instead of explicitly checking for `.jcheck/conf` here through a remote call, I think it would be better to react to if the `Optional<CensusInstance>` below comes back empty or not. If it's empty and `bot.confOverrideRepository().isEmpty()` then you can print a message.

Yesterday I was wondering whether there is any possible that `Optional<CensusInstance>` is empty due to other errors. Today I read through the code again and confirm that Only `.jcheck/conf` invalid or missing will make `Optional<CensusInstance>` empty. Thx!

erikj79: Maybe it would be even better to throw an exception instead of returning Optional.empty(). We can't recover from it  in CheckWorkItem anyway, and with a typed exception we can communicate more exactly what went wrong, and we can catch exactly that exception type and write the message.

zhaosongzs: > Maybe it would be even better to throw an exception instead of returning Optional.empty(). We can't recover from it in CheckWorkItem anyway, and with a typed exception we can communicate more exactly what went wrong, and we can catch exactly that exception type and write the message.

I thought about it in the very beginning and I found that  if we choose to throw an exception, we will need to surround a lot of code with try catch because the scope of `census` is big.

erikj79: Note that there are a lot of different `CensusInstance` implementations. They aren't shared between different bot modules. The one in `bots.pr` is only used in two locations, and the other location would probably also like a quick exit on this failure.

zhaosongzs: > 

Yes, I just mean we need to surround a lot of code with try catch in `CheckWorkItem#prRun`.  If it is ok, I will change it.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/LimitedCensusInstance.java
```diff
@@ -39,7 +44,7 @@ static Optional<LimitedCensusInstance> createLimitedCensusInstance(HostedReposit
             Optional<JCheckConfiguration> configuration = jCheckConfiguration(hostedRepositoryPool,
                     repository, ref, confOverrideRepo, confOverrideName, confOverrideRef);
             if (configuration.isEmpty()) {
-                return Optional.empty();
+                throw new InvalidJCheckConfException(".jcheck/conf is invalid or missing");
```
zhaosongzs, line 47: Hi Erik, I don't know if I understood you exactly to throw the exception here.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -224,12 +224,25 @@ public Collection<WorkItem> prRun(Path scratchPath) {
...
+        } catch (InvalidJCheckConfException invalidJCheckConfException) {
+            if (bot.confOverrideRepository().isEmpty()) {
+                var text = " ⚠️ @" + pr.author().username() + " The `.jcheck/conf` in the target branch of this pull request is either invalid or missing completely. "
+                        + "Until that is resolved, this pull request cannot be processed. Please notify the repository owner.";
+                addErrorComment(text, comments);
+            } else {
+                var text = " ⚠️ @" + pr.author().username() + " The `.jcheck/conf` in the override repo is either invalid or missing completely. "
+                        + "Until that is resolved, this pull request cannot be processed. Please notify the repository owner.";
+                addErrorComment(text, comments);
+            }
```
zhaosongzs, line 243: I think this is another situation which is that the override configuration is invalid or missing

bots/pr/src/main/java/org/openjdk/skara/bots/pr/LimitedCensusInstance.java
erikj79: I think it would fit better in `configuration` as an `.orElseThrow()`. In that case it's only thrown if jcheck is missing, so the exception should be named MissingJCheckConfException.

The actual parsing of the configuration seems to only explicitly throw IllegalArgumentException. It's possible that other parse problems would result in other runtime exceptions, such as NPE. I think we should wrap the parse call with try-catch for RuntimeException and throw an `InvalidJCheckConfException` if that happens.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
erikj79: If we throw two separate exception types as suggested below, then we can provide more detailed error messages, for either a missing or an invalid jcheck conf.

In the case of the override jcheck configuration, the file can be configured to have any name. I would suggest:
"The external jcheck configuration for this repository could not be resolved. Until that is fixed, this pull request cannot be processed. Please notify a Skara admin."

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -224,12 +224,25 @@ public Collection<WorkItem> prRun(Path scratchPath) {
...
-        var census = CensusInstance.createCensusInstance(hostedRepositoryPool, bot.censusRepo(), bot.censusRef(), scratchPath.resolve("census"), pr,
-                                           bot.confOverrideRepository().orElse(null), bot.confOverrideName(), bot.confOverrideRef()).orElseThrow();
+        CensusInstance census = null;
         var comments = pr.comments();
         var allReviews = pr.reviews();
         var labels = new HashSet<>(pr.labelNames());
+        try {
+            census = CensusInstance.createCensusInstance(hostedRepositoryPool, bot.censusRepo(), bot.censusRef(), scratchPath.resolve("census"), pr,
+                    bot.confOverrideRepository().orElse(null), bot.confOverrideName(), bot.confOverrideRef()).orElseThrow();
+        } catch (InvalidJCheckConfException invalidJCheckConfException) {
```
erikj79, line 234: Please just name the caught exception variable `e` unless we need explicit names.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/LimitedCensusInstance.java
```diff
@@ -55,27 +64,43 @@ private static Namespace namespace(Census census, String hostNamespace) {
...
+            conf = Optional.of(Arrays.stream(remoteRepo.fileContents(name, ref).split("\n")).toList());
+        } catch (RuntimeException e) {
+            if (e.getMessage().contains("Request returned bad status:")) {
+                if (!e.getMessage().contains("404")) {
+                    throw e;
+                }
+            }
+            conf = Optional.empty();
+        }
+        return conf.map(JCheckConfiguration::parse);
```
zhaosongzs, line 82: First, when we try to read the `.jcheck/conf` from the remote repo, maybe the request will throw `uncheckedRestException`. Only when the status code is `404`, conf will be set to `Optional.empty()`, so that it will be treated as `MissingJcheckConfException` later.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/LimitedCensusInstance.java
```diff
@@ -55,27 +64,43 @@ private static Namespace namespace(Census census, String hostNamespace) {
...
+                configuration = configuration(confOverrideRepo,
+                        confOverrideName,
+                        confOverrideRef);
+            }
+            return configuration;
+        } catch (RuntimeException e) {
+            if (e.getMessage().contains("Request returned bad status:")) {
+                throw e;
+            }
+            throw new InvalidJCheckConfException(e);
```
zhaosongzs, line 102: Here, if we meet some `uncheckedException`, we just throw it, otherwise it will be treated as `InvalidJCheckException` as well.

erikj79 (commented):
This is starting to look pretty good.

erikj79:
One more thing. Any time we catch an `InvalidJcheckConfException` and just print a message to the user, we should also log that exception, so that an admin can more easily investigate what went wrong.

zhaosongzs:
> One more thing. Any time we catch an `InvalidJcheckConfException` and just print a message to the user, we should also log that exception, so that an admin can more easily investigate what went wrong.

Got it! Will add some logs. Thx!

bots/pr/src/main/java/org/openjdk/skara/bots/pr/LimitedCensusInstance.java
```diff
@@ -32,18 +46,15 @@ class LimitedCensusInstance {
...
-                    repository, ref, confOverrideRepo, confOverrideName, confOverrideRef);
-            if (configuration.isEmpty()) {
-                return Optional.empty();
-            }
+            JCheckConfiguration configuration = jCheckConfiguration(repository, ref, confOverrideRepo,
+                    confOverrideName, confOverrideRef).orElseThrow(MissingJCheckConfException::new);
             var census = Census.parse(repoFolder);
             var namespace = namespace(census, repository.namespace());
-            return Optional.of(new LimitedCensusInstance(census, configuration.get(), namespace));
+            return Optional.of(new LimitedCensusInstance(census, configuration, namespace));
```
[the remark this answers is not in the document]
zhaosongzs: Exactly! I will remove it.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -224,12 +224,40 @@ public Collection<WorkItem> prRun(Path scratchPath) {
...
+            census = CensusInstance.createCensusInstance(hostedRepositoryPool, bot.censusRepo(), bot.censusRef(), scratchPath.resolve("census"), pr,
+                    bot.confOverrideRepository().orElse(null), bot.confOverrideName(), bot.confOverrideRef());
+        } catch (MissingJCheckConfException e) {
+            if (bot.confOverrideRepository().isEmpty()) {
+                log.info("No .jcheck/conf found in repo " + bot.repo().name() + ": " + e);
+                var text = " ⚠️ @" + pr.author().username() + " No `.jcheck/conf` found in the target branch of this pull request. "
+                        + "Until that is resolved, this pull request cannot be processed. Please notify the repository owner.";
+                addErrorComment(text, comments);
+            } else {
+                log.info("No .jcheck/conf found in external repo " + bot.confOverrideRepository().get().name() + ": " + e);
```
erikj79, line 241: The file isn't necessarily called `.jcheck/conf`, the name is given in `bot.confOverrideName()`.

The log should be at SEVERE level and the throwable should be included. To achieve this, you need to use the `log` method. By making it SEVERE, we will get admin notifications about this.

```suggestion
                log.log(Level.SEVERE, "Jcheck configuration file " + bot.confOverrideName() + " not found in external repo " + bot.confOverrideRepository().get().name(), e);
```

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -224,12 +224,40 @@ public Collection<WorkItem> prRun(Path scratchPath) {
         // First determine if the current state of the PR has already been checked
         var seedPath = bot.seedStorage().orElse(scratchPath.resolve("seeds"));
         var hostedRepositoryPool = new HostedRepositoryPool(seedPath);
-
-        var census = CensusInstance.createCensusInstance(hostedRepositoryPool, bot.censusRepo(), bot.censusRef(), scratchPath.resolve("census"), pr,
-                                           bot.confOverrideRepository().orElse(null), bot.confOverrideName(), bot.confOverrideRef()).orElseThrow();
+        CensusInstance census = null;
```
erikj79, line 227: I don't think you need to assign null here as all exception paths will return.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
```diff
@@ -196,8 +196,13 @@ public Collection<WorkItem> prRun(Path scratchPath) {
         var seedPath = bot.seedStorage().orElse(scratchPath.resolve("seeds"));
         var hostedRepositoryPool = new HostedRepositoryPool(seedPath);
 
-        var census = CensusInstance.createCensusInstance(hostedRepositoryPool, bot.censusRepo(), bot.censusRef(), scratchPath.resolve("census"), pr,
-                                           bot.confOverrideRepository().orElse(null), bot.confOverrideName(), bot.confOverrideRef()).orElseThrow();
+        CensusInstance census = null;
```
erikj79, line 199: I don't think you need to assign null here.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/LimitedCensusInstance.java
```diff
@@ -55,27 +66,41 @@ private static Namespace namespace(Census census, String hostNamespace) {
...
     }
 
-    private static Optional<JCheckConfiguration> configuration(HostedRepositoryPool hostedRepositoryPool, HostedRepository remoteRepo, String name, String ref) throws IOException {
-        return hostedRepositoryPool.lines(remoteRepo, Path.of(name), ref).map(JCheckConfiguration::parse);
+    private static Optional<JCheckConfiguration> configuration(HostedRepository remoteRepo, String name, String ref) {
+        Optional<List<String>> conf = Optional.empty();
+        try {
+            conf = Optional.of(Arrays.stream(remoteRepo.fileContents(name, ref).split("\n")).toList());
+        } catch (NoSuchElementException ignored) {
+            // NoSuchElementException will only work for tests
```
erikj79, line 77: I'm not sure I understand this part.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
zhaosongzs: Got it, I will change it.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/LimitedCensusInstance.java
erikj79: Can we change `TestHostedRepository` to behave more like other remote repo implementations in this regard? It could throw UncheckedRestException with 404 if the file doesn't exist.

zhaosongzs: Yes, it's a good idea!

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -224,12 +224,42 @@ public Collection<WorkItem> prRun(Path scratchPath) {
...
+        CensusInstance census;
         var comments = pr.comments();
         var allReviews = pr.reviews();
         var labels = new HashSet<>(pr.labelNames());
+        try {
+            census = CensusInstance.createCensusInstance(hostedRepositoryPool, bot.censusRepo(), bot.censusRef(), scratchPath.resolve("census"), pr,
+                    bot.confOverrideRepository().orElse(null), bot.confOverrideName(), bot.confOverrideRef());
+        } catch (MissingJCheckConfException e) {
+            if (bot.confOverrideRepository().isEmpty()) {
+                log.info("No .jcheck/conf found in repo " + bot.repo().name() + ": " + e);
```
erikj79, line 236: I think we want SEVERE log messages with the exception included for all the cases here.

zhaosongzs: > I think we want SEVERE log messages with the exception included for all the cases here.

I thought if the jcheck configuration file is not overridden, we just need to tell the user to contact the repo owner and only the repo owner could fix this problem, admin could not. 

For the case that  jcheck configuration file is  overridden, we need to notify the admin because admin could fix this problem.

But it is also reasonable to notify admin in the first case, maybe the admin could contact the repo owner too.

erikj79: Your reasoning is valid, but I think it's better that we get notified as admins too. We need to help repo owners with getting things right, also logging the exception gives us access to the stack trace which can help diagnose further.

zhaosongzs: > Your reasoning is valid, but I think it's better that we get notified as admins too. We need to help repo owners with getting things right, also logging the exception gives us access to the stack trace which can help diagnose further.

Yes, I will change the log!

erikj79 (approved):

PR #1409, opened 2022-11-01
1663: Mark integrated Pull Requests as properly merged in their repositories

TheShermanTanker:
Skara currently closes integrated Pull Requests, since the actual integration is done internally and then pushed to the repository separately. This makes every integrated request always look like it was closed to an outside observer, forcing the use of a special integrated flag to distinguish between integrated and closed, or rejected and closed. On top of that, it just looks awful in the interface all around (and is also not included in the accepted Pull Request count by GitHub for the committer, which is frustrating for newer contributors to a surprising extent). In the [willful absence of a reply from the GitHub team to allow marking a pull request as merged through a flag](https://github.com/orgs/community/discussions/12437), and [GitLab being unlikely to make a similar change either](https://gitlab.com/gitlab-org/gitlab/-/issues/16701), we can look at implementing this as an integration feature in Skara itself instead.

Every pull has a corresponding pre-integration branch created at the time it was made, and this special branch is marked for deletion when the pull is integrated, and initially it was thought to merge the Pull Request into these corresponding pr/ branches just before their deletion but that proved to present too many challenges to be viable. Instead a related approach of enhancing the Pull Request bots with their own special integration branches can be taken, as well as providing a flexible utility method for extracting the actual branch the Pull Request was integrated into, to address the concern of losing information on what the actual Pull Request target branch was.

https://gitlab.com/gitlab-org/gitlab/-/blob/master/app/models/merge_request.rb#L172

---------
### Progress

### Issue
 * [SKARA-1663](https://bugs.openjdk.org/browse/SKARA-1663): Mark integrated Pull Requests as properly merged in their repositories

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1409/head:pull/1409` \
`$ git checkout pull/1409`

`$ git checkout pull/1409` \
`$ git pull https://git.openjdk.org/skara.git pull/1409/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1409`

`$ git pr show -t 1409`

</details>

<a href="https://git.openjdk.org/skara/pull/1409.diff">https://git.openjdk.org/skara/pull/1409.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1409#issuecomment-1301652127)

### DIFF
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/prbranch/PullRequestBranchNotifier.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/prbranch/PullRequestBranchNotifier.java
@@ -77,7 +77,7 @@ public class PullRequestBranchNotifier implements Notifier, PullRequestListener 
 
     @Override
     public void onStateChange(PullRequest pr, Path scratchPath, Issue.State oldState) {
-        if (pr.state() == Issue.State.CLOSED) {
+        if (pr.state() != Issue.State.OPEN) {
             var retargetedDependencies = PreIntegrations.retargetDependencies(pr);
             deleteBranch(pr);
             if (pr.labelNames().contains("integrated")) {
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/IntegrateCommand.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/IntegrateCommand.java
@@ -24,6 +24,8 @@ package org.openjdk.skara.bots.pr;
 
 import org.openjdk.skara.forge.*;
 import org.openjdk.skara.issuetracker.Comment;
+import org.openjdk.skara.vcs.Branch;
+import org.openjdk.skara.vcs.Commit;
 import org.openjdk.skara.vcs.Hash;
 import org.openjdk.skara.vcs.Repository;
 
@@ -162,7 +164,7 @@ public class IntegrateCommand implements CommandHandler {
 
         Optional<Hash> prepushHash = checkForPrePushHash(bot, pr, scratchPath, allComments, "integrate");
         if (prepushHash.isPresent()) {
-            markIntegratedAndClosed(pr, prepushHash.get(), reply);
+            markIntegratedAndMerged(bot, scratchPath, pr, prepushHash.get(), reply);
             return;
         }
 
@@ -190,7 +192,28 @@ public class IntegrateCommand implements CommandHandler {
             // Now that we have the integration lock, refresh the PR metadata
             pr = pr.repository().pullRequest(pr.id());
 
-            Repository localRepo = materializeLocalRepo(bot, pr, scratchPath, "integrate");
+            Repository localRepo = new HostedRepositoryPool(bot.seedStorage().orElse(scratchPath.resolve("seeds")))
+                    .materialize(pr.repository(), scratchPath.resolve(pr.headHash().hex()));
+            localRepo.fetch(pr.repository().url(), pr.targetRef(), true);
+            localRepo.checkout(new Branch(pr.targetRef()), false);
+
+            // See markIntegratedAndMerged for the logic for rogue mark as merged handling
+            final List<Commit> commits = localRepo.commits(2).asList();
+            commits.forEach(commit -> log.fine(commit.toString()));
+            final Commit currentMarkAsMergedCommit =
+                    isMarkAsMergeCommit(commits.get(0)) ? commits.get(0) : null;
+            final Commit lastMarkAsMergedCommit =
+                    (commits.size() == 2 ? (isMarkAsMergeCommit(commits.get(1)) ? commits.get(1) : null) : null);
+
+            if (lastMarkAsMergedCommit != null) {
+                localRepo.reset(lastMarkAsMergedCommit.parents().get(0), true);
+                localRepo.push(lastMarkAsMergedCommit.parents().get(0), pr.repository().url(), pr.targetRef(), true);
+            } else if (currentMarkAsMergedCommit != null) {
+                localRepo.reset(currentMarkAsMergedCommit.parents().get(0), true);
+                localRepo.push(currentMarkAsMergedCommit.parents().get(0), pr.repository().url(), pr.targetRef(), true);
+            }
+
+            localRepo = materializeLocalRepo(bot, pr, scratchPath, "integrate");
             var checkablePr = new CheckablePullRequest(pr, localRepo, bot.ignoreStaleReviews(),
                                                        bot.confOverrideRepository().orElse(null),
                                                        bot.confOverrideName(),
@@ -241,7 +264,7 @@ public class IntegrateCommand implements CommandHandler {
                 var amendedHash = checkablePr.amendManualReviewers(localHash, censusInstance.namespace(), original);
                 addPrePushComment(pr, amendedHash, rebaseMessage.toString());
                 localRepo.push(amendedHash, pr.repository().url(), pr.targetRef());
-                markIntegratedAndClosed(pr, amendedHash, reply);
+                markIntegratedAndMerged(bot, scratchPath, pr, amendedHash, reply);
             } else {
                 reply.print("Warning! Your commit did not result in any changes! ");
                 reply.println("No push attempt will be made.");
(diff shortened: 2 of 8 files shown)


TheShermanTanker:
I've only just realized that there are tests that require careful change whenever the integration marking code is modified. The existing integration sequence hasn't been touched whatsoever but there is new logic executed before it that I can't figure out whether the corresponding test code needs to be updated. Will add that as a task for now

erikj79:
First of all, for a change as big as this, a bug is definitely needed. It should really be filed first and explain what and why you intend to change.

This is an interesting idea, but I'm not sure its worth the tradeoff of losing the information about which branch a PR was integrated into. We already have multi branch development happening in some repos and more of that is coming. 

At first glance this also looks brittle. You need to understand that the bots are designed to be interruptible at any point in time, restarted and being able to continue. The integration logic and order is carefully crafted to function this way. As an example, assuming that the second to last target targetRef change is the one we actually integrated into could be false if someone else was changing that around the time of integration. There are a lot of corner cases that would need careful testing before this change could be accepted, and once it went live, I would expect at least some fallout.

We are currently blocked from using pr/X branches in GitLab due to [SKARA-1173](https://bugs.openjdk.org/browse/SKARA-1173), so this proposed change would be blocked until that is resolved regardless.

TheShermanTanker:
> First of all, for a change as big as this, a bug is definitely needed. It should really be filed first and explain what and why you intend to change.
> 
> This is an interesting idea, but I'm not sure its worth the tradeoff of losing the information about which branch a PR was integrated into. We already have multi branch development happening in some repos and more of that is coming.
> 
> At first glance this also looks brittle. You need to understand that the bots are designed to be interruptible at any point in time, restarted and being able to continue. The integration logic and order is carefully crafted to function this way. As an example, assuming that the second to last target targetRef change is the one we actually integrated into could be false if someone else was changing that around the time of integration. There are a lot of corner cases that would need careful testing before this change could be accepted, and once it went live, I would expect at least some fallout.

My mistake- I forgot to create the issue and enable the test workflows, both have now been done.

I've since reverted the changes to PreIntegrations, which should resolve the problem of concurrent changes to the targetRef in question. The only other solution I can think of though is saving the targetRef before the pr/ branch swap happens, and then immediately resetting it back to the saved instance after the merge in `markIntegratedAndMerged` (and if really required, also printing the branch the Pull Request was integrated into in the "pushed as commit" message). I unfortunately currently have no way of testing if GitHub even allows doing this however, and documentation doesn't really seem to mention anything about this either

> We are currently blocked from using pr/X branches in GitLab due to [SKARA-1173](https://bugs.openjdk.org/browse/SKARA-1173), so this proposed change would be blocked until that is resolved regardless.

That's a bit of a bummer, though maybe that can be circumvented with a few `instanceof GitHubPullRequest`s? The change is, after all, only actually implemented for GitHub

erikj79:
> That's a bit of a bummer, though maybe that can be circumvented with a few instanceof GitHubPullRequests? The change is, after all, only actually implemented for GitHub

I don't understand what you mean by the change only being implemented for GitHub. We strive to keep Skara feature equivalent on GitHub and GitLab.

The pr/X branch feature is optional, so those branches are currently not present in all repositories that the PR bot operates on. This configuration is done in the notifier bot, as that is currently the only bot that needs to know about it, so the PR bot does not know if the feature is active. It would have to check for the presence of such a branch at integration time, and we would have to trust that no other branches with such names ever show up in repos where we haven't activated the pr/X feature.

I don't think the proposed strategy of using the pr/X branches to trick the forge into marking a PR as "merged" is viable for us. To me, the tradeoff between extra complications and the potential gain isn't worth it.

TheShermanTanker:
> The pr/X branch feature is optional, so those branches are currently not present in all repositories that the PR bot operates on.

My mistake, I misunderstood the original comment, I didn't realize that the PR bot had no way of telling if the feature was active on the main repos

dholmes-ora:
@TheShermanTanker  this PR is generating a lot of noise on the mailing list. It appears this is still a work-in-progress so perhaps it would be better to return to a draft state until you think it is ready. Thanks.

TheShermanTanker:
Oops, didn't realize I forgot to mark it as a draft, so sorry for the spam - That said I'm pretty much done, I don't really know what else I can do here without any guidance from further reviews

magicus:
Is the core issue here if this is even a good idea? My understanding is that Erik is highly skeptical towards the entire approach, and that a specific code review is not really relevant before it can be decided if the problem this PR attempts to solve is worth the possible risk it introduces.

I think that @TheShermanTanker could perhaps make a better argument, like, concise and effective, in explaining why this would be a good thing.

And maybe @erikj79 could be more specific about what worries you, and if there are any way the proposed change could be made safe enough. Or, as the Skara Team lead, you need to declare that this PR will not be accepted, regardless of any changes.

TheShermanTanker:
I'm fairly certain the primary concern is that switching to merge integrated PRs into their dependency branches is highly disruptive and creates problems throughout Skara's codebase (this PR no longer does that as of now), and the secondary concern is that which branch the PR was integrated into will end up being lost when implemented this way. I'm still trying to conduct tests on my end, so it will take me a while to get back to you on the comparison on pros and cons of this change however

TheShermanTanker:
An early look at how integrated Pull Requests look like after this change: https://github.com/Elshout/Sandbox/pull/2

I will definitely need more time to work on this to make it run better though, currently it's a bit of a mess

TheShermanTanker:
Improved Version: https://github.com/Elshout/Sandbox/pull/1

altrisi:
Is there any reason why Skara doesn't use (or emulate in a way that is detected if there's no direct way/endpoint) what GitHub does when using the "Rebase and Merge" button?

That marks pull requests as merged unless it's changed recently, without making the rebase directly in the PR branch.

altrisi:
I'm talking about the `Rebase and Merge` option on the dropdown when manually merging a PR from the web UI, just wondering if there's a reason not to use that or something similar in Skara. I don't know if it's exposed to apps, or if Skara has more requirements that feature doesn't meet though.

altrisi:
Sorry I've said it wrong, I was referring to Squash and merge this entire time, for some reason confused it with rebase.

erikj79:
> Is there any reason why Skara doesn't use (or emulate in a way that is detected if there's no direct way/endpoint) what GitHub does when using the "Rebase and Merge" button?
> 
> That marks pull requests as merged unless it's changed recently, without making the rebase directly in the PR branch.
> Sorry I've said it wrong, I was referring to Squash and merge this entire time, for some reason confused it with rebase.

An important design choice in Skara was to make sure it was compatible with at least 2 different hosting providers (currently github and gitlab) so that we did not build in a hard dependency on a single provider. This means we can't lean too heavily on functionality specific to any one provider, but rather prefer to implement general solutions that can be used with both/all. 

I can't say if there is anything in particular that the GitHub API of today is missing that would hinder us from using it for squash merging PRs, at least not without spending a considerable amount of time investigating. I'm pretty sure the original implementors of Skara found that the current strategy was necessary to support the OpenJDK process on both GitLab and GitHub at the time it was implemented, and I trust their judgement.

TheShermanTanker:
> I can't say if there is anything in particular that the GitHub API of today is missing that would hinder us from using it for squash merging PRs, at least not without spending a considerable amount of time investigating.

From my time experimenting with the APIs of both hosts, the only thing really missing is the ability to set the committer and author of the actual Pull Request, like we currently do in `CheckablePullRequest`. Otherwise quite a lot of the existing logic could (with substantial effort) be migrated to use the Pull Request API instead of the current strategy of creating a commit on our own local repositories before pushing to remote ("commit_title" and "commit_message" can already be set from the API for instance, same with GitLab). A rather extreme option to circumvent this issue if we really wanted to utilize the host's merging abilities as altrisi suggests would be giving Skara the ability to force the Pull Request itself (the branch that contains its commits in essence) to allow edits from the pr bot (both GitHub and GitLab support doing so) and pushing the commit from our own local repo into the PR instead, before then merging it through the API as if from the UI as per normal

TheShermanTanker:
> I can't say if there is anything in particular that the GitHub API of today is missing that would hinder us from using it for squash merging PRs, at least not without spending a considerable amount of time investigating.

From my time experimenting with the APIs of both hosts, the only thing really missing is the ability to set the committer and author of the actual Pull Request, like we currently do in `CheckablePullRequest`. Otherwise quite a lot of the existing logic could (with substantial effort) be migrated to use the Pull Request API instead of the current strategy of creating a commit on our own local repositories before pushing to remote ("commit_title" and "commit_message" can already be set from the API for instance, same with GitLab). A rather extreme option to circumvent this issue if we really wanted to utilize the host's merging abilities as altrisi suggests would be giving Skara the ability to force the Pull Request itself (the branch that contains its commits in essence) to allow edits from the pr bot (both GitHub and GitLab support doing so) and pushing the commit from our own local repo into the PR instead, before then merging it through the API as if from the UI as per normal

(Side note: Doing it this way would also result in having to make messy decisions on what to do with the existing prepush comment system and the handling for the "Pushed as commit" comments, since those would no longer be accurate in such a scenario)

PR #1411, opened 2022-11-02
1659: Fix polling and retries in testinfo bot

erikj79:
The testinfo bot monitors "checks" in the source repository of PRs and copies over the results to the PR itself. This makes it easier for PR reviewers to see results of e.g. GitHub actions.

Since this bot needs to react to things that aren't part of the PR, it needs some kind of periodic polling of PRs where new check updates are likely. This is currently handled by the TestInfoBotWorkItem calling back with an "expiration time" for a retry, depending on the conclusion of the current run. The idea seems to be that a new WorkItem should not be scheduled until after this expiration time. This isn't exactly how it ends up working however. The big issue here is that if no callback is made, the bot will happily schedule another WorkItem without delay in the next call to getPeriodicItems. There is also a race between the execution of getPeriodicItems and the execution of a WorkItem. If the WorkItem isn't calling back with an expiration in time for the next getPeriodicItems, then a new WorkItem will also be scheduled without delay.

This patch attempts to solve this by using the `PullRequestPoller` for getting updated PRs. For any PR that hasn't been updated, the only way to get scheduled is if a WorkItem is calling back with a "recheckAt" time. The poller has this functionality built in already with the retry with a date feature. This will change the bot to recheck some PRs more often (every time they are touched by a user or another bot), but it will also stop rechecking certain PRs constantly.

I'm also changing the how and when the `TestInfoBotWorkItem` calls back with a recheckAt. I'm making sure each category has a built in termination condition, to avoid ending up with an endless loop of rechecking every time interval. This is either if a PR is closed, or if checks are still running, for max of 24h since the PR was touched.

The test fix was a perplexing problem to solve. The old test setup created "checks" on a draft PR instead of a repository, which confused me. The TestInfoBot copies checks from a repository to a PR so that is what we are supposed to test. The reason for this was that `TestHostedRepository` didn't have support for having checks itself, instead the `checks` method just returned all checks from all PRs associated with the TestHost. As far as I can tell, this was just an unnecessary simplification of the test classes that made implementing a test for the `TestInfoBot` more complicated than necessary. To make things even weirder, by having `TestHostedRepository::checks` return all checks from all PRs, depending on which order the "draft" PR and the real PR were processed by TestInfoBot, they would get each others checks copied repeatedly. The old test happened to "work" because the right PR was processed first, but with my new implementation, the processing order happened to be reversed. My fix was to add a checks field on `TestHostedRepository` and put the checks there in the test setups instead.

---------
### Progress

### Issue
 * [SKARA-1659](https://bugs.openjdk.org/browse/SKARA-1659): Fix polling and retries in testinfo bot

### Reviewers
 * [Erik Helin](https://openjdk.org/census#ehelin) (@edvbld - **Reviewer**) ⚠️ Review applies to [4cc4ab09](https://git.openjdk.org/skara/pull/1411/files/4cc4ab093486c013740aaf1d3ef0808dab8b4dbe)
 * [Magnus Ihse Bursie](https://openjdk.org/census#ihse) (@magicus - **Reviewer**) ⚠️ Review applies to [8485b360](https://git.openjdk.org/skara/pull/1411/files/8485b360b41e53c4cdaedf038ce7122a922d94ba)
 * [Zhao Song](https://openjdk.org/census#zsong) (@zhaosongzs - Author)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1411/head:pull/1411` \
`$ git checkout pull/1411`

`$ git checkout pull/1411` \
`$ git pull https://git.openjdk.org/skara pull/1411/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1411`

`$ git pr show -t 1411`

</details>

<a href="https://git.openjdk.org/skara/pull/1411.diff">https://git.openjdk.org/skara/pull/1411.diff</a>

</details>

### DIFF
--- a/bots/testinfo/src/main/java/org/openjdk/skara/bots/testinfo/TestInfoBot.java
+++ b/bots/testinfo/src/main/java/org/openjdk/skara/bots/testinfo/TestInfoBot.java
@@ -43,6 +43,13 @@ import java.util.logging.Logger;
  * TestInfoBotWorkItem calls back to the bot with a re-check request which
  * causes the bot to submit that PR again after the specified amount of time,
  * or earlier if another change the PR has been detected.
+ * <p>
+ * Note that if there is a check update, this will cause an update to the PR
+ * that the next call to updatedPullRequests, and subsequently getPeriodicItems
+ * will also include. So as long as there are check updates, there will
+ * essentially be a series of rechecks with only the getPeriodicItem call delay
+ * until no update was found, at which point the bot would fall back to adding
+ * the retry interval again.
  */
 public class TestInfoBot implements Bot {
     private final HostedRepository repo;
@@ -55,12 +62,11 @@ public class TestInfoBot implements Bot {
 
     @Override
     public List<WorkItem> getPeriodicItems() {
-        var now = Instant.now();
         var prs = poller.updatedPullRequests();
         return prs.stream()
                 .filter(pr -> pr.sourceRepository().isPresent())
                 .map(pr -> (WorkItem) new TestInfoBotWorkItem(pr,
-                        delay -> poller.retryPullRequest(pr, now.plus(delay))))
+                        delay -> poller.retryPullRequest(pr, Instant.now().plus(delay))))
                 .toList();
     }
 
--- a/bots/testinfo/src/main/java/org/openjdk/skara/bots/testinfo/TestInfoBot.java
+++ b/bots/testinfo/src/main/java/org/openjdk/skara/bots/testinfo/TestInfoBot.java
@@ -33,12 +33,11 @@ import java.util.logging.Logger;
 
 public class TestInfoBot implements Bot {
     private final HostedRepository repo;
-    private final Map<String, Instant> expirations = new ConcurrentHashMap<>();
-
-    private static final Logger log = Logger.getLogger("org.openjdk.skara.bots");
+    private final PullRequestPoller poller;
 
     TestInfoBot(HostedRepository repo) {
         this.repo = repo;
+        this.poller = new PullRequestPoller(repo, true);
     }
 
     private String pullRequestToKey(PullRequest pr) {
@@ -47,24 +46,12 @@ public class TestInfoBot implements Bot {
 
     @Override
     public List<WorkItem> getPeriodicItems() {
-        var prs = repo.pullRequestsAfter(ZonedDateTime.now().minus(Duration.ofDays(1)));
-        var ret = new ArrayList<WorkItem>();
-        for (var pr : prs) {
-            if (pr.sourceRepository().isEmpty()) {
-                continue;
-            }
-
-            var expirationKey = pullRequestToKey(pr);
-            if (expirations.containsKey(expirationKey)) {
-                var expiresAt = expirations.get(expirationKey);
-                if (expiresAt.isAfter(Instant.now())) {
-                    continue;
-                }
-            }
-
-            ret.add(new TestInfoBotWorkItem(pr, expiresIn -> expirations.put(expirationKey, Instant.now().plus(expiresIn))));
-        }
-        return ret;
+        var prs = poller.updatedPullRequests();
+        return prs.stream()
+                .filter(pr -> pr.sourceRepository().isPresent())
+                .map(pr -> (WorkItem) new TestInfoBotWorkItem(pr,
+                        recheckIn -> poller.retryPullRequest(pr, Instant.now().plus(recheckIn))))
+                .toList();
     }
 
     @Override
--- a/bots/testinfo/src/main/java/org/openjdk/skara/bots/testinfo/TestInfoBot.java
+++ b/bots/testinfo/src/main/java/org/openjdk/skara/bots/testinfo/TestInfoBot.java
@@ -55,11 +55,12 @@ public class TestInfoBot implements Bot {
 
     @Override
     public List<WorkItem> getPeriodicItems() {
+        var now = Instant.now();
         var prs = poller.updatedPullRequests();
         return prs.stream()
                 .filter(pr -> pr.sourceRepository().isPresent())
                 .map(pr -> (WorkItem) new TestInfoBotWorkItem(pr,
-                        recheckIn -> poller.retryPullRequest(pr, Instant.now().plus(recheckIn))))
+                        delay -> poller.retryPullRequest(pr, now.plus(delay))))
                 .toList();
     }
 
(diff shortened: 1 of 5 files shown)


edvbld (approved):
Looks much better. just a minor suggestion. I would also have renamed `recheckIn` to `retry`, but that is just my personal (a bit more terse) style.

Thanks,
Erik

erikj79:
> Looks much better. just a minor suggestion. I would also have renamed `recheckIn` to `retry`, but that is just my personal (a bit more terse) style.

I didn't want to use the word "retry" as at least to me it implies that the previous attempt failed. This is an expected need for

zhaosongzs:
The new `TestHostedRepository#allChecks` is much more intuitive and understandable for me!

zhaosongzs:
Hi Erik, 

You mentioned 

> For any PR that hasn't been updated, the only way to get scheduled is if a WorkItem is calling back with a "recheckAt" time.

And I am thinking that we will update checks in the end of `TestInfoBotWorkItem#run`, so whether a pr will be considered as 'updated' if the checks are updated. And if so, I think in the next call of `getPeriodicItems`, the bot will also schedule another `WorkItem`.

zhaosongzs:
I just made more tests and found the checks will always be updated and I thought it's strange.

After that , I found a typo at line149 in `TestInfoBotWorkItem#run`
```
 if ((current.status() != check.status()) ||
                        (!current.summary().equals(check.summary())) ||
                        (!current.title().equals(check.summary()))) {
```

It should not be `!current.title().equals(check.summary()`

bots/testinfo/src/main/java/org/openjdk/skara/bots/testinfo/TestInfoBot.java
```diff
@@ -31,40 +31,37 @@
...
-
-            ret.add(new TestInfoBotWorkItem(pr, expiresIn -> expirations.put(expirationKey, Instant.now().plus(expiresIn))));
-        }
-        return ret;
+        var now = Instant.now();
+        var prs = poller.updatedPullRequests();
+        return prs.stream()
+                .filter(pr -> pr.sourceRepository().isPresent())
+                .map(pr -> (WorkItem) new TestInfoBotWorkItem(pr,
+                        delay -> poller.retryPullRequest(pr, now.plus(delay))))
```
magicus, line 63: Maybe I'm getting lost here, but this feels wrong to me. The `now` is computed as the now of the call to `getPeriodicItems()`, but it is used in the lambda for the `retry` operation of `TestInfoBotWorkItem`. But that retry should happen `delay` time after it is evaluated, right? 

It might not have much practical impact, but it just looks odd to me, like mixing different time streams.

magicus (approved):
Basically looks good to me (with the exception of the `now` issue).

bots/testinfo/src/main/java/org/openjdk/skara/bots/testinfo/TestInfoBot.java
magicus: To expand a bit: the `now.plus(delay)` will be evaluated in the `run` method of the work item, but then the result will be offset not from the time of `run` but from the time of the call to this method. What would happen if it, for some reason, would take more than 2 minutes between these two events, and we'd get a time in the past. Will it be retried immediately, or will it cause trouble with the poller? 

And also it feel conceptually wrong; the retry time feels like it should be specified from when the decision is made for the next check.

erikj79:
> I just made more tests and found the checks will always be updated and I thought it's strange.
> 
> After that , I found a typo at line149 in `TestInfoBotWorkItem#run`
> 
> ```
>  if ((current.status() != check.status()) ||
>                         (!current.summary().equals(check.summary())) ||
>                         (!current.title().equals(check.summary()))) {
> ```
> 
> It should not be `!current.title().equals(check.summary()`

Good catch, fixing that.

erikj79:
> Hi Erik,
> 
> You mentioned
> 
> > For any PR that hasn't been updated, the only way to get scheduled is if a WorkItem is calling back with a "recheckAt" time.
> 
> And I am thinking that sometimes we will update checks in the end of `TestInfoBotWorkItem#run`, so whether a pr will be considered as 'updated' if the checks are updated. And if so, I think in the next call of `getPeriodicItems`, the bot will also schedule another `TestInfoBotWorkItem`.
> 
> Maybe it's not a big deal. Once the checks are not updated., it will stop.

You are correct, and this is actually a behavior exhibited by some other bots as well. A WorkItem modifying the PR will trigger another round by itself. In certain cases (in other bots) I believe this is intentional, as a WorkItem will return early on first modification and then let the next round do further changes. In this case, it's not really something I wanted, but it also doesn't hurt much. It could even be considered a feature that after a check update, we quickly check again a single time in case checks update quickly, then we fall off to the 2 minute recheck interval.

All that said, I should probably add some comment acknowledging this behavior so that it's more obvious to someone reading the code.

bots/testinfo/src/main/java/org/openjdk/skara/bots/testinfo/TestInfoBot.java
erikj79: You are right, that is odd, it really should be X minutes after it was evaluated, not after the `getPeriodicItems` call. When bots are busy, the time difference can easily be much longer than 2 minutes, and in that case, effectively ignoring the delay is not helping in reducing the load.

The poller should handle past time fine though, it will basically be equivalent to a retry without a time component.

zhaosongzs:
> > Hi Erik,
> > You mentioned
> > > For any PR that hasn't been updated, the only way to get scheduled is if a WorkItem is calling back with a "recheckAt" time.
> > 
> > 
> > And I am thinking that sometimes we will update checks in the end of `TestInfoBotWorkItem#run`, so whether a pr will be considered as 'updated' if the checks are updated. And if so, I think in the next call of `getPeriodicItems`, the bot will also schedule another `TestInfoBotWorkItem`.
> > Maybe it's not a big deal. Once the checks are not updated., it will stop.
> 
> You are correct, and this is actually a behavior exhibited by some other bots as well. A WorkItem modifying the PR will trigger another round by itself. In certain cases (in other bots) I believe this is intentional, as a WorkItem will return early on first modification and then let the next round do further changes. In this case, it's not really something I wanted, but it also doesn't hurt much. It could even be considered a feature that after a check update, we quickly check again a single time in case checks update quickly, then we fall off to the 2 minute recheck interval.
> 
> All that said, I should probably add some comment acknowledging this behavior so that it's more obvious to someone reading the code.

Make sense to me!

zhaosongzs (approved):
I think it's very good now!

PR #1412, opened 2022-11-02
1594: Convert rest of bots polling PRs and Issues to use new pollers

erikj79:
The new PullRequestPoller has now been in use in the PullRequestBot for a while, and hopefully most of the kinks have been ironed out. I think it's time to convert the rest of the bots that poll PRs (and the CSRIssueBot that polls Issues) to use the new pollers. 

The conversion is pretty straight forward. In the case of the CSRBots, they were missing the retry callback, so I added that functionality in line with other bots.

The new factory tests needed to be modified slightly, as the PullRequestPoller calls to the Forge/Host object of a repository during initialization, so we need a TestHost instance there instead of `null`.

I deliberately left out two bots here, the JEPBot and the TestInfoBot. The TestInfoBot is handled separately already in #1411. For the JEPBot, I think it needs a general overhaul. In the end it should function more or less the same as the CSRBot, but I'm still not sure that the CSRBot is where I want it to be.

---------
### Progress

### Issue
 * [SKARA-1594](https://bugs.openjdk.org/browse/SKARA-1594): Convert rest of bots polling PRs and Issues to use new pollers

### Reviewers
 * [Erik Helin](https://openjdk.org/census#ehelin) (@edvbld - **Reviewer**) ⚠️ Review applies to [ead5f3ce](https://git.openjdk.org/skara/pull/1412/files/ead5f3ce6e5b943cb9e899f6319a0b915f8035b0)
 * [Magnus Ihse Bursie](https://openjdk.org/census#ihse) (@magicus - **Reviewer**) ⚠️ Review applies to [ead5f3ce](https://git.openjdk.org/skara/pull/1412/files/ead5f3ce6e5b943cb9e899f6319a0b915f8035b0)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1412/head:pull/1412` \
`$ git checkout pull/1412`

`$ git checkout pull/1412` \
`$ git pull https://git.openjdk.org/skara pull/1412/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1412`

`$ git pr show -t 1412`

</details>

<a href="https://git.openjdk.org/skara/pull/1412.diff">https://git.openjdk.org/skara/pull/1412.diff</a>

</details>

### DIFF
--- a/bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRIssueBot.java
+++ b/bots/csr/src/main/java/org/openjdk/skara/bots/csr/CSRIssueBot.java
@@ -22,6 +22,7 @@
  */
 package org.openjdk.skara.bots.csr;
 
+import java.time.Duration;
 import java.time.Instant;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
@@ -34,6 +35,7 @@ import org.openjdk.skara.bot.Bot;
 import org.openjdk.skara.bot.WorkItem;
 import org.openjdk.skara.forge.HostedRepository;
 import org.openjdk.skara.issuetracker.Issue;
+import org.openjdk.skara.issuetracker.IssuePoller;
 import org.openjdk.skara.issuetracker.IssueProject;
 
 /**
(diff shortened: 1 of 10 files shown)


edvbld (approved):
Looks good, the code using `PullRequestPoller` turned out much nicer 👍

I feel like I'm always comment on naming, feel free to ignore such suggestions, they are just suggestions 😃 In this case I would have renamed `PullRequestPoller.lastBatchHandled` to `PullRequestPoller.ignore(List<PullRequest> prs)` and the bots would typically call `poller.ignore(prs);`

magicus (approved):
Nice to get this in place!

erikj79:
> I feel like I'm always comment on naming, feel free to ignore such suggestions, they are just suggestions 😃 In this case I would have renamed `PullRequestPoller.lastBatchHandled` to `PullRequestPoller.ignore(List<PullRequest> prs)` and the bots would typically call `poller.ignore(prs);`

I'm not completely satisfied with the naming of the methods either, but I don't really get the "ignore" suggestion here. The idea with the method is to acknowledge to the poller that what it returned has now been accepted. In most cases, having this two step ack protocol isn't needed, but in some bots, there are further remote calls made in `getPeriodicItems` after `updatedPullRequests` is called, which could potentially fail, and if they do, I want to make sure the next call to `getPeriodicItems` will include those PRs again. I'm sure you figured this out already.

In a separate discussion, you also pointed out that the naming of the `retryPullRequest` methods was strange to you. You may have a point there too. There is also an undocumented requirement in the PullRequestPoller, that `retryPullRequest` must not be called before `lastBatchHandled`. This is how I implicitly assumed it was always going to be used, but it does need to be made explicit, at least in documentation. I will look into this.

PR #1413, opened 2022-11-02
1660: Add confname field to slack log config

erikj79:
The Slack log handler currently takes a "username" value as input which if set is sent to Slack as the user name of the message. With modern webhooks, it unfortunately seems that we have lost the feature of overriding the user name of a message. Our current configs put a descriptive string identifying the bot configuration in this field, which we can no longer see.

I want to add a specific new configuration field to the Slack log handler for the "configname". If set, this will appear first in every Slack message.

---------
### Progress

### Issue
 * [SKARA-1660](https://bugs.openjdk.org/browse/SKARA-1660): Add confname field to slack log config

### Reviewers
 * [Zhao Song](https://openjdk.org/census#zsong) (@zhaosongzs - Author) ⚠️ Review applies to [978e28aa](https://git.openjdk.org/skara/pull/1413/files/978e28aac44dcb5d7d27ac42dd6e7513e933311d)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1413/head:pull/1413` \
`$ git checkout pull/1413`

`$ git checkout pull/1413` \
`$ git pull https://git.openjdk.org/skara pull/1413/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1413`

`$ git pr show -t 1413`

</details>

<a href="https://git.openjdk.org/skara/pull/1413.diff">https://git.openjdk.org/skara/pull/1413.diff</a>

</details>

### DIFF
--- a/bots/cli/src/main/java/org/openjdk/skara/bots/cli/BotLauncher.java
+++ b/bots/cli/src/main/java/org/openjdk/skara/bots/cli/BotLauncher.java
@@ -76,7 +76,7 @@ public class BotLauncher {
             }
             var handler = new BotSlackHandler(URIBuilder.base(config.get("log").get("slack").get("webhook").asString()).build(),
                     config.get("log").get("slack").get("username").asString(),
-                    config.get("log").get("slack").get("configname").asString(),
+                    config.get("log").get("slack").get("prefix").asString(),
                     maxRate,
                     details);
             handler.setLevel(level);
--- a/bots/cli/src/main/java/org/openjdk/skara/bots/cli/BotLauncher.java
+++ b/bots/cli/src/main/java/org/openjdk/skara/bots/cli/BotLauncher.java
@@ -75,9 +75,10 @@ public class BotLauncher {
                                                           o -> o.get("link").asString()));
             }
             var handler = new BotSlackHandler(URIBuilder.base(config.get("log").get("slack").get("webhook").asString()).build(),
-                                              config.get("log").get("slack").get("username").asString(),
-                                              maxRate,
-                                              details);
+                    config.get("log").get("slack").get("username").asString(),
+                    config.get("log").get("slack").get("configname").asString(),
+                    maxRate,
+                    details);
             handler.setLevel(level);
             log.addHandler(handler);
         }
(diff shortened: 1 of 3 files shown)


zhaosongzs (approved):
LGTM

edvbld (commented):
Hey Erik, nice work! Just two questions on the patch:

- why keep the field and config value `username`? If modern Slack hooks do not allow a username to be chosen, then shouldn't we remove that property from the config and the corresponding field from `BotSlackHandler`?
- I would probably have used the name `prefix` instead of `configname` for the configuration setting, it is an arbitrary prefix that we just happen to set to the name of the configuration 😄 

Thanks,
Erik

erikj79:
> Hey Erik, nice work! Just two questions on the patch:
> 
> * why keep the field and config value `username`? If modern Slack hooks do not allow a username to be chosen, then shouldn't we remove that property from the config and the corresponding field from `BotSlackHandler`?

I'm not sure if this is a property of all slack webhooks or just the one app we are able to use at the moment. I figured we could leave the code in there for now and just not configure a username. 

> * I would probably have used the name `prefix` instead of `configname` for the configuration setting, it is an arbitrary prefix that we just happen to set to the name of the configuration 😄

That is a good idea. I will change it.

edvbld:
> > Hey Erik, nice work! Just two questions on the patch:
> > 
> > * why keep the field and config value `username`? If modern Slack hooks do not allow a username to be chosen, then shouldn't we remove that property from the config and the corresponding field from `BotSlackHandler`?
> 
> I'm not sure if this is a property of all slack webhooks or just the one app we are able to use at the moment. I figured we could leave the code in there for now and just not configure a username.

Ok, yeah, lets leave it in for a while.

edvbld (approved):

magicus (approved):
The patch looks good. but I'm thinking if you should use some delimiter for the prefix, like `[` ... `]`. Or is the plan to make formatting for the prefix to be clearly marked as a prefix in the configuration? (I'm just worried that if this is not done properly it can be hard to distinguish the prefix from the message proper, making it look odd)

erikj79:
> The patch looks good. but I'm thinking if you should use some delimiter for the prefix, like `[` ... `]`. Or is the plan to make formatting for the prefix to be clearly marked as a prefix in the configuration? (I'm just worried that if this is not done properly it can be hard to distinguish the prefix from the message proper, making it look odd)

My plan was to have the prefix config completely raw and put any formatting in the config. That makes it more flexible and easier to change as well.

PR #1416, opened 2022-11-08
SKARA-1658

zhaosongzs:
Currently, when the pr bot tries to update the PR body, users are also likely to update the PR body at the same time. Therefore, users' updates are likely to be overwritten by bot.  

This is a race between the bot and user and the race window is a little big right now.

To reduce the race window, in this patch, when PR bot is trying to update PR body, it will check whether the PR body has been modified by user. And if the PR bot has been modified, PR bot would not overwrite it.

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -738,9 +738,14 @@ class CheckRun {
         }
         var newBody = originalBody + "\n\n" + progressMarker + "\n" + message;
 
-        // TODO? Retrieve the body again here to lower the chance of concurrent updates
-        log.info("Updating PR body");
-        pr.setBody(newBody);
+        // Retrieve the body again here to lower the chance of concurrent updates
+        if (pr.body().equals(pr.latestBody())) {
+            log.info("Updating PR body");
+            pr.setBody(newBody);
+        } else {
+            log.info("PR body has been modified, won't update PR body this time");
+            return description;
+        }
         return newBody;
     }
 
--- a/forge/src/main/java/org/openjdk/skara/forge/PullRequest.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/PullRequest.java
@@ -248,4 +248,6 @@ public interface PullRequest extends Issue {
                 })
                 .toList();
     }
+
+    String latestBody();
 }
--- a/forge/src/test/java/org/openjdk/skara/forge/github/GitHubRestApiTests.java
+++ b/forge/src/test/java/org/openjdk/skara/forge/github/GitHubRestApiTests.java
@@ -148,4 +148,15 @@ public class GitHubRestApiTests {
         assertTrue(updateComment.body().contains("..."));
         assertTrue(updateComment.body().contains("2"));
     }
+
+    @Test
+    void testLatestBody(){
+        var testRepoOpt = githubHost.repository("openjdk/playground");
+        assumeTrue(testRepoOpt.isPresent());
+        var testRepo = testRepoOpt.get();
+        var testPr = testRepo.pullRequest("99");
+
+        String latestBody = testPr.latestBody();
+        assertEquals("test", latestBody);
+    }
 }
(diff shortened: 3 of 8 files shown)


forge/src/test/java/org/openjdk/skara/forge/github/GitHubRestApiTests.java
```diff
@@ -148,4 +148,15 @@ void testOversizeComment() {
...
     }
+
+    @Test
+    void testLatestBody(){
+        var testRepoOpt = githubHost.repository("openjdk/playground");
+        assumeTrue(testRepoOpt.isPresent());
+        var testRepo = testRepoOpt.get();
+        var testPr = testRepo.pullRequest("99");
+
+        String latestBody = testPr.latestBody();
```
magicus, line 159: This looks a bit fragile, but maybe that's a price we have to pay for the manual tests.

However, I think you should also add an automatic test, that verifies that an updated body makes the bot skip the update that time around, but not otherwise. I think it will be possible to define such a setup without having to resort to race-prone logic.

test/src/main/java/org/openjdk/skara/test/TestIssue.java
```diff
@@ -91,7 +91,7 @@ public void setTitle(String title) {
 
     @Override
     public String body() {
-        return store.body();
+        return body;
```
zhaosongzs, line 94: Make `TestIssue#body()` consistent with `GitHubPullRequest#body()` and `GitLabMergeRequest#body()`.

It was returning the latest body from the 'remote' instead of the cache. So it is different from `GitHubPullRequest` and `GitLabMergeRequest`

magicus: In effect, it sounds like `TestIssue` did not properly mimick the real classes. So fixing this, and fixing the tests so they pass, is probably a good idea. Maybe you should do that separately, yes, and integrate that first. Then you can easily add a proper test to this PR.

zhaosongzs: > In effect, it sounds like `TestIssue` did not properly mimick the real classes. So fixing this, and fixing the tests so they pass, is probably a good idea. Maybe you should do that separately, yes, and integrate that first. Then you can easily add a proper test to this PR.

Got it! Thx!

erikj79:
Looks like I forgot to publish the review for these comments yesterday. Sorry about that.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -738,9 +738,14 @@ private String updateStatusMessage(String message) {
...
-        // TODO? Retrieve the body again here to lower the chance of concurrent updates
-        log.info("Updating PR body");
-        pr.setBody(newBody);
+        // Retrieve the body again here to lower the chance of concurrent updates
+        if (pr.body().equals(pr.latestBody())) {
+            log.info("Updating PR body");
+            pr.setBody(newBody);
+        } else {
+            log.info("PR body has been modified, won't update PR body this time");
+            return description;
```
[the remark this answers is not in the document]
zhaosongzs: I thought the body will be updated in next round, so I didn't throw exception here.  Why is retry here better than updating body in next round?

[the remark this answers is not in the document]
erikj79: Hm, you are right, if the body has been updated since this CheckRun was triggered, we will get another one scheduled anyway. Since this isn't actually a failure condition, it's better to not throw exceptions that will annoy admins.

forge/src/main/java/org/openjdk/skara/forge/PullRequest.java
```diff
@@ -248,4 +248,6 @@ static List<Review> calculateReviewTargetRefs(List<Review> reviews, List<Referen
                 })
                 .toList();
     }
+
+    String latestBody();
```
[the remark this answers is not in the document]
zhaosongzs: Will change it.

erikj79 (approved):

PR #1422, opened 2022-11-17
1680: NotifyBot::getPeriodicItems takes too long

erikj79:
When starting up the notify bot, the first round of `getPeriodicItems` was observed to make >2500 calls to github for repository branches (taking a total of 18 minutes). This call originates from `NotifyBot::isOfInterest`, which looks for pre-integration branches matching the PR. There is one call for every PR being considered, plus pagination. The JDK repository has so many open PRs at the moment that this call needs 8 pages. ~250 PRs x 8 pages is 2000 calls just for that repo. On top of this, there is also another check in `isOfInterest` that looks for a "ready comment". This check needs to parse all comments of a PR, which there can also be quite a few of, making it deceptively expensive.

The method `Bot::getPeriodicItems` is expected to a be a quick method that gets called at a set interval to spawn new `WorkItem` instances. The `BotRunner` calls all the `getPeriodicItems` serially, so if they start taking significant time, it adds up pretty quickly. This is especially true in the very first round after a bot restart, where a lot of PRs need to be evaluated. The intent of having this "quick" discarding of PRs in `getPeriodicItems` was to avoid the overhead of spawning WorkItems for PRs that could easily be identified as not needing it, but in this case, it seems to have backfired.

To fix this, I'm moving the `isOfInterest` method away from `NotifyBot` and into `PullRequestWorkItem::run` and making it the first thing to be called. By doing this, we add the overhead of always spawning a `WorkItem`, but we move the heavy processing into the WorkItems, who execute with high concurrency, away from the serially executed `getPeriodicItem`. Looking at metrics data, it's pretty clear that the initial `getPeriodicItems` isn't able to saturate the thread pool with enough WorkItems today due to all these expensive checks, so this change alone should improve latency quite significantly for the first round.

To further reduce unnecessary work, I'm rearranging `isOfInterest` a bit, and only checking branches if the labels aren't found on the PR, and the prbranch notifier is actually configured on the repo. I'm also changing how we check for the existence of the pre integration branch. Instead of fetching all branches (which may result in 8+ pages), I'm rewriting the `HostedRepository::branchHash` method to return an `Optional<Hash>` that will be empty on 404. That makes it usable for checking if a branch exists.

---------
### Progress

### Issue
 * [SKARA-1680](https://bugs.openjdk.org/browse/SKARA-1680): NotifyBot::getPeriodicItems takes too long

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1422/head:pull/1422` \
`$ git checkout pull/1422`

`$ git checkout pull/1422` \
`$ git pull https://git.openjdk.org/skara pull/1422/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1422`

`$ git pr show -t 1422`

</details>

<a href="https://git.openjdk.org/skara/pull/1422.diff">https://git.openjdk.org/skara/pull/1422.diff</a>

</details>

### DIFF
--- a/forge/src/main/java/org/openjdk/skara/forge/github/GitHubRepository.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/github/GitHubRepository.java
@@ -293,9 +293,14 @@ public class GitHubRepository implements HostedRepository {
     }
 
     @Override
-    public Hash branchHash(String ref) {
-        var branch = request.get("branches/" + ref).execute();
-        return new Hash(branch.get("commit").get("sha").asString());
+    public Optional<Hash> branchHash(String ref) {
+        var branch = request.get("branches/" + ref)
+                .onError(r -> r.statusCode() == 404 ? Optional.of(JSON.object().put("NOT_FOUND", true)) : Optional.empty())
+                .execute();
+        if (branch.contains("NOT_FOUND")) {
+            return Optional.empty();
+        }
+        return Optional.of(new Hash(branch.get("commit").get("sha").asString()));
     }
 
     @Override
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/NotifyBot.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/NotifyBot.java
@@ -46,8 +46,6 @@ public class NotifyBot implements Bot, Emitter {
     private final String integratorId;
     private final PullRequestPoller poller;
 
-    private ZonedDateTime lastFullUpdate;
-
     NotifyBot(HostedRepository repository, Path storagePath, Pattern branches, StorageBuilder<UpdatedTag> tagStorageBuilder,
               StorageBuilder<UpdatedBranch> branchStorageBuilder, StorageBuilder<PullRequestState> prStateStorageBuilder,
               Map<String, Pattern> readyComments, String integratorId) {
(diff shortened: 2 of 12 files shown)


zhaosongzs (approved):
LGTM!

magicus:
Have you been able to verify if this does any difference in startup speed for the bot?

forge/src/main/java/org/openjdk/skara/forge/github/GitHubRepository.java
```diff
@@ -293,9 +293,14 @@ public long id() {
...
 
     @Override
-    public Hash branchHash(String ref) {
-        var branch = request.get("branches/" + ref).execute();
-        return new Hash(branch.get("commit").get("sha").asString());
+    public Optional<Hash> branchHash(String ref) {
+        var branch = request.get("branches/" + ref)
+                .onError(r -> r.statusCode() == 404 ? Optional.of(JSON.object().put("NOT_FOUND", true)) : Optional.empty())
+                .execute();
+        if (branch.contains("NOT_FOUND")) {
```
magicus, line 300: I'm trying to figure out the logic here. So for a 404, you return a special "NOT_FOUND" object, which you detect after the request is done, to return Optional.empty().

But for any other errors..? This `var` syntax might be convenient when working in an IDE, but it sure makes code reviewing difficult. I assume `branch` is a JSON object, and that the results from onError are unwrapped before returning it from the request. But how will this work with the return statement below? 

branch.get("commit").get("sha") seems like it will fail with an exception...

erikj79: This is an established pattern in Skara that I Just copied. The `onError` lambda returns an optional. If that optional is empty, then the default error behavior happens, which is to throw exception. So by only returning an object on 404, then that's the only result code we are changing the behavior on.

erikj79:
> Have you been able to verify if this does any difference in startup speed for the bot?

Yes, I did try deploying this patch live for a bit and there was significant speedup, but I don't remember exactly by how much. I believe it was at least cut in half, if not more.

forge/src/main/java/org/openjdk/skara/forge/github/GitHubRepository.java
magicus: Aha, I see. A bit odd design, but it works I guess.

magicus (approved):
LGTM

PR #1423, opened 2022-11-17
1682: PullRequestBot::getPeriodicItems takes too long

erikj79:
Similar to the NotifyBot in [SKARA-1680](https://bugs.openjdk.org/browse/SKARA-1680), the PullRequestBot is also suffering from spending a very long time in the very first round of getPeriodicItems after a bot restart. The last time we redeployed this bot, it took ~40 minutes. This was long enough to trigger several users to ask if Skara was having issues.

The cause is similar. There is an `isReady` call for each PR, which checks for comments. The comment checking here is even worse than for the NotifyBot because every PR is checked for comments, not just those that match the "rfr" or "integrated" labels.

My proposed solution is similar. We need to move this expensive evaluation to the relevant WorkItems where they can run concurrently instead of serially in `getPeriodicItems`. I chose to make the call from the abstract superclass `PullRequestWorkItem`. This does add the check unnecessarily to the `LabelerWorkItem`, but that one is running so rarely that I don't think it matters performance wise. For correctness, I don't think it hurts checking an extra time.

---------
### Progress

### Issue
 * [SKARA-1682](https://bugs.openjdk.org/browse/SKARA-1682): PullRequestBot::getPeriodicItems takes too long

### Reviewers
 * [Zhao Song](https://openjdk.org/census#zsong) (@zhaosongzs - Committer) ⚠️ Review applies to [1ead6076](https://git.openjdk.org/skara/pull/1423/files/1ead6076eb1f747904b759cecc9b28ec813432ec)
 * [Magnus Ihse Bursie](https://openjdk.org/census#ihse) (@magicus - **Reviewer**) ⚠️ Review applies to [43272b56](https://git.openjdk.org/skara/pull/1423/files/43272b5692f728a9b59a5089be79c799e684c661)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1423/head:pull/1423` \
`$ git checkout pull/1423`

`$ git checkout pull/1423` \
`$ git pull https://git.openjdk.org/skara pull/1423/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1423`

`$ git pr show -t 1423`

</details>

<a href="https://git.openjdk.org/skara/pull/1423.diff">https://git.openjdk.org/skara/pull/1423.diff</a>

</details>

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
@@ -51,8 +51,9 @@ class CheckWorkItem extends PullRequestWorkItem {
     private static final Pattern BACKPORT_ISSUE_TITLE_PATTERN = Pattern.compile("^Backport\\s*(?:(?<prefix>[A-Za-z][A-Za-z0-9]+)-)?(?<id>[0-9]+)\\s*$");
     private static final String ELLIPSIS = "…";
 
-    CheckWorkItem(PullRequestBot bot, String prId, Consumer<RuntimeException> errorHandler, ZonedDateTime prUpdatedAt) {
-        super(bot, prId, errorHandler, prUpdatedAt);
+    CheckWorkItem(PullRequestBot bot, String prId, Consumer<RuntimeException> errorHandler, ZonedDateTime prUpdatedAt,
+            boolean needsReadyCheck) {
+        super(bot, prId, errorHandler, prUpdatedAt, needsReadyCheck);
     }
 
     private String encodeReviewer(HostUser reviewer, CensusInstance censusInstance) {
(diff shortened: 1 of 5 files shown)


zhaosongzs (approved):
LGTM, just a small question, what don't we move `isReady()` to `PullRequestWorkItem`?

erikj79:
> LGTM, just a small question, why don't we move `isReady()` to `PullRequestWorkItem`?

In this patch, it was laziness mostly. Moving it would require providing the underlying data to the WorkItem as well, either by calling the bot for it or adding as additional fields in the work item.

I did think some more on this patch since yesterday, and realized that we will be doing the isReady check quite a lot more with this change. This is because CheckWorkItem and PullRequestCommandWorkItem often spawn each other in sequence, and now each of those instances would repeat the check unnecessarily (or at least this wasn't considered needed before). I think I will change this by adding a new boolean field `readyCheckNeeded` to `PullRequestWorkItem` so we can avoid doing it when not needed, and while at it, I will move `isReady` as well.

magicus (approved):
LGTM

PR #1426, opened 2022-11-22
1683: The force push notifier needs an overhaul

zhaosongzs:
A user reported a bug related with the force-push notification. Erik investigated it and found the implementation of this notification should not be in the `PullRequestBranchNotifier`. Also, users preferred to not get force-push notification if the force-push happens when the pr is in draft state.

In this patch, the implementation of checking force-push and notifying user are moved to `CheckWorkItem`. Also, force-pushing or rebasing a draft pr will not warn user any more.

---------
### Progress

### Issue
 * [SKARA-1683](https://bugs.openjdk.org/browse/SKARA-1683): The force push notifier needs an overhaul

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1426/head:pull/1426` \
`$ git checkout pull/1426`

`$ git checkout pull/1426` \
`$ git pull https://git.openjdk.org/skara pull/1426/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1426`

`$ git pr show -t 1426`

</details>

<a href="https://git.openjdk.org/skara/pull/1426.diff">https://git.openjdk.org/skara/pull/1426.diff</a>

</details>

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
@@ -389,11 +389,11 @@ class CheckWorkItem extends PullRequestWorkItem {
                 return List.of(new CheckWorkItem(bot, prId, errorHandler, prUpdatedAt));
             }
 
-            // Check Force push
+            // Check force push
             if (!pr.isDraft()) {
                 var lastForcePushTime = pr.lastForcePushTime();
                 if (lastForcePushTime.isPresent()) {
-                    var lastForcePushSuggestion = pr.comments().stream()
+                    var lastForcePushSuggestion = comments.stream()
                             .filter(comment -> comment.body().contains(FORCE_PUSH_MARKER))
                             .reduce((a, b) -> b);
                     if (lastForcePushSuggestion.isEmpty() || lastForcePushSuggestion.get().createdAt().isBefore(lastForcePushTime.get())) {
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
@@ -51,6 +51,13 @@ class CheckWorkItem extends PullRequestWorkItem {
     private static final Pattern BACKPORT_ISSUE_TITLE_PATTERN = Pattern.compile("^Backport\\s*(?:(?<prefix>[A-Za-z][A-Za-z0-9]+)-)?(?<id>[0-9]+)\\s*$");
     private static final String ELLIPSIS = "…";
 
+    protected static final String FORCE_PUSH_MARKER = "<!-- force-push suggestion -->";
+    protected static final String FORCE_PUSH_SUGGESTION= """
+            Please do not rebase or force-push to an active PR as it invalidates existing review comments. \
+            All changes will be squashed into a single commit automatically when integrating. \
+            See [OpenJDK Developers’ Guide](https://openjdk.org/guide/#working-with-pull-requests) for more information.
+            """;
+
     CheckWorkItem(PullRequestBot bot, String prId, Consumer<RuntimeException> errorHandler, ZonedDateTime prUpdatedAt) {
         super(bot, prId, errorHandler, prUpdatedAt);
     }
@@ -380,6 +387,20 @@ class CheckWorkItem extends PullRequestWorkItem {
                 var updatedPr = bot.repo().pullRequest(prId);
                 logLatency("Time from PR updated to title corrected ", updatedPr.updatedAt(), log);
                 return List.of(new CheckWorkItem(bot, prId, errorHandler, prUpdatedAt));
+            }
+
+            // Check Force push
+            if (!pr.isDraft()) {
+                var lastForcePushTime = pr.lastForcePushTime();
+                if (lastForcePushTime.isPresent()) {
+                    var lastForcePushSuggestion = pr.comments().stream()
+                            .filter(comment -> comment.body().contains(FORCE_PUSH_MARKER))
+                            .reduce((a, b) -> b);
+                    if (lastForcePushSuggestion.isEmpty() || lastForcePushSuggestion.get().createdAt().isBefore(lastForcePushTime.get())) {
+                        log.info("Found force-push for " + pr.repository().name() + "#" + pr.id() + ", adding force-push suggestion");
+                        pr.addComment("@" + pr.author().username() + " " + FORCE_PUSH_SUGGESTION + FORCE_PUSH_MARKER);
+                    }
+                }
             }
 
             try {
--- a/forge/src/main/java/org/openjdk/skara/forge/PullRequest.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/PullRequest.java
@@ -182,8 +182,8 @@ public interface PullRequest extends Issue {
 
     /**
      * Return the last time something was force pushed while not in draft state.
-     * If there is no force-push in pull request
-     * or the restful api doesn't support force-push, return empty.
+     * If there is no force-push in pull request or the restful api doesn't
+     * support force-push, return empty.
      */
     Optional<ZonedDateTime> lastForcePushTime();
 
--- a/forge/src/main/java/org/openjdk/skara/forge/PullRequest.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/PullRequest.java
@@ -182,7 +182,8 @@ public interface PullRequest extends Issue {
 
     /**
      * Return the last time something was force pushed while not in draft state.
-     * If there is no force-push in pull request or the restful api doesn't support force-push, return empty.
+     * If there is no force-push in pull request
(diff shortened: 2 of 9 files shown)


zhaosongzs:
My current implementation has a bug. If a user force-pushes or rebases when the pr is in ready state, and then he converts the pr to draft state very quickly, in such a case,  our pullRequestBot won't have enough time to send warning, and then user will never get the notification of this force-push or rebase.

forge/src/main/java/org/openjdk/skara/forge/github/GitHubPullRequest.java
```diff
@@ -795,4 +796,20 @@ public String limitBodySize(String body) {
         }
         return body;
     }
+
+    private ZonedDateTime lastMarkedAsReadyTime() {
+        ZonedDateTime readyTime = createdAt();
+        Optional<ZonedDateTime> latestTime = request.get("issues/" + json.get("number").toString() + "/timeline")
```
erikj79, line 802: This fetches the timeline a second time. Can you try to reuse the timeline result from lastForcePushTime()?

Could you also update the javadoc on `lastForcePushTime` that it's returning the last time something was force pushed while not in draft state.

test/src/main/java/org/openjdk/skara/test/TestPullRequest.java
```diff
@@ -272,7 +272,10 @@ public URI filesUrl(Hash hash) {
 
     @Override
     public Optional<ZonedDateTime> lastForcePushTime() {
-        return Optional.ofNullable(store().lastForcePushTime());
+        if(store().lastForcePushTime()!= null && store().lastForcePushTime().isAfter(store().lastMarkedAsReadyTime())) {
```
erikj79, line 275:
```suggestion
        if (store().lastForcePushTime()!= null && store().lastForcePushTime().isAfter(store().lastMarkedAsReadyTime())) {
```

test/src/main/java/org/openjdk/skara/test/TestPullRequestStore.java
```diff
@@ -47,13 +47,18 @@ public class TestPullRequestStore extends TestIssueStore {
...
+    private ZonedDateTime lastMarkedAsReadyTime;
+
     public TestPullRequestStore(String id, HostUser author, String title, List<String> body,
             TestHostedRepository sourceRepository, String targetRef, String sourceRef, boolean draft) {
         super(id, null, author, title, body, null);
         this.sourceRepository = sourceRepository;
         this.targetRef = targetRef;
         this.sourceRef = sourceRef;
         this.draft = draft;
+        if(!draft){
```
erikj79, line 59:
```suggestion
        if (!draft) {
```

test/src/main/java/org/openjdk/skara/test/TestPullRequestStore.java
```diff
@@ -124,9 +129,16 @@ public void setSourceRef(String sourceRef) {
 
     public void setDraft(boolean draft) {
         this.draft = draft;
+        if(!draft){
```
erikj79, line 132:
```suggestion
        if (!draft) {
```

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -382,6 +389,20 @@ public Collection<WorkItem> prRun(Path scratchPath) {
                 return List.of(new CheckWorkItem(bot, prId, errorHandler, prUpdatedAt));
             }
 
+            // Check Force push
+            if (!pr.isDraft()) {
+                var lastForcePushTime = pr.lastForcePushTime();
+                if (lastForcePushTime.isPresent()) {
+                    var lastForcePushSuggestion = pr.comments().stream()
```
erikj79, line 396: `pr.comments()` have already been fetched at the top, so no need to fetch them again.
```suggestion
                    var lastForcePushSuggestion = comments.stream()
```

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -382,6 +389,20 @@ public Collection<WorkItem> prRun(Path scratchPath) {
                 return List.of(new CheckWorkItem(bot, prId, errorHandler, prUpdatedAt));
             }
 
+            // Check Force push
```
erikj79, line 392: I'm not sure if we should put this check here in `CheckWorkItem`. This will cause it to be run quite often. Maybe it would be better to put it in `CheckRun` where it would be guarded by the check metadata condition, so it would only need to be run if something significant has changed since the last check.

erikj79 (commented):
Nice work!
I have some comments, but nothing really major. 

I put in some style corrections in some places. Please try to match the existing style with whitespace between keyword and opening paren as well as closing paren and opening brace.

test/src/main/java/org/openjdk/skara/test/TestPullRequestStore.java
zhaosongzs: > Nice work! I have some comments, but nothing really major.
> 
> I put in some style corrections in some places. Please try to match the existing style with whitespace between keyword and opening paren as well as closing paren and opening brace.

My bad, I have the habit of using intellij to automatically adjust the format, but this time I forgot at some places, I will fix it.

forge/src/main/java/org/openjdk/skara/forge/github/GitHubPullRequest.java
zhaosongzs: Good idea! I will try to reuse it. I will update the java doc too.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
zhaosongzs: > I'm not sure if we should put this check here in `CheckWorkItem`. This will cause it to be run quite often. Maybe it would be better to put it in `CheckRun` where it would be guarded by the check metadata condition, so it would only need to be run if something significant has changed since the last check.

I think this is already guarded by the check metadata condition. It will only be executed when `currentCheckValid(census, comments, activeReviews, labels)` is false. The if clause is very long(line 273-425).

erikj79: You are right, by bad. I misremembered the location of currentCheckValid. This looks like the right place to do this then.

forge/src/main/java/org/openjdk/skara/forge/PullRequest.java
```diff
@@ -181,8 +181,8 @@ default boolean isSame(PullRequest other) {
     }
 
     /**
-     * Get the last force-push time. If there is no force-push in pull request
-     * or the restful api doesn't support force-push, return empty.
+     * Return the last time something was force pushed while not in draft state.
+     * If there is no force-push in pull request or the restful api doesn't support force-push, return empty.
```
erikj79, line 185: Try to keep javadoc comments to 80 char line width.

forge/src/main/java/org/openjdk/skara/forge/github/GitHubPullRequest.java
```diff
@@ -795,4 +797,19 @@ public String limitBodySize(String body) {
...
+        ZonedDateTime readyTime = createdAt();
+        Optional<ZonedDateTime> latestTime = timelineJSON
+                .stream()
+                .map(JSONValue::asObject)
+                .filter(obj -> obj.contains("event"))
+                .filter(obj -> obj.get("event").asString().equals("ready_for_review"))
+                .reduce((a, b) -> b)
+                .map(obj -> ZonedDateTime.parse(obj.get("created_at").asString()));
+        if (latestTime.isPresent()) {
+            readyTime = latestTime.get();
```
erikj79, line 811: You could express this as
```
latestTime.orElseGet(() -> createdAt())
```
and just put it last in the stream call chain.

forge/src/main/java/org/openjdk/skara/forge/PullRequest.java
zhaosongzs: OK, will fix it!

forge/src/main/java/org/openjdk/skara/forge/PullRequest.java
```diff
@@ -181,7 +181,8 @@ default boolean isSame(PullRequest other) {
     }
 
     /**
-     * Get the last force-push time. If there is no force-push in pull request
+     * Return the last time something was force pushed while not in draft state.
+     * If there is no force-push in pull request
      * or the restful api doesn't support force-push, return empty.
```
erikj79, line 186: While this is strictly less than 80, we also like to keep the lines even. I'm not sure if Intellij has any automation for this. I've just capitulated to doing it manually.

zhaosongzs: I also don't know whether IntelliJ can do this. I will do it manually.

erikj79 (approved):

PR #1428, opened 2022-11-29
SKARA-1690

zhaosongzs:
Currently, the command `git skara jcheck` can only use the `.jcheck/conf` configuration file in the current checking commit.
However, sometimes users want to run jcheck with different configurations to validate their commits. Therefore, we need to upgrade `git skara jcheck`.

In this patch, `git skara jcheck` would be able to support following usecases.

1. Run jcheck on a commit or a series of commits using the .jcheck/conf in the same commit. (what we do by default today)
`git skara jcheck`

2. Run jcheck on a commit or a series of commits using the .jcheck/conf in a different specified commit. 
`git skara jcheck --specified-conf-commit <COMMIT HASH>`

3. Run jcheck on a commit or a series of commits using the .jcheck/conf in my workspace. 
`git skara jcheck --workspace-conf`

4. Run jcheck on a commit or a series of commits using a config file that I point to directly, that may have any name. 
`git skara jcheck --workspace-conf --conf-file <FILENAME>`

5. Run jcheck on the diff in my current workspace, either --staged or not using the .jcheck/conf in my workspace.
`git skara jcheck --workspace-diff`

### DIFF
--- a/cli/src/main/java/org/openjdk/skara/cli/GitJCheck.java
+++ b/cli/src/main/java/org/openjdk/skara/cli/GitJCheck.java
@@ -98,6 +98,17 @@ public class GitJCheck {
                   .describe("CHECKS")
                   .helptext("Ignore errors from checks with the given name")
                   .optional(),
+            Option.shortcut("")
+                  .fullname("specified-conf-commit")
+                  .describe("COMMIT HASH")
+                  .helptext("Use jcheck configuration in specified commit")
+                  .optional(),
+            Option.shortcut("")
+                  .fullname("conf-file")
+                  .describe("FILE")
+                  .helptext("Use this file as jcheck configuration instead of .jcheck/conf, " +
+                          "this flag can only be used when 'workspace-conf' is enabled")
+                  .optional(),
             Switch.shortcut("")
                   .fullname("setup-pre-push-hook")
                   .helptext("Set up a pre-push hook that runs jcheck on commits to be pushed")
@@ -125,6 +136,15 @@ public class GitJCheck {
             Switch.shortcut("v")
                   .fullname("version")
                   .helptext("Print the version of this tool")
+                  .optional(),
+            Switch.shortcut("")
+                  .fullname("workspace-conf")
+                  .helptext("Use jcheck configuration in current workspace")
+                  .optional(),
+            Switch.shortcut("")
+                  .fullname("workspace-diff")
+                  .helptext("Run jcheck includes diff in workspace(either staged or not), " +
+                          "Please note that after run jcheck with this flag, all your staged diff will be unstaged")
                   .optional()
         );
 
@@ -213,15 +233,73 @@ public class GitJCheck {
             }
         }
 
+        JCheckConfiguration overridingConfig = null;
+        // This two flags are mutually exclusive
+        if (arguments.contains("workspace-conf") && arguments.contains("specified-conf-commit")) {
+            System.err.println(String.format("error: you can only choose one from using jcheck " +
+                    "configuration in work space or in a specified commit"));
+            return 1;
+        }
+        // Using jcheck configuration in workspace
+        if (arguments.contains("workspace-conf")) {
+            String confFileName = ".jcheck/conf";
+            if (arguments.contains("conf-file")) {
+                confFileName = arguments.get("conf-file").asString();
+            }
+            try {
+                var content = Files.readAllBytes(Path.of(confFileName));
+                var lines = new String(content, StandardCharsets.UTF_8).lines().toList();
+                overridingConfig = JCheck.parseConfiguration(lines, List.of()).get();
+            } catch (NoSuchFileException e) {
+                System.err.println(String.format("error: File %s doesn't exist!", confFileName));
+                return 1;
+            } catch (IllegalArgumentException e) {
+                System.err.println(String.format("error: Invalid jcheck configuration: %s,", e.getMessage()));
+                return 1;
+            }
+        }
+        // Using jcheck configuration in a specified commit
+        if (arguments.contains("specified-conf-commit")) {
+            var shortHash = arguments.get("specified-conf-commit").asString();
+            var confCommitHash = repo.wholeHash(shortHash);
+            if (confCommitHash.isEmpty()) {
+                System.err.println(String.format("error: commit %s is invalid!", shortHash));
+                return 1;
+            }
+            try {
+                overridingConfig = JCheck.parseConfiguration(repo, confCommitHash.get(), List.of()).get();
+            } catch (IllegalArgumentException e) {
+                System.err.println(String.format("error: Invalid jcheck configuration: %s", e.getMessage()));
+                return 1;
+            }
+        }
+        // Commit local changes temporarily
+        Boolean checkWorkspaceDiff = arguments.contains("workspace-diff");
+        Hash originHash = null;
+        if (checkWorkspaceDiff) {
+            originHash = repo.head();
+            repo.add(Paths.get("."));
+            repo.commit("temp commit", "jcheck", "jcheck@openjdk.com");
+        }
         var isLax = getSwitch("lax", arguments);
         var visitor = new JCheckCLIVisitor(ignore, isMercurial, isLax);
         var commitMessageParser = isMercurial ? CommitMessageParsers.v0 : CommitMessageParsers.v1;
         for (var range : ranges) {
-            try (var errors = JCheck.check(repo, census, commitMessageParser, range)) {
+            try (var errors = JCheck.check(repo, census, commitMessageParser, range, overridingConfig)) {
                 for (var error : errors) {
                     error.accept(visitor);
                 }
+            } catch (Exception e) {
+                System.err.println(String.format("error: JCheck failed due to exception: %s", e.getMessage()));
+                if (checkWorkspaceDiff) {
+                    repo.reset(originHash, false);
+                }
+                return 1;
             }
+        }
+        // Reset the changes
+        if (checkWorkspaceDiff) {
+            repo.reset(originHash, false);
         }
         return visitor.hasDisplayedErrors() ? 1 : 0;
     }
--- a/jcheck/src/main/java/org/openjdk/skara/jcheck/JCheck.java
+++ b/jcheck/src/main/java/org/openjdk/skara/jcheck/JCheck.java
@@ -256,7 +256,8 @@ public class JCheck {
     public static Issues check(ReadOnlyRepository repository,
                                Census census,
                                CommitMessageParser parser,
-                               String revisionRange) throws IOException {
+                               String revisionRange,
+                               JCheckConfiguration overridingConfig) throws IOException {
         if (repository.isEmpty()) {
             return new Issues(new ArrayList<Issue>().iterator(), null);
         }
@@ -270,7 +271,7 @@ public class JCheck {
         var branchRegex = conf.isPresent() ? conf.get().repository().branches() : ".*";
         var tagRegex = conf.isPresent() ? conf.get().repository().tags() : ".*";
 
-        return check(repository, parser, branchRegex, tagRegex, revisionRange, List.of(), null, census);
+        return check(repository, parser, branchRegex, tagRegex, revisionRange, List.of(), overridingConfig, census);
     }
 
     public static Set<Check> checksFor(ReadOnlyRepository repository, Hash hash) throws IOException {
--- a/vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
+++ b/vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
@@ -1659,4 +1659,15 @@ public class GitRepository implements Repository {
     public Hash initialHash() {
         return EMPTY_TREE;
     }
+
+    @Override
+    public Optional<Hash> wholeHash(String hash) {
+        try (var p = capture("git", "rev-parse", hash)) {
+            var res = p.await();
+            if (res.status() == 0 && res.stdout().size() == 1) {
+                return Optional.of(new Hash(res.stdout().get(0)));
+            }
+            return Optional.empty();
+        }
+    }
 }
--- a/jcheck/src/test/java/org/openjdk/skara/jcheck/JCheckTests.java
+++ b/jcheck/src/test/java/org/openjdk/skara/jcheck/JCheckTests.java
@@ -296,7 +296,7 @@ class JCheckTests {
             var census = Census.parse(censusPath);
 
             var visitor = new TestVisitor();
-            try (var issues = JCheck.check(repo, census, CommitMessageParsers.v1, first.hex() + ".." + second.hex())) {
+            try (var issues = JCheck.check(repo, census, CommitMessageParsers.v1, first.hex() + ".." + second.hex(), null)) {
                 for (var issue : issues) {
                     issue.accept(visitor);
                 }
@@ -329,7 +329,7 @@ class JCheckTests {
             // Check the last commit without reviewers, should pass since .jcheck/conf was updated
             var range = initialCommit.hash().hex() + ".." + secondCommit.hex();
             var visitor = new TestVisitor();
-            try (var issues = JCheck.check(repo, census, CommitMessageParsers.v1, range)) {
+            try (var issues = JCheck.check(repo, census, CommitMessageParsers.v1, range, null)) {
                 for (var issue : issues) {
                     issue.accept(visitor);
                 }
--- a/jcheck/src/test/java/org/openjdk/skara/jcheck/TestRepository.java
+++ b/jcheck/src/test/java/org/openjdk/skara/jcheck/TestRepository.java
@@ -347,4 +347,9 @@ class TestRepository implements ReadOnlyRepository {
     public Hash initialHash() {
         return null;
     }
+
+    @Override
+    public Optional<Hash> wholeHash(String hash) {
+        return Optional.empty();
+    }
 }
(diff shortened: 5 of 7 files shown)


zhaosongzs:
I still have two questions.

1. For point 4, "Run jcheck on a commit or a series of commits using a config file that I point to directly, that may have any name.", currently, I think in most cases it will be used with point 3. So in my implementation, user could only point a config file with any name in workspace.

If needed, I can make user be able to point any file in any commit as the configuration file.

2. For point 5, to run jcheck on the diff in workspace, I don't have some good ideas. So I just commit the changes temporarily and then reset the head after running jcheck. But it will make user's staged changes to be unstaged.

jcheck/src/main/java/org/openjdk/skara/jcheck/JCheck.java
```diff
@@ -270,7 +271,7 @@ public static Issues check(ReadOnlyRepository repository,
         var branchRegex = conf.isPresent() ? conf.get().repository().branches() : ".*";
         var tagRegex = conf.isPresent() ? conf.get().repository().tags() : ".*";
```
zhaosongzs, line 272: I also wonder the logic of choosing jcheck configuration for repository check.

erikj79:
We need to work on the parameter names.

> 2. Run jcheck on a commit or a series of commits using the .jcheck/conf in a different specified commit.
>    `git skara jcheck --specified-conf-commit <COMMIT HASH>`

I would suggest `--conf-rev REV`. In Git we want to operate on revisions rather than just hashes. A REV can be at least either a hash, branch or tag. 

> 3. Run jcheck on a commit or a series of commits using the .jcheck/conf in my workspace.
>    `git skara jcheck --workspace-conf`

Thinking on this more, we have two different possibilities, the staged file or the one in the working tree. I would suggest: `--conf-staged` and `--conf-working-tree` for those two cases.

> 4. Run jcheck on a commit or a series of commits using a config file that I point to directly, that may have any name.
>    `git skara jcheck --workspace-conf --conf-file <FILENAME>`

I think `--conf-file` is good, but I hope you aren't requiring `--workspace-conf` to go along with it?

> 5. Run jcheck on the diff in my current workspace, either --staged or not using the .jcheck/conf in my workspace.
>    `git skara jcheck --workspace-diff`

I should have been clearer in that I wanted to cover two different cases here, either working tree, or staged. I would suggest the following to match with the conf arguments: `--staged` and `--working-tree`.

> 1. For point 4, "Run jcheck on a commit or a series of commits using a config file that I point to directly, that may have any name.", currently, I think in most cases it will be used with point 3. So in my implementation, user could only point a config file with any name in workspace.

My intention here was that the file would be anywhere on the local filesystem. That could be in a different repository or a directory outside of the repository.

> If needed, I can make user be able to point any file in any commit as the configuration file.

No, I don't think we need to be able to treat any historic file as the `.jcheck/conf` file. I don't see a usecase for that.

> 2. For point 5, to run jcheck on the diff in workspace, I don't have some good ideas. So I just commit the changes temporarily and then reset the head after running jcheck. But it will make user's staged changes to be unstaged.

That's not an acceptable solution. I realize that this will likely require some more involved changes to Jcheck itself. If you can't figure it out, better leave this feature out. I will go take a look and see if I can figure something out.

zhaosongzs:
> We need to work on the parameter names.
> 
> > 2. Run jcheck on a commit or a series of commits using the .jcheck/conf in a different specified commit.
> >    `git skara jcheck --specified-conf-commit <COMMIT HASH>`
> 
> I would suggest `--conf-rev REV`. In Git we want to operate on revisions rather than just hashes. A REV can be at least either a hash, branch or tag.
> 
> > 3. Run jcheck on a commit or a series of commits using the .jcheck/conf in my workspace.
> >    `git skara jcheck --workspace-conf`
> 
> Thinking on this more, we have two different possibilities, the staged file or the one in the working tree. I would suggest: `--conf-staged` and `--conf-working-tree` for those two cases.
> 
> > 4. Run jcheck on a commit or a series of commits using a config file that I point to directly, that may have any name.
> >    `git skara jcheck --workspace-conf --conf-file <FILENAME>`
> 
> I think `--conf-file` is good, but I hope you aren't requiring `--workspace-conf` to go along with it?
> 
> > 5. Run jcheck on the diff in my current workspace, either --staged or not using the .jcheck/conf in my workspace.
> >    `git skara jcheck --workspace-diff`
> 
> I should have been clearer in that I wanted to cover two different cases here, either working tree, or staged. I would suggest the following to match with the conf arguments: `--staged` and `--working-tree`.
> 
> > 1. For point 4, "Run jcheck on a commit or a series of commits using a config file that I point to directly, that may have any name.", currently, I think in most cases it will be used with point 3. So in my implementation, user could only point a config file with any name in workspace.
> 
> My intention here was that the file would be anywhere on the local filesystem. That could be in a different repository or a directory outside of the repository.
> 
> > If needed, I can make user be able to point any file in any commit as the configuration file.
> 
> No, I don't think we need to be able to treat any historic file as the `.jcheck/conf` file. I don't see a usecase for that.
> 
> > 2. For point 5, to run jcheck on the diff in workspace, I don't have some good ideas. So I just commit the changes temporarily and then reset the head after running jcheck. But it will make user's staged changes to be unstaged.
> 
> That's not an acceptable solution. I realize that this will likely require some more involved changes to Jcheck itself. If you can't figure it out, better leave this feature out. I will go take a look and see if I can figure something out.

Thanks for making it clear! I will try my best to fix the problems.

erikj79:
For 5, I think what we need is to create a "fake" Commit object based on the working-tree or staged content of a local repository. We could add methods like `ReadOnlyRepository::workingTree` and `::staged` that return `Commit` instances with the respective diffs, the current user as author and the current HEAD as parent. In JCheck, we would need to define some special string constants that symbolize these special "commits". So if revisionRange has one of those values, JCheck would call one of these special methods instead of `ReadOnlyRepository::commits`.

I would suggest values like `SKARA_GIT_WORKING_TREE_AS_REV` and `SKARA_GIT_STAGED_AS_REV`, something really long and unlikely to actually be a valid REV in any repository. I think these constants should stay in the jcheck module and not taint `GitRepository`, because there doesn't seem to be a standardized way of referring to the working tree as a "commit" in Git.

For Mercurial, I would just throw `UnsupportedOperationException`. It's not worth the time to implement this for Mercurial at this point.

kevinrushforth:
It case 5 really worth the complexity? The `--working-tree` option is equivalent to `--conf-file .jcheck/conf`. As for `--staged`, I can imagine that it might be an occasionally useful "nice to have", but how often will developers really want to run `jcheck` against the version of `.jcheck/conf` in the index?

erikj79:
> It case 5 really worth the complexity? The `--working-tree` option is equivalent to `--conf-file .jcheck/conf`. As for `--staged`, I can imagine that it might be an occasionally useful "nice to have", but how often will developers really want to run `jcheck` against the version of `.jcheck/conf` in the index?

The idea with 5 is that the user won't have to commit before performing a whitespace check. I find it rather cumbersome to have to commit before I can verify jcheck.

I agree that `--conf-working-tree` is equivalent to `--conf-file .jcheck/conf` so we don't really need it. OTOH, I figured it would be pretty simple to implement.

vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
```diff
@@ -1659,4 +1659,65 @@ public int commitCount() throws IOException {
...
+        return new Commit(commitMetaData, List.of(diffStaged()));
+    }
+
+    @Override
+    public Commit workingTree() throws IOException {
+        var author = new Author("jcheck", "jcheck@none.none");
+        var commitMetaData = new CommitMetadata(new Hash("working-tree"), List.of(head()), author, ZonedDateTime.now(),
+                author, ZonedDateTime.now(), List.of(""));
+        return new Commit(commitMetaData, List.of(diff(head())));
+    }
```
zhaosongzs, line 1699: I tried to implement case 5 by Erik's idea. It seems work fine. But I don't know whether my implementation is acceptable, like the author and hash in the fake commit.

kevinrushforth: I'd be more worried about what it might do to my current index or working tree. Is there any chance things could go badly and mess that up?

zhaosongzs: > I'd be more worried about what it might do to my current index or working tree. Is there any chance things could go badly and mess that up?

No, I believe under any condition it won't do anything to your index or working tree.

zhaosongzs: > I'd be more worried about what it might do to my current index or working tree. Is there any chance things could go badly and mess that up?

It will only create a fake `Commit` object but never use git to commit

cli/src/main/java/org/openjdk/skara/cli/GitJCheck.java
```diff
@@ -125,6 +134,23 @@ public static int run(Repository repo, String[] args) throws IOException {
...
+                  .fullname("conf-staged")
+                  .helptext("Use staged .jcheck/conf")
+                  .optional(),
+            Switch.shortcut("")
+                  .fullname("conf-working-tree")
+                  .helptext("Use .jcheck/conf in current working tree")
+                  .optional(),
+            Switch.shortcut("")
+                  .fullname("staged")
+                  .helptext("Run jcheck includes staged changes and by default jcheck will use staged .jcheck/conf")
```
erikj79, line 148:
```suggestion
                  .helptext("Check staged changes as if they were committed")
```

cli/src/main/java/org/openjdk/skara/cli/GitJCheck.java
```diff
@@ -125,6 +134,23 @@ public static int run(Repository repo, String[] args) throws IOException {
...
+                  .optional(),
+            Switch.shortcut("")
+                  .fullname("working-tree")
+                  .helptext("Run jcheck includes changes in working tree and by default " +
+                          "jcheck will use .jcheck/conf in working tree")
```
erikj79, line 153:
```suggestion
                  .helptext("Check changes in working tree as if they were committed")
```

cli/src/main/java/org/openjdk/skara/cli/GitJCheck.java
```diff
@@ -213,11 +239,98 @@ public static int run(Repository repo, String[] args) throws IOException {
             }
         }
 
+        var staged = arguments.contains("staged");
+        var working_tree = arguments.contains("working-tree");
```
erikj79, line 243: Please use camelCase for local variables.
```suggestion
        var workingTree = arguments.contains("working-tree");
```

cli/src/main/java/org/openjdk/skara/cli/GitJCheck.java
```diff
@@ -213,11 +239,98 @@ public static int run(Repository repo, String[] args) throws IOException {
...
+        // This two flags are mutually exclusive
+        if (staged && working_tree) {
+            System.err.println(String.format("error: you can only choose one from staged or working-tree"));
```
erikj79, line 246:
```suggestion
        // These two flags are mutually exclusive
        if (staged && working_tree) {
            System.err.println(String.format("error: can only use one of --staged or --working-tree"));
```

cli/src/main/java/org/openjdk/skara/cli/GitJCheck.java
```diff
@@ -98,6 +97,16 @@ public static int run(Repository repo, String[] args) throws IOException {
                   .describe("CHECKS")
                   .helptext("Ignore errors from checks with the given name")
                   .optional(),
+            Option.shortcut("")
+                  .fullname("conf-rev")
+                  .describe("REV")
+                  .helptext("Use .jcheck/conf in specified rev")
```
erikj79, line 103:
```suggestion
                  .helptext("Use .jcheck/conf in the specified revision")
```

cli/src/main/java/org/openjdk/skara/cli/GitJCheck.java
```diff
@@ -213,11 +239,98 @@ public static int run(Repository repo, String[] args) throws IOException {
...
+        if (confWorkingTree) {
+            confFlagCount++;
+        }
+        if (confFile) {
+            confFlagCount++;
+        }
+        // This four flags are mutually exclusive
+        if (confFlagCount > 1) {
+            System.err.println(String.format("error: you can only choose one from using jcheck " +
+                    "configuration in work space or in a specified commit"));
```
erikj79, line 271:
```suggestion
            System.err.println(String.format("error: can only use one source for jcheck configuration"));
```

cli/src/main/java/org/openjdk/skara/cli/GitJCheck.java
```diff
@@ -213,11 +239,98 @@ public static int run(Repository repo, String[] args) throws IOException {
...
+        if (confStaged) {
+            confFlagCount++;
+        }
+        if (confWorkingTree) {
+            confFlagCount++;
+        }
+        if (confFile) {
+            confFlagCount++;
+        }
+        // This four flags are mutually exclusive
```
erikj79, line 268:
```suggestion
        // These four flags are mutually exclusive
```

vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
```diff
@@ -1659,4 +1659,65 @@ public int commitCount() throws IOException {
     public Hash initialHash() {
         return EMPTY_TREE;
     }
+
+    @Override
+    public Optional<Hash> wholeHash(String rev) {
```
erikj79, line 1664: Isn't this just `::resolve(String)`?

vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
```diff
@@ -1659,4 +1659,65 @@ public int commitCount() throws IOException {
...
+            if (res.status() == 0) {
+                return Optional.of(res.stdout());
+            }
+        }
+        return Optional.empty();
+    }
+
+    @Override
+    public Commit staged() throws IOException {
+        var author = new Author("jcheck", "jcheck@none.none");
```
erikj79, line 1687: For the default user and email, there is already a `::userName` method. You can create one for email as well, and then use these values as fallbacks (`Optional.orElse`).

vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
```diff
@@ -1659,4 +1659,65 @@ public int commitCount() throws IOException {
...
+    public Optional<List<String>> stagedFile(Path path) {
+        try (var p = capture("git", "cat-file", "-p", ":" + path.toString())) {
+            var res = p.await();
+            if (res.status() == 0) {
+                return Optional.of(res.stdout());
+            }
+        }
+        return Optional.empty();
+    }
+
```
erikj79, line 1684:
```suggestion
    /**
     * Creates a fake Commit instance representing the currently staged diff.
     */
```

vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
```diff
@@ -1659,4 +1659,65 @@ public int commitCount() throws IOException {
...
+    @Override
+    public Commit staged() throws IOException {
+        var author = new Author("jcheck", "jcheck@none.none");
+        var commitMetaData = new CommitMetadata(new Hash("staged"), List.of(head()), author, ZonedDateTime.now(),
+                author, ZonedDateTime.now(), List.of(""));
+        return new Commit(commitMetaData, List.of(diffStaged()));
+    }
+
```
erikj79, line 1692:
```suggestion
    /**
     * Creates a fake Commit instance representing the current working tree.
     */
```

vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
erikj79: Zhao is correct, this is completely read-only in regards to the local Git repository.

vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
```diff
@@ -1659,4 +1659,65 @@ public int commitCount() throws IOException {
...
+            var res = p.await();
+            if (res.status() == 0 && res.stdout().size() == 1) {
+                return Optional.of(new Hash(res.stdout().get(0)));
+            }
+            return Optional.empty();
+        }
+    }
+
+    @Override
+    public Optional<List<String>> stagedFile(Path path) {
```
erikj79, line 1675:
```suggestion
    public Optional<List<String>> stagedFileContents(Path path) {
```

erikj79 (commented):
There are a lot of minor things, but all in all this turned out better than I had imagined. Looking forward to using it. 

Have you tried all the options against a repository with various degrees of local changes in it?

cli/src/main/java/org/openjdk/skara/cli/GitJCheck.java
zhaosongzs: Will change it

vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
zhaosongzs: > There are a lot of minor things, but all in all this turned out better than I had imagined. Looking forward to using it.
> 
> Have you tried all the options against a repository with various degrees of local changes in it?

Yes,  I tried all the options and all the combination of options.

vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
erikj79: What I mean is, you could use resolve instead of introducing this new method.

zhaosongzs: Oh, I don't know why I didn't see some of your review yesterday. Ah, right, I will use resolve and delete the new method

erikj79 (approved):

PR #1433, opened 2022-12-05
1713: Fix gradle deprecations

erikj79:
This patch tries to fix the deprecation warnings currently printed by Gradle when building Skara.

* enabled -> required: a simple rename of an attribute
* TestReport was a little trickier, but as far as I can tell, this is how it's supposed to be done now, and it works just as well/bad as it did before this change. It's not clear to me what the testReport task is good for here, as it seems to me that it will only run if all tests are successful.
* ModulePlugin.java: Was using some deprecated APIs. While at it I let Intellij fix some things that it was warning about. The null checks I removed were weird as the returned object was essentially an `Optional` (a different implementation of the same thing). Switched to using the modern instanceof assignment.
* I updated gradle-wrapper.properties to match the version used by the Skara gradlew wrapper script. Intellij will by default pick the version from the properties file, so as long as we have that file around, we should make sure they match. I needed to do this to make the build succeed in Intellij as some of these changes didn't work in Gradle 7.2.

---------
### Progress

### Issue
 * [SKARA-1713](https://bugs.openjdk.org/browse/SKARA-1713): Fix gradle deprecations

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1433/head:pull/1433` \
`$ git checkout pull/1433`

`$ git checkout pull/1433` \
`$ git pull https://git.openjdk.org/skara pull/1433/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1433`

`$ git pr show -t 1433`

</details>

<a href="https://git.openjdk.org/skara/pull/1433.diff">https://git.openjdk.org/skara/pull/1433.diff</a>

</details>

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -80,7 +80,7 @@ configure(subprojects.findAll() { it.name != 'bots' }) {
             exceptionFormat "full"
         }
 
-        reports.html.enabled = false
+        reports.html.required = false
     }
 
     tasks.withType(Test).configureEach {
@@ -92,8 +92,8 @@ configure(subprojects.findAll() { it.name != 'bots' }) {
     }
 
     tasks.withType(Test).configureEach {
-        reports.html.enabled = false
-        reports.junitXml.enabled = false
+        reports.html.required = false
+        reports.junitXml.required = false
     }
 
     tasks.withType(JavaCompile).configureEach {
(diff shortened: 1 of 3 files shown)


magicus (approved):
LGTM. Thanks for doing the house cleaning stuff.

PR #1435, opened 2022-12-05
1707: Bot warns about "No .jcheck/conf found" when it should not

zhaosongzs:
In [SKARA-1393](https://bugs.openjdk.org/browse/SKARA-1393), I added some checks to ensure the target branch of a pr contains valid jcheck configuration. This change went live on November 28, however, some users found that the warning of "No .jcheck/conf found" printed after they integrated their pr. After more investigation, I am thinking it's maybe a GitLab bug. Since when the issue happens, GitLab will always return "Commit Not Found", so we could have a temporary workaround right now.

For GitLab REST API:
If commit not found, it will return "404 Commit Not Found"

If file not found it will return "404 File Not Found"

For Github REST API:
If commit not found, it will return "No commit found for the ref "

If file not found it will return "Not Found"

---------
### Progress

### Issue
 * [SKARA-1707](https://bugs.openjdk.org/browse/SKARA-1707): Bot warns about "No .jcheck/conf found" when it should not

### Reviewers
 * [Magnus Ihse Bursie](https://openjdk.org/census#ihse) (@magicus - **Reviewer**) ⚠️ Review applies to [d800c86c](https://git.openjdk.org/skara/pull/1435/files/d800c86c004f5f2cbae6b503f984ae095f207422)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1435/head:pull/1435` \
`$ git checkout pull/1435`

`$ git checkout pull/1435` \
`$ git pull https://git.openjdk.org/skara pull/1435/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1435`

`$ git pr show -t 1435`

</details>

<a href="https://git.openjdk.org/skara/pull/1435.diff">https://git.openjdk.org/skara/pull/1435.diff</a>

</details>

### DIFF
--- a/bots/common/src/main/java/org/openjdk/skara/bots/common/BotUtils.java
+++ b/bots/common/src/main/java/org/openjdk/skara/bots/common/BotUtils.java
@@ -27,7 +27,6 @@ import java.util.Optional;
 import org.openjdk.skara.forge.PullRequest;
 import org.openjdk.skara.jbs.JdkVersion;
 import org.openjdk.skara.jcheck.JCheckConfiguration;
-import org.openjdk.skara.network.UncheckedRestException;
 
 /**
  * This class contains utility methods used by more than one bot. These methods
--- a/bots/common/src/main/java/org/openjdk/skara/bots/common/BotUtils.java
+++ b/bots/common/src/main/java/org/openjdk/skara/bots/common/BotUtils.java
@@ -41,9 +41,9 @@ public class BotUtils {
      * repositories where the fix version is configured in .jcheck/conf.
      */
     public static Optional<JdkVersion> getVersion(PullRequest pr) {
-        String confFile;
-        confFile = pr.repository().fileContents(".jcheck/conf", pr.headHash().hex())
-                .orElse(pr.repository().fileContents(".jcheck/conf", pr.targetRef()).orElseThrow());
+        var confFile = pr.repository().fileContents(".jcheck/conf", pr.headHash().hex())
+                .orElse(pr.repository().fileContents(".jcheck/conf", pr.targetRef())
+                        .orElseThrow(() -> new RuntimeException("Could not find .jcheck/conf in neither src or target of PR " + pr)));
         var configuration = JCheckConfiguration.parse(confFile.lines().toList());
         var version = configuration.general().version().orElse(null);
         if (version == null || "".equals(version)) {
--- a/bots/common/src/main/java/org/openjdk/skara/bots/common/BotUtils.java
+++ b/bots/common/src/main/java/org/openjdk/skara/bots/common/BotUtils.java
@@ -23,6 +23,7 @@
 package org.openjdk.skara.bots.common;
 
 import java.util.Optional;
+
 import org.openjdk.skara.forge.PullRequest;
 import org.openjdk.skara.jbs.JdkVersion;
 import org.openjdk.skara.jcheck.JCheckConfiguration;
@@ -42,15 +43,8 @@ public class BotUtils {
      */
     public static Optional<JdkVersion> getVersion(PullRequest pr) {
         String confFile;
-        try {
-            confFile = pr.repository().fileContents(".jcheck/conf", pr.headHash().hex());
-        } catch (UncheckedRestException e) {
-            if (e.getStatusCode() == 404) {
-                confFile = pr.repository().fileContents(".jcheck/conf", pr.targetRef());
-            } else {
-                throw e;
-            }
-        }
+        confFile = pr.repository().fileContents(".jcheck/conf", pr.headHash().hex())
+                .orElse(pr.repository().fileContents(".jcheck/conf", pr.targetRef()).orElseThrow());
         var configuration = JCheckConfiguration.parse(confFile.lines().toList());
         var version = configuration.general().version().orElse(null);
         if (version == null || "".equals(version)) {
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/LimitedCensusInstance.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/LimitedCensusInstance.java
@@ -70,13 +70,9 @@ class LimitedCensusInstance {
 
     private static Optional<JCheckConfiguration> configuration(HostedRepository remoteRepo, String name, String ref) {
         Optional<List<String>> conf = Optional.empty();
-        try {
-            conf = Optional.of(Arrays.stream(remoteRepo.fileContents(name, ref).split("\n")).toList());
-        } catch (UncheckedRestException e) {
-            // Throw the exception if the error is not exactly "File not found"
-            if (e.getStatusCode() != 404 || (!e.getBody().equals("File Not Found") && !e.getBody().equals("Not found"))) {
-                throw e;
-            }
+        var contents = remoteRepo.fileContents(name, ref);
+        if (contents.isPresent()) {
+            conf = Optional.of(Arrays.stream(contents.get().split("\n")).toList());
         }
         return conf.map(JCheckConfiguration::parse);
     }
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/LimitedCensusInstance.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/LimitedCensusInstance.java
@@ -74,7 +74,7 @@ class LimitedCensusInstance {
             conf = Optional.of(Arrays.stream(remoteRepo.fileContents(name, ref).split("\n")).toList());
         } catch (UncheckedRestException e) {
             // Throw the exception if the error is not exactly "File not found"
-            if (e.getStatusCode() != 404 || e.getBody().contains("Commit Not Found") || e.getBody().contains("No commit found")) {
+            if (e.getStatusCode() != 404 || (!e.getBody().equals("File Not Found") && !e.getBody().equals("Not found"))) {
                 throw e;
             }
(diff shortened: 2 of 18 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/LimitedCensusInstance.java
```diff
@@ -73,7 +73,8 @@ private static Optional<JCheckConfiguration> configuration(HostedRepository remo
         try {
             conf = Optional.of(Arrays.stream(remoteRepo.fileContents(name, ref).split("\n")).toList());
         } catch (UncheckedRestException e) {
-            if (e.getStatusCode() != 404) {
+            // Throw the exception if the error is not exactly "File not found"
+            if (e.getStatusCode() != 404 || e.getBody().contains("Commit Not Found") || e.getBody().contains("No commit found")) {
```
erikj79, line 77: We should check for precisely "File Not Found" or "Not found".

```suggestion
            if (e.getStatusCode() != 404 || (!e.getBody().equals("File Not Found") && !e.getBody().equals("Not found"))) {
```

erikj79 (commented):
I think this is good enough for now, but as we talked about offline, we should follow up with changing the signature of `HostedRepository::fileContents` so that it returns an Optional instead. That Optional should only be empty on 404 File Not Found and any other error should throw an exception.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/LimitedCensusInstance.java
zhaosongzs: Yes, it will be better. Thanks!

erikj79 (approved):

forge/src/main/java/org/openjdk/skara/forge/github/GitHubRepository.java
```diff
@@ -261,14 +261,23 @@ public VCS repositoryType() {
     }
 
     @Override
-    public String fileContents(String filename, String ref) {
-        var conf = request.get("contents/" + filename)
-                          .param("ref", ref)
-                          .execute().asObject();
+    public Optional<String> fileContents(String filename, String ref) {
+        JSONValue conf;
```
magicus, line 265: That's a bit odd name for the variable. I realize you did not come up with it, but I think it is a remnant from the time this function was only used to read the configuration file. I recommend changing it to `content`.

forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabRepository.java
```diff
@@ -286,21 +286,34 @@ public VCS repositoryType() {
...
-                          .onError(response -> {
-                              log.warning("First time request returned bad status: " + response.statusCode());
-                              log.info("First time response body: " + response.body());
-                              // Retry once with additional escaping of the path fragment
-                              var escapedConfName = URLEncoder.encode(confName, StandardCharsets.UTF_8);
-                              return Optional.of(request.get("repository/files/" + escapedConfName)
-                                            .param("ref", ref).execute());
-                          })
-                          .execute();
+        JSONValue conf = null;
```
magicus, line 291: The same here. Also update `confName`, perhaps `encodedFileName`?

magicus (approved):
Apart from the naming issue I think this looks good, but please await Erik's review as well.

zhaosongzs:
> Apart from the naming issue I think this looks good, but please await Erik's review as well.

Thanks for the review! I will change the variable name.

bots/common/src/main/java/org/openjdk/skara/bots/common/BotUtils.java
```diff
@@ -42,15 +42,8 @@ public class BotUtils {
...
-            confFile = pr.repository().fileContents(".jcheck/conf", pr.headHash().hex());
-        } catch (UncheckedRestException e) {
-            if (e.getStatusCode() == 404) {
-                confFile = pr.repository().fileContents(".jcheck/conf", pr.targetRef());
-            } else {
-                throw e;
-            }
-        }
+        confFile = pr.repository().fileContents(".jcheck/conf", pr.headHash().hex())
+                .orElse(pr.repository().fileContents(".jcheck/conf", pr.targetRef()).orElseThrow());
```
erikj79, line 46: Could you add a message to the exception in orElseThrow describing what's happening. Something like
```
() -> new RuntimeException("Missing .jcheck/conf in both src and target of PR " + pr)
```

Also, no need to split declaration of confFile to a separate line anymore, nor declare the type.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/LimitedCensusInstance.java
```diff
@@ -70,12 +70,9 @@ private static Namespace namespace(Census census, String hostNamespace) {
...
-            conf = Optional.of(Arrays.stream(remoteRepo.fileContents(name, ref).split("\n")).toList());
-        } catch (UncheckedRestException e) {
-            if (e.getStatusCode() != 404) {
-                throw e;
-            }
+        var contents = remoteRepo.fileContents(name, ref);
+        if (contents.isPresent()) {
+            conf = Optional.of(Arrays.stream(contents.get().split("\n")).toList());
         }
         return conf.map(JCheckConfiguration::parse);
```
erikj79, line 77: This seems a bit awkward. I think something like this should do it (untested and likely not completely right):
```
return remoteRepo.fileContents(name, ref).map(contents -> JCheckConfiguration.parse(Arrays.stream(contents.split("\n")).toList()));
```

forge/src/main/java/org/openjdk/skara/forge/github/GitHubRepository.java
```diff
@@ -261,14 +261,23 @@ public VCS repositoryType() {
...
+            content = request.get("contents/" + filename)
+                    .param("ref", ref)
+                    .execute();
+        } catch (UncheckedRestException e) {
+            // returns Optional.empty() on 404 File Not Found
+            if (e.getStatusCode() == 404 && e.getBody().contains("Not Found")) {
+                return Optional.empty();
+            }
+            throw e;
+        }
```
erikj79, line 276: Instead of catching this exception, please use the onError method on the request to register an error handler.

forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabRepository.java
```diff
@@ -286,21 +286,34 @@ public VCS repositoryType() {
...
+    public Optional<String> fileContents(String filename, String ref) {
+        var encodedFileName = URLEncoder.encode(filename, StandardCharsets.UTF_8);
+        JSONValue content;
+        try {
+            content = request.get("repository/files/" + encodedFileName)
+                    .param("ref", ref)
+                    .onError(response -> {
+                        // Retry once with additional escaping of the path fragment
+                        // Only retry when the error is exactly "File Not Found"
+                        if (response.statusCode() == 404 && response.body().contains("File Not Found")) {
```
erikj79, line 298: The comment says 'exactly "File Not Found"' but the conditional uses `.contains`. I think we can switch to `.equals`.

forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabRepository.java
```diff
@@ -286,21 +286,34 @@ public VCS repositoryType() {
...
+                        }
+                        return Optional.empty();
+                    })
+                    .execute();
+        } catch (UncheckedRestException e) {
+            // After retry, if the error is still 404 File Not Found, return Optional.empty().
+            if (e.getStatusCode() == 404 && e.getBody().contains("File Not Found")) {
+                return Optional.empty();
+            }
+            throw e;
```
erikj79, line 313: Instead of catching exception, I think it's better to register yet another `.onError` handler on the nested request object. You would then return a special JSONValue instance that you can check for and then the outer method can return `Optional.empty()`. There are other examples of this pattern in this file.

forge/src/main/java/org/openjdk/skara/forge/github/GitHubRepository.java
zhaosongzs: Sorry, I couldn't find a good way to use errorHandler with the current `RestRequest#transformBadResponse`.
We will not be able to return `Optional.empty()` when the error is 404 Not Found.

forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabRepository.java
zhaosongzs: I use contains because the real body is `{"message":"404 File Not Found"}`.  I will use  `response.body().equals("{\"message\":\"404 File Not Found\"}")`

bots/common/src/main/java/org/openjdk/skara/bots/common/BotUtils.java
magicus: But Eriks suggested message sounds a bit contradictory. I suggest instead using:
"Could not find .jcheck/conf in neither src or target of PR"

forge/src/main/java/org/openjdk/skara/forge/github/GitHubRepository.java
zhaosongzs: Ah, see your comments below. I know how to do it now. Thank you

erikj79 (approved):

PR #1436, opened 2022-12-06
1710: CommitCommandWorkItem always fetches Commit

erikj79:
During bot restart, quite a few CommitCommandWorkItems are spawned. Most of them will discover that nothing needs to be done. One of the first things done in the run method is fetching the commit from the remote repository. The Commit instance will always be created with the full set of diffs for the commit. In certain cases, where commits are very large, this can take considerable time. Specifically in the jfx repos, there are examples where the REST request is split into 147 pages. (e.g. https://github.com/openjdk/jfx11u/commit/4bd6877dbd46e37f6774c50dcdc98a8a1ab41214) The way these WorkItems are spawned on startup, most of these gigantic commits will be included for a long time. The Commit instance isn't needed to figure out if a new command is present, just the comments are. This means we can move this query to after we have established that a command is to be executed.

To further reduce overhead when fetching commits from remote repositories, I'm also making the inclusion of the "parentDiffs" optional. Of all the uses of `HostedRepository::commit`, there is currently only one that actually needs the diffs (BackportCommand). This will require the caller to be mindful of the state of the returned `Commit` object, especially when sending it other consumers, but I believe it's worth it to reduce unnecessary load on the system.

This also affects `Forge::search`, where there is one other use where the diffs are used (when evaluating if a backport is clean).

---------
### Progress

### Issue
 * [SKARA-1710](https://bugs.openjdk.org/browse/SKARA-1710): CommitCommandWorkItem always fetches Commit

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1436/head:pull/1436` \
`$ git checkout pull/1436`

`$ git checkout pull/1436` \
`$ git pull https://git.openjdk.org/skara pull/1436/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1436`

`$ git pr show -t 1436`

</details>

<a href="https://git.openjdk.org/skara/pull/1436.diff">https://git.openjdk.org/skara/pull/1436.diff</a>

</details>

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -391,7 +391,7 @@ class CheckRun {
         if (hash == null) {
             return Optional.empty();
         }
-        var commit = pr.repository().forge().search(hash);
+        var commit = pr.repository().forge().search(hash, true);
         if (commit.isEmpty()) {
             throw new IllegalStateException("Backport comment for PR " + pr.id() + " contains bad hash: " + hash.hex());
         }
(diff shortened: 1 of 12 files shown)


magicus (approved):
LGTM

PR #1439, opened 2022-12-07
1691: Run Jcheck twice in CheckRun if .jcheck/conf has changed

zhaosongzs:
When a pr is created, pr bot will run jcheck to check whether this commit is good. However, currently, we only run jcheck with the configuration from the target branch. Sometimes, user wants to modify jcheck configuration and accidentally messes up the format. In this case, pr bot would not be able to find this problem. After this commit with badly formatted jcheck configuration integrated, all the jcheck will fail and our bot gets stuck into a retry loop.

In this patch, if jcheck configuration file is updated, the bot will run jcheck twice, once with the configuration in the target and once with the new one in the source. So if the new configuration has any problem, it will became an integration blocker.

---------
### Progress

### Issue
 * [SKARA-1691](https://bugs.openjdk.org/browse/SKARA-1691): Run Jcheck twice in CheckRun if .jcheck/conf has changed

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1439/head:pull/1439` \
`$ git checkout pull/1439`

`$ git checkout pull/1439` \
`$ git pull https://git.openjdk.org/skara pull/1439/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1439`

`$ git pr show -t 1439`

</details>

<a href="https://git.openjdk.org/skara/pull/1439.diff">https://git.openjdk.org/skara/pull/1439.diff</a>

</details>


bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -1078,14 +1079,29 @@ private void checkStatus() {
             } else {
                 // Determine current status
                 var additionalConfiguration = AdditionalConfiguration.get(localRepo, localHash, pr.repository().forge().currentUser(), comments);
-                checkablePullRequest.executeChecks(localHash, censusInstance, visitor, additionalConfiguration);
+                checkablePullRequest.executeChecks(localHash, censusInstance, visitor, additionalConfiguration, true);
+                if (localRepo.isFileUpdated(".jcheck/conf")) {
```
erikj79, line 1083: We probably don't need to run the second round if confOverride is set.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckablePullRequest.java
```diff
@@ -200,12 +200,18 @@ PullRequestCheckIssueVisitor createVisitor() throws IOException {
...
     }
 
-    void executeChecks(Hash localHash, CensusInstance censusInstance, PullRequestCheckIssueVisitor visitor, List<String> additionalConfiguration) throws IOException {
+    PullRequestCheckIssueVisitor createVisitorUsingHeadHash() throws IOException {
+        var checks = JCheck.checksFor(localRepo, pr.headHash());
+        return new PullRequestCheckIssueVisitor(checks);
+    }
+
+    void executeChecks(Hash localHash, CensusInstance censusInstance, PullRequestCheckIssueVisitor visitor,
+                       List<String> additionalConfiguration, boolean useTargetConf) throws IOException {
```
erikj79, line 209: Instead of a boolean `useTargetConf`, I think it's better if we just supply the hash we want to use (target or src, if confOverride is set, then this method will handle it).

vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
```diff
@@ -1720,4 +1720,17 @@ private Diff diffStaged() throws IOException {
             throw t;
         }
     }
+
+    @Override
+    public boolean isFileUpdated(String filename) throws IOException {
```
erikj79, line 1725: This method seems a bit too specialized to me. It should at least take a Hash argument. I think it would be even better to use the existing `commits` method from which you can get to the Diff and Patch objects. That way we have properly parsed and typed data to find the filename in.

bots/pr/src/test/java/org/openjdk/skara/bots/pr/CheckTests.java
```diff
@@ -2675,4 +2675,134 @@ void testLatestBody(TestInfo testInfo) throws IOException {
...
+            var checkConf = tempFolder.path().resolve(".jcheck/conf");
+            writeToCheckConf(checkConf);
+            localRepo.add(checkConf);
+            var updateHash = localRepo.commit("enable whitespace issue check", "testauthor", "ta@none.none");
+            localRepo.push(updateHash, author.url(), "edit", true);
+
+            TestBotRunner.runPeriodicItems(checkBot);
+
+            // pr body should have the integrationBlocker for whitespace and reviewer check
+            assertTrue(pr.store().body().contains("Whitespace errors(failed with the updated jcheck configuration)"));
```
erikj79, line 2729: Please add a space before the `(` for all of these.

bots/pr/src/test/java/org/openjdk/skara/bots/pr/CheckTests.java
```diff
@@ -2675,4 +2675,134 @@ void testLatestBody(TestInfo testInfo) throws IOException {
...
+            var check = checks.get("jcheck");
+            assertEquals(CheckStatus.SUCCESS, check.status());
+            // pr body should not have the process for whitespace
+            assertFalse(pr.store().body().contains("whitespace"));
+
+            // Make .jcheck/conf invalid
+            var checkConf = tempFolder.path().resolve(".jcheck/conf");
+            try (var output = new FileWriter(checkConf.toFile(), true)) {
+                output.append("\nRandomCharacters");
+            }
```
erikj79, line 2786: This could be done with
```
Files.writeString(checkConf, "text...", StandardOpenOption.APPEND)
```

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -1078,14 +1079,29 @@ private void checkStatus() {
...
+                        PullRequestCheckIssueVisitor visitor2 = checkablePullRequest.createVisitorUsingHeadHash();
+                        log.info("Run jcheck again with the updated configuration");
+                        checkablePullRequest.executeChecks(localHash, censusInstance, visitor2, additionalConfiguration, false);
+                        secondJCheckMessage.addAll(visitor2.messages().stream()
+                                .map(StringBuilder::new)
+                                .map(e -> e.append("(failed with the updated jcheck configuration)"))
+                                .map(StringBuilder::toString)
+                                .toList());
+                    } catch (Exception e) {
+                        secondJCheckMessage.add(e.getMessage() + "(exception thrown when running jcheck with updated jcheck configuration)");
```
erikj79, line 1094: Can we log this exception?

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
zhaosongzs: Good point!

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckablePullRequest.java
zhaosongzs: Yeah, pass hash is better. But even if we pass hash here, it won't affect the logic of using confOverride because confOverride is set in the constructor of `CheckablePullRequest`

bots/pr/src/test/java/org/openjdk/skara/bots/pr/CheckTests.java
zhaosongzs: It's better! Thx

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckablePullRequest.java
zhaosongzs: Now, I prefer not to change this method because I think ` CheckablePullRequest#targetHash` can help us reuse the target hash. If we add hash as an argument, we will need to call `PullRequestUtils.targetHash(localRepo)` more than once.

vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
```diff
@@ -1720,4 +1720,13 @@ private Diff diffStaged() throws IOException {
...
     }
+
+    @Override
+    public boolean isFileUpdated(String filename, Hash h) throws IOException {
+        return commits().asList().stream()
+                .filter(commit -> commit.hash().equals(h))
+                .anyMatch(commit -> commit.parentDiffs().stream()
+                        .anyMatch(diff -> diff.patches().stream()
+                                .anyMatch(patch -> patch.toString().contains(filename))));
+    }
```
zhaosongzs, line 1731: @erikj79 I am not sure I get you correctly. Correct me if I misunderstand what you mean.

erikj79: I wasn't very clear. I don't think we need this method on GitRepository. I figured we could just inline this at the single call site in CheckRun, but since the expression is quite long, then probably a private method in CheckRun would be best.

You can use `commits(String range)` to just get the single commit we are interested in.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckablePullRequest.java
erikj79: `CheckablePullRequest#targetHash` is public so you can call it from `CheckRun`.

zhaosongzs: Oh, then I will change the method. Thanks.

magicus:
Are we sure we are not painting ourselves in a corner were a change in .jcheck/conf is impossible to do since it will fail either before or after applying? I can't really think of such example, but maybe I'm just not trying hard enough.

zhaosongzs:
> Are we sure we are not painting ourselves in a corner were a change in .jcheck/conf is impossible to do since it will fail either before or after applying? I can't really think of such example, but maybe I'm just not trying hard enough.

I am not sure but I couldn't think of such example too.

erikj79:
> Are we sure we are not painting ourselves in a corner were a change in .jcheck/conf is impossible to do since it will fail either before or after applying? I can't really think of such example, but maybe I'm just not trying hard enough.

It's possible, but in that case the contributor will have to split the change into two separate changes. First relax existing requirements and then add the new requirement.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -1178,4 +1197,11 @@ private void checkStatus() {
             throw new RuntimeException("Exception during jcheck", checkException);
         }
     }
+
+    private boolean isFileUpdated(String filename, Hash hash) throws IOException {
+        return localRepo.commits(hash.hex()).asList().stream()
```
erikj79:
```suggestion
        return localRepo.commits(hash.hex()).stream()
```

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -1178,4 +1197,11 @@ private void checkStatus() {
...
+                .anyMatch(commit -> commit.parentDiffs().stream()
+                        .anyMatch(diff -> diff.patches().stream()
+                                .anyMatch(patch -> patch.toString().contains(filename))));
```
erikj79, line 1205:
```suggestion
                                .anyMatch(patch -> patch.source().path().toString().equals(filename)
                                    || patch.target().path().toString().equals(filename))));
```

forge/src/main/java/org/openjdk/skara/forge/PullRequestUtils.java
```diff
@@ -165,6 +165,7 @@ public static Hash targetHash(Repository localRepo) throws IOException {
     public static Repository materialize(HostedRepositoryPool hostedRepositoryPool, PullRequest pr, Path path) throws IOException {
         var localRepo = hostedRepositoryPool.checkout(pr.repository(), pr.headHash().hex(), path);
         localRepo.fetch(pr.repository().url(), "+" + pr.targetRef() + ":prutils_targetref", false);
+        localRepo.fetch(pr.repository().url(), "+" + pr.sourceRef() + ":prutils_sourceref", false);
```
erikj79, line 168: I'm curious as to why this is needed. Line 166 should make sure we are fetching the source/head of the pr. I don't see any use of the local branch `prutils_sourceref`.

zhaosongzs: Sorry, I forget to delete it. I was thinking we need this and then I found it is useless, so I didn't use it any more. But I forget to delete it later.

erikj79 (approved):

zhaosongzs:
Thanks for review!
/integrate

PR #1442, opened 2022-12-12
1598: GitRepository.isHealthy takes a long time in ArchiveWorkItem

erikj79:
In the mlbridge bot, the `ArchiveWorkItem` is spending an inordinate amount of time running `git fsck --connectivity-only`. This is part of a rather sane strategy of verifying a repository before using it. This is done when a repository is materialized using the `HostedRepositoryPool`, if a previous local copy is found and reused, it's first checked using `Repository::isHealthy`, which runs the fsck. In the `ArchiveWorkItem` case, the repository in question is the archive repository, where all sent emails are stored for reference (and to prevent resending duplicate messages). The repository is materialized first in every `ArchiveWorkItem::run` call as it needs to read the current archive for the PR. Especially on bot restart, this triggers a lot of extra work.

This repository is growing quite fast as it contains every PR for every repository we have Skara running against on GitHub. Having every ArchiveWorkItem verifying all this contents just doesn't scale well, when each work item will only ever read and write to a single file in the repository.

To eliminate this inefficiency, I'm changing `ArchiveWorkItem` to read and write the single file over REST API instead. This does introduce another level of inefficiency, where the complete file must be transferred to and from the server each time, instead of just diffs, but I think given the size of the repo, and the rate at which it is growing (not to mention the fsck checks), this is the better tradeoff. 

The patch adds a new method `HostedRepository::writeFileContents` (that pairs with the existing `::fileContents`). I've added manual tests that verifies these two methods together for both GitHub and GitLab. This prompted rewriting of `GitHubRepository::fileContents` to use "raw" mode for retrieving file contents. This in turn made me discover that we could remove all the preview feature `Accept` headers, as all those GitHub APIs are stable by now. (There is currently a bug in `RestRequest` causing only the last header with the same key to be used, so only the last of the list of preview headers was ever actually in effect. This bug was also preventing me from using raw mode, as that requires another `Accept` header. Given this, and that the mockingbird API is verified to be stable now, I'm confident in removing all of them. I will file a separate issue for the bug in `RestRequest`.)

To use "raw" mode, I needed to use `RestRequest::executeUnparsed` and discovered that it wasn't behaving quite the same as the normal `::execute`. I changed it to throw `UncheckedRestException` like its sibling. I inspected all uses of the method and made adjustments where needed.

`TestHostedRepository::fileContents` didn't return the raw contents of files. Specifically it didn't preserve trailing newlines. To fix this, I changed it to no longer convert to lines and then joining them again.

---------
### Progress

### Issue
 * [SKARA-1598](https://bugs.openjdk.org/browse/SKARA-1598): GitRepository.isHealthy takes a long time in ArchiveWorkItem

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1442/head:pull/1442` \
`$ git checkout pull/1442`

`$ git checkout pull/1442` \
`$ git pull https://git.openjdk.org/skara pull/1442/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1442`

`$ git pr show -t 1442`

</details>

<a href="https://git.openjdk.org/skara/pull/1442.diff">https://git.openjdk.org/skara/pull/1442.diff</a>

</details>

### DIFF
--- a/bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveWorkItem.java
+++ b/bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveWorkItem.java
@@ -239,18 +239,17 @@ class ArchiveWorkItem implements WorkItem {
         return ret.toString();
     }
 
+    private String mboxFile() {
+        return bot.codeRepo().name() + "/" + pr.id() + ".mbox";
+    }
+
     @Override
     public Collection<WorkItem> run(Path scratchPath) {
         var path = scratchPath.resolve("mlbridge");
-        var archiveRepo = materializeArchive(path);
-        var mboxBasePath = path.resolve(bot.codeRepo().name());
 
         var sentMails = new ArrayList<Email>();
-        try {
-            var archiveContents = Files.readString(mboxBasePath.resolve(pr.id() + ".mbox"), StandardCharsets.UTF_8);
-            sentMails.addAll(Mbox.splitMbox(archiveContents, bot.emailAddress()));
-        } catch (IOException ignored) {
-        }
+        var archiveContents = bot.archiveRepo().fileContents(mboxFile(), bot.archiveRef());
+        archiveContents.ifPresent(s -> sentMails.addAll(Mbox.splitMbox(s, bot.emailAddress())));
 
         var labels = new HashSet<>(pr.labelNames());
 
@@ -382,14 +381,17 @@ class ArchiveWorkItem implements WorkItem {
             }
 
             // Push all new mails to the archive repository
-            var mbox = MailingListServerFactory.createMboxFileServer(mboxBasePath);
+            var newArchivedContents = new StringBuilder();
+            archiveContents.ifPresent(newArchivedContents::append);
             for (var newMail : newMails) {
                 var forArchiving = Email.from(newMail)
                                         .recipient(EmailAddress.from(pr.id() + "@mbox"))
                                         .build();
-                mbox.post(forArchiving);
+                newArchivedContents.append(Mbox.fromMail(forArchiving));
             }
-            pushMbox(archiveRepo, "Adding comments for PR " + bot.codeRepo().name() + "/" + pr.id());
+            bot.archiveRepo().writeFileContents(newArchivedContents.toString(), mboxFile(), new Branch(bot.archiveRef()),
+                    "Adding comments for PR " + bot.codeRepo().name() + "/" + pr.id(),
+                    bot.emailAddress().fullName().orElseThrow(), bot.emailAddress().address());
 
             // Finally post all new mails to the actual list
             for (var newMail : newMails) {
--- a/bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveWorkItem.java
+++ b/bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveWorkItem.java
@@ -32,7 +32,6 @@ import org.openjdk.skara.mailinglist.*;
 import org.openjdk.skara.vcs.*;
 
 import java.io.*;
-import java.nio.charset.StandardCharsets;
 import java.nio.file.*;
 import java.time.*;
 import java.util.*;
@@ -248,6 +247,7 @@ class ArchiveWorkItem implements WorkItem {
         var path = scratchPath.resolve("mlbridge");
 
         var sentMails = new ArrayList<Email>();
+        // Load in already sent emails from the archive, if there are any.
         var archiveContents = bot.archiveRepo().fileContents(mboxFile(), bot.archiveRef());
         archiveContents.ifPresent(s -> sentMails.addAll(Mbox.splitMbox(s, bot.emailAddress())));
 
@@ -389,7 +389,7 @@ class ArchiveWorkItem implements WorkItem {
                                         .build();
                 newArchivedContents.append(Mbox.fromMail(forArchiving));
             }
-            bot.archiveRepo().writeFileContents(newArchivedContents.toString(), mboxFile(), new Branch(bot.archiveRef()),
+            bot.archiveRepo().writeFileContents(mboxFile(), newArchivedContents.toString(), new Branch(bot.archiveRef()),
                     "Adding comments for PR " + bot.codeRepo().name() + "/" + pr.id(),
                     bot.emailAddress().fullName().orElseThrow(), bot.emailAddress().address());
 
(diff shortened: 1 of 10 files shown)


bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveWorkItem.java
```diff
@@ -239,18 +239,17 @@ private String subjectPrefix() {
...
-        var mboxBasePath = path.resolve(bot.codeRepo().name());
 
         var sentMails = new ArrayList<Email>();
-        try {
-            var archiveContents = Files.readString(mboxBasePath.resolve(pr.id() + ".mbox"), StandardCharsets.UTF_8);
-            sentMails.addAll(Mbox.splitMbox(archiveContents, bot.emailAddress()));
-        } catch (IOException ignored) {
-        }
+        var archiveContents = bot.archiveRepo().fileContents(mboxFile(), bot.archiveRef());
+        archiveContents.ifPresent(s -> sentMails.addAll(Mbox.splitMbox(s, bot.emailAddress())));
```
magicus, line 252: So this adds the old contents to sentMails, and if there is none, we just ignore adding it? I needed some time to figure out why it was safe doing nothing if archiveContents was empty. Perhaps a comment about this?

forge/src/main/java/org/openjdk/skara/forge/HostedRepository.java
```diff
@@ -87,6 +87,17 @@ PullRequest createPullRequest(HostedRepository target,
...
+    /**
+     * Writes new contents to a file in the repo by creating a new commit.
+     * @param content New file content to write, always replacing existing content
+     * @param filename Name of file inside repository to write to
+     * @param branch Branch to add commit on top of
+     * @param message Commit message
+     * @param authorName Name of author and committer for commit
+     * @param authorEmail Email of author and committer for commit
+     */
+    void writeFileContents(String content, String filename, Branch branch, String message, String authorName, String authorEmail);
```
magicus, line 100: Maybe not a big deal, but I thought the symmetry with `fileContents` was a bit lost when `String filename` was not the first argument to this method.

forge/src/main/java/org/openjdk/skara/forge/github/GitHubRepository.java
```diff
@@ -263,18 +260,49 @@ public VCS repositoryType() {
...
+        // Get file contents using raw format. This allows us to get files of
+        // size up to 10MB (up from 1MB if getting in object from).
+        try {
+            var content = request.get("contents/" + filename)
+                    .param("ref", ref)
+                    .header("Accept", "application/vnd.github.raw+json")
+                    .executeUnparsed();
+            return Optional.of(content);
+        } catch (UncheckedRestException e) {
+            // The onError handler is not used with executeParsed, so have to
```
magicus, line 272:
```suggestion
            // The onError handler is not used with executeUnparsed, so have to
```

forge/src/main/java/org/openjdk/skara/forge/github/GitHubRepository.java
```diff
@@ -263,18 +260,49 @@ public VCS repositoryType() {
...
+
+    @Override
+    public void writeFileContents(String content, String filename, Branch branch, String message, String authorName, String authorEmail) {
+        var body = JSON.object()
+                .put("message", message)
+                .put("branch", branch.name())
+                .put("committer", JSON.object()
+                        .put("name", authorName)
+                        .put("email", authorEmail))
+                .put("content", new String(Base64.getEncoder().encode(content.getBytes(StandardCharsets.UTF_8)), StandardCharsets.UTF_8));
```
magicus, line 292: Are we talking about 10's of MB here as well? We don't risk running into limits?

forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabRepository.java
```diff
@@ -313,6 +313,31 @@ public Optional<String> fileContents(String filename, String ref) {
...
+                .put("content", new String(Base64.getEncoder().encode(content.getBytes(StandardCharsets.UTF_8)), StandardCharsets.UTF_8));
+        request.put("repository/files/" + encodedFileName)
+                .body(body)
+                .onError(response -> {
+                    // Gitlab requires POST for creating new files and PUT for updating existing.
+                    // Retry with POST if we get 400 response.
+                    if (response.statusCode() == 400) {
+                        return Optional.of(request.post("repository/files/" + encodedFileName)
+                                .body(body)
+                                .execute());
```
magicus, line 334: You don't need an `onError` to return `Optional.empty()` for the POST case? Or is it the default if you don't specify an error handler?

magicus (approved):

bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveWorkItem.java
erikj79: I didn't reflect on that as it's the same logic as before, only before it was handled with the ignored `IOException`, but I can add a comment.

forge/src/main/java/org/openjdk/skara/forge/HostedRepository.java
erikj79: Sure, makes sense.

forge/src/main/java/org/openjdk/skara/forge/github/GitHubRepository.java
erikj79: Hm, I just realized the comment on fileContents is wrong, with raw we get up to 100MB, so I will fix that. I can't find any documentation regarding a limit when writing file contents. The manual test I have tests up to 10MB, which was fine. The biggest file in our current archive is just over 1MB. The current archive is however hosted on GitLab and not GitHub, so the GitHub limits aren't really relevant at this time.

forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabRepository.java
erikj79: The default for `execute()` is to throw `UncheckedRestException` if we get any kind of error response from the server. Here I made the assumption that replacing files is more common than creating new files, and simply first try PUT to replace a file and if that fails with the documented error code 400, we retry with POST to create a new file. If the POST fails, then we just let it fail.

Another way of doing it would have been to first call GET to see if the file exists and then pick either PUT or POST depending on the result. I just tried to keep the number of calls down for the current common case.

PR #1446, opened 2022-12-20
1699: CommitCommentsWorkItem does excessive unnecessary work

erikj79:
The `CommitCommentsWorkItem` gets scheduled once for each repository by `PullRequestBot::getPeriodicItems`. It's responsible for querying the repository for new comments on commits, and if any are found, spawn `CommitCommandWorkItem` for them. It's unfortunately not very straight forward to query forges for new commit comments. For GitHub we have a rather complicated GraphQL that does the trick, but for GitLab, we need to jump through some rather nasty hoops.

The main issue is that the data we get from the "events" query from GitLab contains notes posted to commits, but without the commit hash. We only get the commit "title" to identify which commit was commented on. To solve this, we have a rather elaborate logic that first builds a complete map from title to set of hashes for every commit in the repo. This is done preemptively on a local clone of the repo. Using this map and some clever comparisons, we can figure out which commit each comment belongs to.

I don't have data on how long it takes to build this map for a big repository, like the JDK, but it's likely not trivially short. In addition to this, before using any local clone of a repository, we always run the git fsck check (see [SKARA-1598](https://bugs.openjdk.org/browse/SKARA-1598)), which can take around 20s for the JDK repo. The local clone is unfortunately also used to check if the commits for the found comments belong to a valid branch. However, both of these operations are read only, so we shouldn't really need a clone for this. We should be fine just using the seed repository directly from the `HostedRepositoryPool`.

This patch tries to make this situation better in several ways.

1. Use the seed repository instead of a clone in `CommitCommentsWorkItem`. (Removes ~20s fsck time every time this is scheduled for a jdk repository, which is in the order of once a minute times the numer of JDK repo clones that Skara operates on)
2. Refactored `HostedRepository::recentCommitComments` to take a `ReadOnlyRepository` instead of the pre calculated map. `GitLabRepository` is then responsible for using this repository to build the map itself.
3. `HostedRepository::recentCommitComments` now takes an `updatedAfter` arg to limit the number of comments that get returned. In `GitLabRepository` this was already limited to last 4 days, but in `GitHubRepository`, we just got the last 100 comments, regardless of age (due to a limitation in GraphQL). We can still filter the returned comments on updatedAt so that we get less comments to process in many cases. This should make a big impact on bot restarts. 
4. `CommitCommentsWorkItem` has a configuration option to ignore comments from certain users (typically meant for bot users, e.g. the notifier who posts a comment on every commit). This configuration wasn't working due to a bug in the factory. (Together with 3, should bring down the number of scheduled `CommitCommandWorkItem` on bot restart from 100 per repo to a handful and often 0.)
5. `GitLabRepository` keeps a cached copy of the commitTitleToHash map. It will also only build or update the map if any new notes are actually found. (This won't have a huge impact, but should speed things up a little bit more)

---------
### Progress

### Issue
 * [SKARA-1699](https://bugs.openjdk.org/browse/SKARA-1699): CommitCommentsWorkItem does excessive unnecessary work

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1446/head:pull/1446` \
`$ git checkout pull/1446`

`$ git checkout pull/1446` \
`$ git pull https://git.openjdk.org/skara pull/1446/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1446`

`$ git pr show -t 1446`

</details>

<a href="https://git.openjdk.org/skara/pull/1446.diff">https://git.openjdk.org/skara/pull/1446.diff</a>

</details>

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CommitCommentsWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CommitCommentsWorkItem.java
@@ -22,6 +22,8 @@
  */
 package org.openjdk.skara.bots.pr;
 
+import java.time.Duration;
+import java.time.ZonedDateTime;
 import org.openjdk.skara.bot.WorkItem;
 import org.openjdk.skara.forge.*;
 import org.openjdk.skara.vcs.*;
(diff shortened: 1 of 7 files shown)


zhaosongzs:
I want to verify my understanding of the` HostedRepositoryPool#materialize` and `HostedRepositoryPool#seedRepository` methods.  As I understand it, the first method creates a new directory separate from the seed directory and clones the hosted repository into it, allowing read and write operations to be performed on the cloned repository. The second method simply returns the updated seed repository, which does not allow write operations and also does not require a health check before being used.

zhaosongzs (approved):
LGTM! I believe this patch will bring us some performance improvements.

erikj79:
> I want to verify my understanding of the` HostedRepositoryPool#materialize` and `HostedRepositoryPool#seedRepository` methods. As I understand it, the first method creates a new directory separate from the seed directory and clones the hosted repository into it, allowing read and write operations to be performed on the cloned repository. The second method simply returns the updated seed repository, which does not allow write operations and also does not require a health check before being used.

That is how I see it. However, the `seedRepository` method doesn't return a `ReadOnlyRepository` (which would have been nicer), and I couldn't make it do so, because the pr branch notifier is currently using the seed repository for creating the pr branch. I don't think this is good behavior. That bot should probably materialize a clone instead.

One could also argue that we really shouldn't be touching the seed repo at all. Maybe we could have two levels of materialize, something like `materializeReadOnly` that would return a `ReadOnlyRepository` and that method wouldn't need to perform as much health checking. 

But for now, I think this is a reasonable change for this bot.

zhaosongzs:
> > I want to verify my understanding of the` HostedRepositoryPool#materialize` and `HostedRepositoryPool#seedRepository` methods. As I understand it, the first method creates a new directory separate from the seed directory and clones the hosted repository into it, allowing read and write operations to be performed on the cloned repository. The second method simply returns the updated seed repository, which does not allow write operations and also does not require a health check before being used.
> 
> That is how I see it. However, the `seedRepository` method doesn't return a `ReadOnlyRepository` (which would have been nicer), and I couldn't make it do so, because the pr branch notifier is currently using the seed repository for creating the pr branch. I don't think this is good behavior. That bot should probably materialize a clone instead.
> 
> One could also argue that we really shouldn't be touching the seed repo at all. Maybe we could have two levels of materialize, something like `materializeReadOnly` that would return a `ReadOnlyRepository` and that method wouldn't need to perform as much health checking.
> 
> But for now, I think this is a reasonable change for this bot.

Agree with you!

magicus (approved):
Looks good.

Did you measure any speed improvements?

erikj79:
> Looks good.
Thanks!
> Did you measure any speed improvements?
I haven't run this live yet, so won't know the real impact until we deploy this for real.

PR #1450, opened 2022-12-25
GitRepository#revert does not work as intended

TheShermanTanker:
The `Repository#revert` interface is meant to allow for rolling the Repository back to a specific commit, but the Git implementation in `GitRepository#revert` implements this via a checkout, which can cause issues, especially with newer versions of Git where it doesn't have any effect at all. More appropriate would be using `git restore --source` to implement the GitRepository variant instead, which is actually the encouraged way to do so nowadays

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1450/head:pull/1450` \
`$ git checkout pull/1450`

`$ git checkout pull/1450` \
`$ git pull https://git.openjdk.org/skara pull/1450/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1450`

`$ git pr show -t 1450`

</details>

<a href="https://git.openjdk.org/skara/pull/1450.diff">https://git.openjdk.org/skara/pull/1450.diff</a>

</details>

### DIFF
--- a/vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
+++ b/vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
@@ -469,7 +469,7 @@ public class GitRepository implements Repository {
 
     @Override
     public void revert(Hash h) throws IOException {
-        try (var p = capture("git", "checkout", "--recurse-submodules", h.hex(), "--", ".")) {
+        try (var p = capture("git", "restore", "--recurse-submodules", "--source", h.hex(), "--", ".")) {
             await(p);
         }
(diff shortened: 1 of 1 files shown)


erikj79 (approved):
This looks correct. It doesn't seem like we are using this call anywhere currently for git repos, so this change shouldn't affect any current functionality.

TheShermanTanker:
Thanks for the review Erik

TheShermanTanker:
@erikj79 Could I trouble you for a sponsor? Sorry for the trouble!

PR #1452, opened 2023-01-03
1173: Dependent PR feature does not work with protected branches in newer Gitlab version

erikj79:
In GitLab, if a branch is protected, it cannot be removed. This is preventing us from using the pr/X feature of Skara to support dependent pull requests. This patch adds the ability to dynamically protect and unprotect branch patterns when creating pr/X branches. This is controlled with a new configuration option on the prbranch notifier like this:

```
        "prbranch": {
          "protect": true,
        },
```
The semantics needed for GitLab is to first add a pattern based protection rule, then create the branch. Unfortunately, I couldn't easily implement the same semantics for GitHubRepository. The REST API for branch protection in GitHub doesn't allow patterns, only existing branches. It could probably be done using GraphQL, but given how complicated that would be to implement, I chose not to, since we don't actually need this feature on GitHub. (GitHub branch protection is flexible enough so we can have a static configuration that allows admins to delete branches, and this is already in use today)

I have tested the basic branch protection functionality using the new manual test and I have run the notifier bot with the new configuration against a playground repository in GitLab and manually created and closed PRs.

---------
### Progress

### Issue
 * [SKARA-1173](https://bugs.openjdk.org/browse/SKARA-1173): Dependent PR feature does not work with protected branches in newer Gitlab version

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1452/head:pull/1452` \
`$ git checkout pull/1452`

`$ git checkout pull/1452` \
`$ git pull https://git.openjdk.org/skara pull/1452/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1452`

`$ git pr show -t 1452`

</details>

<a href="https://git.openjdk.org/skara/pull/1452.diff">https://git.openjdk.org/skara/pull/1452.diff</a>

</details>

### DIFF
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/prbranch/PullRequestBranchNotifier.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/prbranch/PullRequestBranchNotifier.java
@@ -33,10 +33,12 @@ import java.util.logging.Logger;
 
 public class PullRequestBranchNotifier implements Notifier, PullRequestListener {
     private final Path seedFolder;
+    private final boolean protectBranches;
     private final Logger log = Logger.getLogger("org.openjdk.skara.bots.notify");
 
-    public PullRequestBranchNotifier(Path seedFolder) {
+    public PullRequestBranchNotifier(Path seedFolder, boolean protectBranches) {
         this.seedFolder = seedFolder;
+        this.protectBranches = protectBranches;
     }
 
     @Override
@@ -50,6 +52,10 @@ public class PullRequestBranchNotifier implements Notifier, PullRequestListener 
             var seedRepo = hostedRepositoryPool.seedRepository(pr.repository(), false);
             seedRepo.fetch(pr.repository().url(), pr.headHash().hex());
             String branch = PreIntegrations.preIntegrateBranch(pr);
+            if (protectBranches) {
+                log.info("Protecting branch " + branch);
+                pr.repository().protectBranchPattern(branch);
+            }
             log.info("Creating new pull request pre-integration branch " + branch);
             seedRepo.push(pr.headHash(), pr.repository().url(), branch, true);
         } catch (IOException e) {
@@ -60,6 +66,10 @@ public class PullRequestBranchNotifier implements Notifier, PullRequestListener 
     private void deleteBranch(PullRequest pr) {
         String branch = PreIntegrations.preIntegrateBranch(pr);
         var branchExists = pr.repository().branchHash(branch).isPresent();
+        if (protectBranches) {
+            log.info("Removing branch protection for " + branch);
+            pr.repository().unprotectBranchPattern(branch);
+        }
         if (!branchExists) {
             log.info("Pull request pre-integration branch " + branch + " doesn't exist on remote - ignoring");
             return;
--- a/bots/tester/src/test/java/org/openjdk/skara/bots/tester/InMemoryHostedRepository.java
+++ b/bots/tester/src/test/java/org/openjdk/skara/bots/tester/InMemoryHostedRepository.java
@@ -173,6 +173,14 @@ class InMemoryHostedRepository implements HostedRepository {
     }
 
     @Override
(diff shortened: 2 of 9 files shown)


bots/notify/src/main/java/org/openjdk/skara/bots/notify/prbranch/PullRequestBranchNotifier.java
```diff
@@ -50,6 +52,10 @@ private void pushBranch(PullRequest pr) {
             var seedRepo = hostedRepositoryPool.seedRepository(pr.repository(), false);
             seedRepo.fetch(pr.repository().url(), pr.headHash().hex());
             String branch = PreIntegrations.preIntegrateBranch(pr);
+            if (protectBranches) {
+                log.info("Protecting branch " + branch);
+                pr.repository().protectBranchPattern(branch);
+            }
             log.info("Creating new pull request pre-integration branch " + branch);
             seedRepo.push(pr.headHash(), pr.repository().url(), branch, true);
         } catch (IOException e) {
```
zhaosongzs, line 61: I am wondering in the previous pushBranch, we didn't manually make the branch protected, but the branch pushed are protected, is this due to settings in GitLab?

bots/tester/src/test/java/org/openjdk/skara/bots/tester/InMemoryHostedRepository.java
```diff
@@ -172,6 +172,14 @@ public List<HostedBranch> branches() {
         return List.of();
     }
 
+    @Override
+    public void protectBranchPattern(String ref) {
+    }
```
zhaosongzs, line 177: The name of this method is protectBranchPattern, but I think it actually protects a specific branch rather than a pattern.

bots/notify/src/main/java/org/openjdk/skara/bots/notify/prbranch/PullRequestBranchNotifier.java
erikj79: Before this change, we relied on a static configuration for the repository with certain branch patterns being protected. When this change goes live, we will need to adjust those patterns so that they do not apply to pr/X branches. Not sure if this answers your question.

bots/tester/src/test/java/org/openjdk/skara/bots/tester/InMemoryHostedRepository.java
erikj79: Hm, on GitLab the underlying call takes a pattern, but now that I think about it, on GitHub it doesn't. There you have to use GraphQL to add pattern based protection. I should change this method to reflect that. We aren't using patterns anyway.

erikj79: I take that back. I see now that I chose not to implement this on GitHub at all since we don't need it. This does take a branch pattern, which are supported by both GitHub and GitLab in the web UI.

zhaosongzs: > I take that back. I see now that I chose not to implement this on GitHub at all since we don't need it. This does take a branch pattern, which are supported by both GitHub and GitLab in the web UI.

Got it! It makes sense

bots/notify/src/main/java/org/openjdk/skara/bots/notify/prbranch/PullRequestBranchNotifier.java
zhaosongzs: > Before this change, we relied on a static configuration for the repository with certain branch patterns being protected. When this change goes live, we will need to adjust those patterns so that they do not apply to pr/X branches. Not sure if this answers your question.

Yes, it answered my question. Thank you!

zhaosongzs (approved):

magicus (approved):
Looks good. Great that you actually managed to find a workaround for this limitation in Gitlab!

PR #1453, opened 2023-01-03
1731: Improve RestRequestCache rate limiter

erikj79:
In the RestRequestCache, we are trying to limit non GET calls for a specific host and user to one call per second (and all calls, including GET, for a specific host and user, need to be serialized). This throttling is based on recommendations from GitHub (https://docs.github.com/en/rest/guides/best-practices-for-integrators?apiVersion=2022-11-28#dealing-with-abuse-rate-limits). The current implementation isn't really doing that however. Calls are serialized, but depending on timing, we can definitely send more than one non GET call per second, at least in bursts.

This patch tries to rework this, using two locks. During high contention, it's possible that this new implementation will be less favorable to non GET calls than the current implementation, but I'm pretty sure we will at least adhere to the recommendation of performing at most one non-GET call per second. See bug for breakdown of locking order.

---------
### Progress

### Issue
 * [SKARA-1731](https://bugs.openjdk.org/browse/SKARA-1731): Improve RestRequestCache rate limiter

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1453/head:pull/1453` \
`$ git checkout pull/1453`

`$ git checkout pull/1453` \
`$ git pull https://git.openjdk.org/skara pull/1453/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1453`

`$ git pr show -t 1453`

</details>

<a href="https://git.openjdk.org/skara/pull/1453.diff">https://git.openjdk.org/skara/pull/1453.diff</a>

</details>

### DIFF
--- a/network/src/main/java/org/openjdk/skara/network/RestRequestCache.java
+++ b/network/src/main/java/org/openjdk/skara/network/RestRequestCache.java
@@ -89,6 +89,7 @@ enum RestRequestCache {
     private final HttpClient client = HttpClient.newBuilder().connectTimeout(Duration.ofSeconds(10)).build();
     private final Logger log = Logger.getLogger("org.openjdk.skara.network");
     private final ConcurrentHashMap<String, Lock> authLocks = new ConcurrentHashMap<>();
+    private final ConcurrentHashMap<String, Lock> authNonGetLocks = new ConcurrentHashMap<>();
     private final ConcurrentHashMap<String, Instant> lastUpdates = new ConcurrentHashMap<>();
 
     private static class LockWithTimeout implements AutoCloseable {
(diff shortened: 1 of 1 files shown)


zhaosongzs (approved):
It really took me a while to understand how it was able to send two non-GET requests in one second, but once I understood it, I checked your revision and I believe that the clever two-lock solution will solve the problem perfectly.

magicus:
Is this not slowing us down unnecessarily? If we have like an incoming stream of requests like this:
GET1 POST1 POST2 GET2 GET3 GET4
Then we will correctly make sure only a single request is sent at a time, and that the two post operations will have a delay between them. But that will also mean that GET2...GET4 will be artificially delayed, awaiting POST1 and POST2.

My understanding is that we should be able to do like:
GET1 POST1 (wait with POST2) GET2 GET3 POST4 (now enough time has passed) GET4

I think this would be easy to accomplish, as well. Just reverse the order on the post locks, so we first take the non-get lock to ensure one second has passed since the last POST operation, and then we take the general lock. That way additional GET requests can be handled while the POST operation is waiting for the full second to pass, but we will still in the end always need to hold the general lock, so we do not send a POST and a GET at the same time.

magicus:
Actually, that is what you are doing. :-)

I got confused by:
```
 var authNonGetLock = authNonGetLocks.computeIfAbsent(authId, id -> new ReentrantLock());
 ```
thinking it was grabbing the lock.

magicus (approved):

PR #1456, opened 2023-01-11
1737: Draft PR for clean backport mistakenly marked as ready

zhaosongzs:
1. readyForIntegration for normal pr is determined by 
(1) whether there is any failed check 
(2) whether there is any additional errors 
(3) whether there is any uncompleted additional progress 
(4) whether there is any integration blocker 

So if there is a draft normal pr, there will always exist a failed check(ReviewersCheck will always fail because the pr is in draft mode and nobody will approve this pr) 
Therefore, a draft normal pr will not be marked as 'ready' because there is a failed check not because this pr is in draft mode. 

2. readyForIntegration for clean backport is determined by 
(1) whether this pr is ready for review(TooFewReviewersIssue won't make it false) 
(2) whether there is any additional errors 
(3) whether there is any uncompleted additional progress 
(4) whether there is any integration blocker 

So if there is a draft clean backport pr, readyForIntegration will be true in most cases. 
Therefore, the clean backport pr will be marked as 'ready' whether it is in draft mode or not.

In this patch, whether there is an rfr label will determine whether readyforintegration is true or false.

---------
### Progress

### Issue
 * [SKARA-1737](https://bugs.openjdk.org/browse/SKARA-1737): Draft PR for clean backport mistakenly marked as ready

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1456/head:pull/1456` \
`$ git checkout pull/1456`

`$ git checkout pull/1456` \
`$ git pull https://git.openjdk.org/skara pull/1456/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1456`

`$ git pr show -t 1456`

</details>

<a href="https://git.openjdk.org/skara/pull/1456.diff">https://git.openjdk.org/skara/pull/1456.diff</a>

</details>

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -383,24 +383,26 @@ class CheckRun {
         }
     }
 
-    private void updateReadyForReview(PullRequestCheckIssueVisitor visitor, List<String> additionalErrors) {
+    private boolean updateReadyForReview(PullRequestCheckIssueVisitor visitor, List<String> additionalErrors) {
         // Additional errors are not allowed
         if (!additionalErrors.isEmpty()) {
             newLabels.remove("rfr");
-            return;
+            return false;
         }
 
         // Draft requests are not for review
         if (pr.isDraft()) {
             newLabels.remove("rfr");
-            return;
+            return false;
         }
 
         // Check if the visitor found any issues that should be resolved before reviewing
         if (visitor.isReadyForReview()) {
             newLabels.add("rfr");
+            return true;
         } else {
             newLabels.remove("rfr");
+            return false;
         }
     }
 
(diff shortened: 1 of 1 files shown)


erikj79 (commented):
While this will function correctly, I'm not sure using the contents of newLabels is the right thing to do here. I think it would be better to be explicit and check `pr.isDraft()` directly. It makes it clearer when reading the code what the requirements are. The rest of the requirements for readyForReview are already repeated in the readyForIntegration checks.

It would also be possible to more clearly define a boolean for `readyForReview` (maybe as a return value from `updateReadyForReview`?) and use that in this check, but then also remove the redundant checks in `readyForIntegration`.

zhaosongzs:
> While this will function correctly, I'm not sure using the contents of newLabels is the right thing to do here. I think it would be better to be explicit and check `pr.isDraft()` directly. It makes it clearer when reading the code what the requirements are. The rest of the requirements for readyForReview are already repeated in the readyForIntegration checks.
> 
> It would also be possible to more clearly define a boolean for `readyForReview` (maybe as a return value from `updateReadyForReview`?) and use that in this check, but then also remove the redundant checks in `readyForIntegration`.

It's a good idea to define a boolean called `readyForReview` and remove the redundant checks. I will do it now. Thanks for the advice!

erikj79 (approved):

PR #1458, opened 2023-01-18
1789: Skara PR commands are not interpreted on Github review comments

zhaosongzs:
A user reported that Skara bots were not responding to commands in review comments or the body of a review. 

This patch addresses that issue by ensuring that the Skara bots can correctly detect and interpret commands.

---------
### Progress

### Issue
 * [SKARA-1789](https://bugs.openjdk.org/browse/SKARA-1789): Skara PR commands are not interpreted on Github review comments

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1458/head:pull/1458` \
`$ git checkout pull/1458`

`$ git checkout pull/1458` \
`$ git pull https://git.openjdk.org/skara pull/1458/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1458`

`$ git pr show -t 1458`

</details>

<a href="https://git.openjdk.org/skara/pull/1458.diff">https://git.openjdk.org/skara/pull/1458.diff</a>

</details>

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -179,7 +179,7 @@ public class PullRequestCommandWorkItem extends PullRequestWorkItem {
     public Collection<WorkItem> prRun(Path scratchPath) {
         log.info("Looking for PR commands");
 
-        var comments = prComments();
+        var comments = getAllComments();
         var nextCommand = nextCommand(pr, comments);
 
         if (nextCommand.isEmpty()) {
@@ -227,5 +227,22 @@ public class PullRequestCommandWorkItem extends PullRequestWorkItem {
     @Override
     public String workItemName() {
         return "pr-command";
+    }
+
+    /**
+     * This method returns all the comments in the pr including comments in reviews(review body) and file specific comments
+     */
+    private List<Comment> getAllComments() {
+        // Regular comments
+        List<Comment> comments = new ArrayList<>(prComments());
+        // Comments in reviews
+        comments.addAll(pr.reviews().stream()
+                .map(review -> new Comment("Review" + review.id(), review.body().orElse(""), review.reviewer(), review.createdAt(), null))
+                .toList());
+        // File specific comments
+        comments.addAll(pr.reviewComments().stream()
+                .map(reviewComment -> new Comment("ReviewComment" + reviewComment.id(), reviewComment.body(), reviewComment.author(), reviewComment.createdAt(), reviewComment.updatedAt()))
+                .toList());
+        return comments;
     }
 }
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
@@ -240,9 +240,10 @@ public class PullRequestCommandWorkItem extends PullRequestWorkItem {
                 .map(review -> new Comment("Review" + review.id(), review.body().orElse(""), review.reviewer(), review.createdAt(), null))
                 .toList());
         // File specific comments
-        comments.addAll(pr.reviewComments().stream()
-                .map(reviewComment -> new Comment("ReviewComment" + reviewComment.id(), reviewComment.body(), reviewComment.author(), reviewComment.createdAt(), reviewComment.updatedAt()))
-                .toList());
+//        comments.addAll(pr.reviewComments().stream()
+//                .map(reviewComment -> new Comment("ReviewComment" + reviewComment.id(), reviewComment.body(), reviewComment.author(), reviewComment.createdAt(), reviewComment.updatedAt()))
+//                .toList());
+        comments = comments.stream().sorted(Comparator.comparing(Comment::createdAt)).toList();
         return comments;
     }
 }
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
@@ -233,13 +233,8 @@ public class PullRequestCommandWorkItem extends PullRequestWorkItem {
      * This method returns all the comments in the pr including comments in reviews(review body)
      */
     private List<Comment> getAllComments() {
-        // Regular comments
-        List<Comment> comments = new ArrayList<>(prComments());
-        // Comments in reviews
-        comments.addAll(pr.reviews().stream()
-                .map(review -> new Comment("Review" + review.id(), review.body().orElse(""), review.reviewer(), review.createdAt(), null))
-                .toList());
-        comments = comments.stream().sorted(Comparator.comparing(Comment::createdAt)).toList();
-        return comments;
+        return Stream.concat(prComments().stream(),
+                        pr.reviews().stream().map(review -> new Comment("Review" + review.id(), review.body().orElse(""), review.reviewer(), review.createdAt(), null)))
+                .sorted(Comparator.comparing(Comment::createdAt)).toList();
     }
 }
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
@@ -230,7 +230,7 @@ public class PullRequestCommandWorkItem extends PullRequestWorkItem {
     }
 
     /**
-     * This method returns all the comments in the pr including comments in reviews(review body) and file specific comments
+     * This method returns all the comments in the pr including comments in reviews(review body)
      */
     private List<Comment> getAllComments() {
         // Regular comments
@@ -239,10 +239,6 @@ public class PullRequestCommandWorkItem extends PullRequestWorkItem {
         comments.addAll(pr.reviews().stream()
                 .map(review -> new Comment("Review" + review.id(), review.body().orElse(""), review.reviewer(), review.createdAt(), null))
                 .toList());
-        // File specific comments
-//        comments.addAll(pr.reviewComments().stream()
-//                .map(reviewComment -> new Comment("ReviewComment" + reviewComment.id(), reviewComment.body(), reviewComment.author(), reviewComment.createdAt(), reviewComment.updatedAt()))
-//                .toList());
         comments = comments.stream().sorted(Comparator.comparing(Comment::createdAt)).toList();
         return comments;
     }
--- a/bots/pr/src/test/java/org/openjdk/skara/bots/pr/PullRequestCommandTests.java
+++ b/bots/pr/src/test/java/org/openjdk/skara/bots/pr/PullRequestCommandTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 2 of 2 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
```diff
@@ -228,4 +228,21 @@ public String toString() {
...
+        // Regular comments
+        List<Comment> comments = new ArrayList<>(prComments());
+        // Comments in reviews
+        comments.addAll(pr.reviews().stream()
+                .map(review -> new Comment("Review" + review.id(), review.body().orElse(""), review.reviewer(), review.createdAt(), null))
+                .toList());
+        // File specific comments
+        comments.addAll(pr.reviewComments().stream()
+                .map(reviewComment -> new Comment("ReviewComment" + reviewComment.id(), reviewComment.body(), reviewComment.author(), reviewComment.createdAt(), reviewComment.updatedAt()))
+                .toList());
```
zhaosongzs, line 245: I am not sure if the "review", "review comment", and "comment" in Github refer to the same numbering system, but based on my experiments, it is likely that they have their own independent numbering system. Although the probability of ID conflicts is small, I think it would be better to add "review" before the ID of a review, and "ReviewComment" before the ID of a review comment.

erikj79 (commented):
I will need to think more on what this means and how it would work. The bug only asked for adding commands in `Review`, not `ReviewComment`. I don't think it makes sense to run commands in a `ReviewComment`. (The latter is a comment on a specific file or line inside a review)

If we do this, then commands must still be processed in order, so the list needs to be sorted on create time.

We also need to consider how this applies to Gitlab, where reviews are different.

erikj79:
> We also need to consider how this applies to Gitlab, where reviews are different.

In GitLab Reviews don't have a body, so this wouldn't have any effect there at least.

zhaosongzs:
> I will need to think more on what this means and how it would work. The bug only asked for adding commands in `Review`, not `ReviewComment`. I don't think it makes sense to run commands in a `ReviewComment`. (The latter is a comment on a specific file or line inside a review)
> 
> If we do this, then commands must still be processed in order, so the list needs to be sorted on create time.
> 
> We also need to consider how this applies to Gitlab, where reviews are different.

In the description of the bug, Magnus said "I guess the same goes if you would make a single comment in review mode as well, but I have not tested this. ". I thought he would expect the command in single comment(which is reviewComment) could be interpreted too. I don't know whether we need it or not.

It's a good point that we should sort the comments by created time. I will fix it now.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
erikj79: Have you found a conflict? I did a quick check and didn't find any, but that's definitely no guarantee.

magicus:
I am a bit uncertain about PR commands in review comments. I don't think there is much of a use case for that, and if it is costly for us to process, I think we can skip it. But otoh, if the overhead is not an issue, I think having a consistent UX is a good thing. If someone were to be adding a PR command in a comment, they will be surprised if it doesn't work. 

But the main part is the review itself, yes. That definitely feels like a "normal" comment.

erikj79:
> In the description of the bug, Magnus said "I guess the same goes if you would make a single comment in review mode as well, but I have not tested this. ". I thought he would expect the command in single comment(which is reviewComment) could be interpreted too. I don't know whether we need it or not.

You are right, I misinterpreted that part. 

> I am a bit uncertain about PR commands in review comments. I don't think there is much of a use case for that, and if it is costly for us to process, I think we can skip it. But otoh, if the overhead is not an issue, I think having a consistent UX is a good thing. If someone were to be adding a PR command in a comment, they will be surprised if it doesn't work.

My concern is that review threads can be marked as resolved and no longer visible. They can also be hard to find if the author force pushes to the source branch. It makes it hard to backtrack through the comment history to understand when a command was issued.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
erikj79: I thought some more on your proposed solution here and I think it's sound.

zhaosongzs:
> > In the description of the bug, Magnus said "I guess the same goes if you would make a single comment in review mode as well, but I have not tested this. ". I thought he would expect the command in single comment(which is reviewComment) could be interpreted too. I don't know whether we need it or not.
> 
> You are right, I misinterpreted that part.
> 
> > I am a bit uncertain about PR commands in review comments. I don't think there is much of a use case for that, and if it is costly for us to process, I think we can skip it. But otoh, if the overhead is not an issue, I think having a consistent UX is a good thing. If someone were to be adding a PR command in a comment, they will be surprised if it doesn't work.
> 
> My concern is that review threads can be marked as resolved and no longer visible. They can also be hard to find if the author force pushes to the source branch. It makes it hard to backtrack through the comment history to understand when a command was issued.

Due to Erik's concern, I think it would be better to temporarily disregard commands within ReviewComments. If a use case for this feature is identified in the future, we can revisit the topic.

dholmes-ora:
I also have misgivings about scanning review comments for PR commands - so it seems there is consensus there. 

Even in the main review comment what gets scanned - all the text? Only the start of lines? I would worry about text being misconstrued as a command. This would especially be true for skara PRs where review comments may well refer to PR commands as part of the discussion.

kevinrushforth:
> I also have misgivings about scanning review comments for PR commands - so it seems there is consensus there.

I definitely agree with David and others on this.

> Even in the main review comment what gets scanned - all the text? Only the start of lines? I would worry about text being misconstrued as a command. This would especially be true for skara PRs where review comments may well refer to PR commands as part of the discussion.

If it works the same as PR commands in ordinary comments, they are only parsed at the beginning of a line after trimming whitespace. @erikj79 or @zhaosongzs can confirm.

magicus:
It sounds reasonable to not allow PR comments in review comments on individual lines, yes. As Erik says, they are presented differently, can be hidden, and form a "different kind" of communication.

I still think the main review "comment" (or whatever to call it) deserves to be parsed for PR comments. Once it is posted, it looks just like a regular comment. If it were not for Skara treating this differently, I would have had no idea it was even treated differently internally in Github. To me, it is like a "approve this PR + also add this comment", not "approve this PR with this special kind of comment that can only be generated here but looks like any other normal comment".

David:
> Even in the main review comment what gets scanned - all the text? Only the start of lines? I would worry about text being misconstrued as a command. This would especially be true for skara PRs where review comments may well refer to PR commands as part of the discussion.

I am not sure what you mean by that? My example is the kind of comment I've been doing far too many times before remembering that it will not work, I look at a change, check the build changes, presses the green button, selects "Approve" and writes:
```
This looks good from a build perspective, but you'll need someone from core-libs to check it as well.

/reviewers 2
```

I have no idea how a text like that could be "misconstrued" as a PR command. As always, PR commands must be on a single line and start with a `/`. The risk that anyone does this by mistake seems to me to be just as high for the "main review comment" (or whatever it is called) than as in a normal comment to the PR.

If you were to discuss PR commands in a Skara PR, I would assume that you would backtick them like so `/reviewers 2` when you are mentioning them.

magicus:
*LOL* 

Thank you @openjdk bot for providing a quick example that the risk for "misconstruing" commands are just as present in normal commands, and not something that will be a special problems for main review comments. :-D

/reviewers 1

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
```diff
@@ -228,4 +228,22 @@ public String toString() {
     public String workItemName() {
         return "pr-command";
     }
+
+    /**
+     * This method returns all the comments in the pr including comments in reviews(review body) and file specific comments
```
kevinrushforth, line 233: Given the discussion around only using "ordinary" PR comments, and not inline comments, I think the "and file specific comments" part is now wrong.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
```diff
@@ -228,4 +228,22 @@ public String toString() {
...
+        // Regular comments
+        List<Comment> comments = new ArrayList<>(prComments());
+        // Comments in reviews
+        comments.addAll(pr.reviews().stream()
+                .map(review -> new Comment("Review" + review.id(), review.body().orElse(""), review.reviewer(), review.createdAt(), null))
+                .toList());
+        // File specific comments
+//        comments.addAll(pr.reviewComments().stream()
+//                .map(reviewComment -> new Comment("ReviewComment" + reviewComment.id(), reviewComment.body(), reviewComment.author(), reviewComment.createdAt(), reviewComment.updatedAt()))
+//                .toList());
```
kevinrushforth, line 245: Since we have decided not to do this, I recommend to remove this commented out code.

bots/pr/src/test/java/org/openjdk/skara/bots/pr/PullRequestCommandTests.java
```diff
@@ -357,4 +359,40 @@ void summaryCommandInBodyWithBotComment(TestInfo testInfo) throws IOException {
...
+            pr.addReview(Review.Verdict.APPROVED, "/reviewers 3");
+            // Run the bot
+            TestBotRunner.runPeriodicItems(prBot);
+            assertLastCommentContains(pr, "The total number of required reviews for this PR (including the jcheck configuration and the last /reviewers command) is now set to 3");
+
+            //Test command in reviewComment
+//            pr.addReviewComment(masterHash, editHash, Path.of("appendable.txt").toString(), 2, "/reviewers 4");
+//            // Run the bot
+//            TestBotRunner.runPeriodicItems(prBot);
+//            assertLastCommentContains(pr, "The total number of required reviews for this PR (including the jcheck configuration and the last /reviewers command) is now set to 4");
```
kevinrushforth, line 395: Ditto the recommendation to remove this.

kevinrushforth (commented):
Since there is general agreement not to attempt to use inline "file" comments, I left a couple suggestions to remove some commented out code.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
zhaosongzs: > Given the discussion around only using "ordinary" PR comments, and not inline comments, I think the "and file specific comments" part is now wrong.

Good catch! Thanks Kevin. Will fix it.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
zhaosongzs: Sure

bots/pr/src/test/java/org/openjdk/skara/bots/pr/PullRequestCommandTests.java
zhaosongzs: Sure

zhaosongzs:
> > I also have misgivings about scanning review comments for PR commands - so it seems there is consensus there.
> 
> I definitely agree with David and others on this.
> 
> > Even in the main review comment what gets scanned - all the text? Only the start of lines? I would worry about text being misconstrued as a command. This would especially be true for skara PRs where review comments may well refer to PR commands as part of the discussion.
> 
> If it works the same as PR commands in ordinary comments, they are only parsed at the beginning of a line after trimming whitespace. @erikj79 or @zhaosongzs can confirm.

Yes, you are right. It just works the same as PR commands in regular comments.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
```diff
@@ -228,4 +228,18 @@ public String toString() {
     public String workItemName() {
         return "pr-command";
     }
+
+    /**
+     * This method returns all the comments in the pr including comments in reviews(review body)
+     */
+    private List<Comment> getAllComments() {
+        // Regular comments
+        List<Comment> comments = new ArrayList<>(prComments());
```
erikj79, line 237: It's not really important, but you could use Streams.concat() instead of creating two intermediate lists.

erikj79 (approved):

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
zhaosongzs: It would be better

erikj79 (approved):

dholmes-ora:
@magicus  you demonstrated my concern about misconstrued comments beautifully - thank you :)

If they are only processed as commands at the start of the line that greatly reduces the risk and concern, and for the rest people just have to be careful. :)

PR #1459, opened 2023-01-19
1725: Clarify "no known role" error message

zhaosongzs:
A user reported that when people review PRs, the bot adds a list of reviewers into description. If a user is not registered in OpenJDK Census, a note is added: 

(no known github.com user name / role) 

What it actually means is "no known OpenJDK user name / role".

Github.com is being used here because we are utilizing the name of the namespace. However, the purpose of the namespace is to map users from this namespace to either openjdk or internal. Actually, we couldn't find the user in openjdk or internal, not in the namespace. The census repository name can indicate whether it is openjdk or internal.

---------
### Progress

### Issue
 * [SKARA-1725](https://bugs.openjdk.org/browse/SKARA-1725): Clarify "no known role" error message

### Reviewers
 * [Guoxiong Li](https://openjdk.org/census#gli) (@lgxbslgx - Committer) ⚠️ Review applies to [d6479ee3](https://git.openjdk.org/skara/pull/1459/files/d6479ee38263c5fa46f41c23b64bf3aecea851eb)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1459/head:pull/1459` \
`$ git checkout pull/1459`

`$ git checkout pull/1459` \
`$ git pull https://git.openjdk.org/skara pull/1459/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1459`

`$ git pr show -t 1459`

</details>

<a href="https://git.openjdk.org/skara/pull/1459.diff">https://git.openjdk.org/skara/pull/1459.diff</a>

</details>

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -491,9 +491,9 @@ class CheckRun {
             ret.append("@");
             ret.append(user.username());
             ret.append(" (no known ");
-            var censusRepoName = workItem.bot.censusRepo().name();
-            var index = censusRepoName.lastIndexOf("-");
-            ret.append(censusRepoName.substring(index + 1));
+            var censusDomain = censusInstance.configuration().census().domain();
+            var index = censusDomain.lastIndexOf('.');
+            ret.append(censusDomain, 0, index);
             ret.append(" user name / role)");
         } else {
             // The HostUser is null and the Contributor is not null
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -491,9 +491,7 @@ class CheckRun {
             ret.append("@");
             ret.append(user.username());
             ret.append(" (no known ");
-            var censusDomain = censusInstance.configuration().census().domain();
-            var index = censusDomain.lastIndexOf('.');
-            ret.append(censusDomain, 0, index);
+            ret.append(censusInstance.configuration().census().domain());
             ret.append(" user name / role)");
         } else {
             // The HostUser is null and the Contributor is not null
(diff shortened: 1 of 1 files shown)


magicus:
And the OpenJDK census name ends with `OpenJDK`, so the effect is to output `no known OpenJDK user name / role`?

erikj79:
I'm not sure using the name of the internal data repository as the source for this error message is the right thing. Maybe we need to add some new field to the census data that we can extract.

erikj79:
The jcheck configuration has a field `domain` in the `[census]` section, perhaps that would be a better description of the space we are looking for the user in? In the OpenJDK projects, this field has the value `openjdk.org` (internally we use oracle.com).

magicus:
I think "has no OpenJDK user name" is better than "has no openjdk.org user name", so we would need to do something about that if we were to use that field. A "openjdk.org user name" sounds like something you could use on the home page, which is not a thing. The accounts we are talking about are commonly referred to as "OpenJDK user name" and I think we should stay with that terminology.

zhaosongzs:
> And the OpenJDK census name ends with `-OpenJDK`, so the effect is to output `no known OpenJDK user name / role`?

Yes, exactly.

zhaosongzs:
> The jcheck configuration has a field `domain` in the `[census]` section, perhaps that would be a better description of the space we are looking for the user in? In the OpenJDK projects, this field has the value `openjdk.org` (internally we use oracle.com).

Yes, it's a good idea to  use this field `domain`. Thx!

lgxbslgx (approved):
Now it becomes `no known openjdk user` and `no known oracle user`. Looks good if we can't find a good way to construct `OpenJDK` and `Oracle`.

erikj79:
Until we have a better field for this stored somewhere in the census data, I would prefer the domain field raw without modification.

zhaosongzs:
> Until we have a better field for this stored somewhere in the census data, I would prefer the domain field raw without modification.

It will be `no known openjdk.org user` and 'no known oracle.com user'.  I think it's also ok.

erikj79 (approved):

PR #1463, opened 2023-01-25
1806: Optimize the CSR progress message and log when CSR not found

zhaosongzs:
A user reported that she have a PR which can't find its corresponding CSR. 
Erik found that the issue was caused by the user not updating their branch in a long time, and the fixVersion in the .jcheck/conf file was still set at 20. However, the fixVersion of the CSR was 21, make the bot unable to locate the corresponding CSR. 

In this patch, I optimized the CSR progress message and log.

---------
### Progress

### Issue
 * [SKARA-1806](https://bugs.openjdk.org/browse/SKARA-1806): Optimize the CSR progress message and log when CSR not found

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1463/head:pull/1463` \
`$ git checkout pull/1463`

`$ git checkout pull/1463` \
`$ git pull https://git.openjdk.org/skara pull/1463/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1463`

`$ git pr show -t 1463`

</details>

<a href="https://git.openjdk.org/skara/pull/1463.diff">https://git.openjdk.org/skara/pull/1463.diff</a>

</details>

### DIFF
--- a/bots/common/src/main/java/org/openjdk/skara/bots/common/BotUtils.java
+++ b/bots/common/src/main/java/org/openjdk/skara/bots/common/BotUtils.java
@@ -43,7 +43,7 @@ public class BotUtils {
     public static Optional<JdkVersion> getVersion(PullRequest pr) {
         var confFile = pr.repository().fileContents(".jcheck/conf", pr.headHash().hex())
                 .orElse(pr.repository().fileContents(".jcheck/conf", pr.targetRef()).orElse(null));
-        if(confFile == null){
+        if (confFile == null) {
             return Optional.empty();
         }
         var configuration = JCheckConfiguration.parse(confFile.lines().toList());
--- a/bots/common/src/main/java/org/openjdk/skara/bots/common/BotUtils.java
+++ b/bots/common/src/main/java/org/openjdk/skara/bots/common/BotUtils.java
@@ -42,8 +42,10 @@ public class BotUtils {
      */
     public static Optional<JdkVersion> getVersion(PullRequest pr) {
         var confFile = pr.repository().fileContents(".jcheck/conf", pr.headHash().hex())
-                .orElse(pr.repository().fileContents(".jcheck/conf", pr.targetRef())
-                        .orElseThrow(() -> new RuntimeException("Could not find .jcheck/conf in neither src or target of PR " + pr)));
+                .orElse(pr.repository().fileContents(".jcheck/conf", pr.targetRef()).orElse(null));
+        if(confFile == null){
+            return Optional.empty();
+        }
         var configuration = JCheckConfiguration.parse(confFile.lines().toList());
         var version = configuration.general().version().orElse(null);
         if (version == null || "".equals(version)) {
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -30,6 +30,7 @@ import org.openjdk.skara.forge.*;
 import org.openjdk.skara.host.HostUser;
 import org.openjdk.skara.issuetracker.*;
 import org.openjdk.skara.jbs.Backports;
+import org.openjdk.skara.jbs.JdkVersion;
 import org.openjdk.skara.vcs.*;
 import org.openjdk.skara.vcs.openjdk.Issue;
 
@@ -72,6 +73,8 @@ class CheckRun {
     private final boolean reviewCleanBackport;
 
     private Duration expiresIn;
+
+    private Optional<JdkVersion> versionOpt;
 
     private CheckRun(CheckWorkItem workItem, PullRequest pr, Repository localRepo, List<Comment> comments,
                      List<Review> allReviews, List<Review> activeReviews, Set<String> labels,
@@ -142,7 +145,7 @@ class CheckRun {
         if (issueProject == null) {
             return List.of();
         }
-        var versionOpt = BotUtils.getVersion(pr);
+        versionOpt = BotUtils.getVersion(pr);
         if (versionOpt.isEmpty()) {
             return List.of();
         }
@@ -271,7 +274,8 @@ class CheckRun {
                 .filter(issue -> !isWithdrawnCSR(issue))
                 .toList();
         if (csrIssues.isEmpty() && pr.labelNames().contains("csr")) {
-            ret.put("Change requires a CSR request (needs to be created) to be approved", false);
+            ret.put("Change requires a CSR request matching fixVersion " + (versionOpt.isPresent() ? versionOpt.get().raw() : "(No fixVersion in .jcheck/conf)")
+                    + " to be approved (needs to be created)", false);
         }
         for (var csrIssue : csrIssues) {
             if (!csrIssue.isClosed()) {
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -1124,7 +1124,11 @@ class CheckRun {
             }
             PullRequestCheckIssueVisitor visitor = checkablePullRequest.createVisitor();
 
-            var version = BotUtils.getVersion(pr).orElse(null);
+            JdkVersion version = null;
+            try {
+                version = BotUtils.getVersion(pr).orElse(null);
+            } catch (Exception e) {
+            }
             // issues without CSR issues and JEP issues
             var issues = issues();
             var csrIssueTrackerIssues = getCsrIssueTrackerIssues(issues, version);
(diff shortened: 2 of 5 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -73,6 +74,8 @@ class CheckRun {
 
     private Duration expiresIn;
 
+    private Optional<JdkVersion> versionOpt;
```
erikj79, line 77: Fields should not be `Optional`. We should also not add fields that are arbitrarily initiated as a side effect of some method. This needs to be initiated in one place and then propagated through parameters. Looking at the calling tree of the various getIssue methods, this looks like a pretty bad mess. Not sure how to untangle this to propagate the version value.

zhaosongzs: I add this field because I don't want to call `BotUtils.getVersion(pr)` twice. If adding a field is a bad idea, I think just call `BotUtils.getVersion(pr)` twice is better than add `JdkVersion` to many method interfaces.

zhaosongzs: I think the calling tree of `issues()` is in a mess because we have two different type of issues(`org.openjdk.skara.vcs.openjdk` and `org.openjdk.skara.issuetracker.Issue`). Besides,  we don't have class fields to store issues, csrIssues, JepIssues. So we need to call `issues()` three times.

zhaosongzs: > Yes, the two different Issue types definitely adds to the confusion, but we also have the `::issues` method that takes two booleans. It has three callers and only one of those sets the parameters true. This should just have been split into different methods. As it is now, they are calling each other in a circle, only saved by the boolean being false in the second call. Untangling this would make it pretty straightforward to provide a JdkVersion argument to the method for getting just CSR issues.
> 
> I think there is value in only calling `BotUtils.getVersion(pr)` once as it's quite expensive to evaluate.

Good idea, I will split them into different methods. Thx!

erikj79: I started looking at this call chain some more. We are already fetching issues multiple times, that's really what's causing this problem in the first place. We should only fetch issues once in a CheckRun. It will take some refactoring, since we need to preserve the IssueTracker Issues for CSR issues, but still worth it I think.

zhaosongzs: > I started looking at this call chain some more. We are already fetching issues multiple times, that's really what's causing this problem in the first place. We should only fetch issues once in a CheckRun. It will take some refactoring, since we need to preserve the IssueTracker Issues for CSR issues, but still worth it I think.
> 
> So instead of sending the fixVersion around, we should be fetching issues at a higher level and send those around.

I think so too. So I already did some refactoring, now we will only fetch issues once and fetch csr issues once too.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -161,9 +164,9 @@ private List<org.openjdk.skara.issuetracker.Issue> getCsrIssueTrackerIssues(List
     /**
      * Get the csr issue. Note: this `Issue` is not the issue in module `issuetracker`.
      */
-    private List<Issue> getCsrIssues(List<Issue> issues) {
+    private List<Issue> getCsrIssues(List<org.openjdk.skara.issuetracker.Issue> CsrIssueTrackerIssues) {
```
erikj79, line 167:
```suggestion
    private List<Issue> getCsrIssues(List<org.openjdk.skara.issuetracker.Issue> csrIssueTrackerIssues) {
```

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -161,9 +164,9 @@ private List<org.openjdk.skara.issuetracker.Issue> getCsrIssueTrackerIssues(List
...
 
-        return getCsrIssueTrackerIssues(issues).stream()
+        return CsrIssueTrackerIssues.stream()
```
erikj79, line 169:
```suggestion
        return csrIssueTrackerIssues.stream()
```

erikj79 (commented):
This is definitely better. I still feel like the whole class could use some kind of overhaul, but no need to dig further now.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -1124,7 +1124,11 @@ private void checkStatus() {
             }
             PullRequestCheckIssueVisitor visitor = checkablePullRequest.createVisitor();
 
-            var version = BotUtils.getVersion(pr).orElse(null);
+            JdkVersion version = null;
+            try {
+                version = BotUtils.getVersion(pr).orElse(null);
+            } catch (Exception e) {
+            }
```
erikj79, line 1131: I'm not sure ignoring an exception here is a good idea. In what situation did it throw one?

zhaosongzs: > I'm not sure ignoring an exception here is a good idea. In what situation did it throw one?

It's because `BotUtils.getVersion` only works for fixVersion is configured in` .jcheck/conf`. If we use the override jcheck conf, it will throw exception. So it will make some override jcheck conf tests fail. I think we could fix it, but not in this pr.

erikj79: It needs to be solved in some way. Catching `Exception` and ignoring it is not allowed in bot code. That could be triggered by an IOException, or failure in contacting the remote or almost anything else. If that happens, we need to handle the failure correct so that the WorkItem will be retried.

BotUtils.getVersion either can't throw RuntimeException when it can't find a .jcheck/conf file, or we need to handle the fallback to the override repo. The latter would be cleaner, but we probably don't really need that support. I think returning Optional.empty() from BotUtils.getVersion() is probably ok if no .jcheck/conf file could be found.

zhaosongzs: > It needs to be solved in some way. Catching `Exception` and ignoring it is not allowed in bot code. That could be triggered by an IOException, or failure in contacting the remote or almost anything else. If that happens, we need to handle the failure correct so that the WorkItem will be retried.
> 
> BotUtils.getVersion either can't throw RuntimeException when it can't find a .jcheck/conf file, or we need to handle the fallback to the override repo. The latter would be cleaner, but we probably don't really need that support. I think returning Optional.empty() from BotUtils.getVersion() is probably ok if no .jcheck/conf file could be found.

Got it! I will fix it

zhaosongzs: Now I realized that this jcheck conf issue happens because the place of `JdkVersion version = BotUtils.getVersion(pr).orElse(null);`. In `CheckWorkItem`, we have check for jcheck configuration in **target branch**. So if jcheck conf missing or invalid in **target branch**, the bot will return immediately. 

However, in `BotUtils#getVersion`,  the bot would try to get conf from **source branch** first. Although we also have check for **source branch**, it is after the call of JdkVersion version = BotUtils.getVersion(pr).orElse(null);`. So bad conf in **source branch** would cause getVersion throw exceptions.

bots/common/src/main/java/org/openjdk/skara/bots/common/BotUtils.java
```diff
@@ -42,8 +42,10 @@ public class BotUtils {
      */
     public static Optional<JdkVersion> getVersion(PullRequest pr) {
         var confFile = pr.repository().fileContents(".jcheck/conf", pr.headHash().hex())
-                .orElse(pr.repository().fileContents(".jcheck/conf", pr.targetRef())
-                        .orElseThrow(() -> new RuntimeException("Could not find .jcheck/conf in neither src or target of PR " + pr)));
+                .orElse(pr.repository().fileContents(".jcheck/conf", pr.targetRef()).orElse(null));
+        if(confFile == null){
+            return Optional.empty();
+        }
```
erikj79, line 48:
```suggestion
        if (confFile == null) {
            return Optional.empty();
        }
```

erikj79: If you fix this formatting, then this is as good as it gets now. I'm not happy with the state of CheckRun, but at least this version should work as we want it without doing excessive extra work.

erikj79 (approved):

PR #1467, opened 2023-02-08
1813: URLs printed by Skara should have ".git" at the end of the repo name

zhaosongzs:
In this patch, `.git` is added to the URLs printed by Skara.

---------
### Progress

### Issue
 * [SKARA-1813](https://bugs.openjdk.org/browse/SKARA-1813): URLs printed by Skara should have ".git" at the end of the repo name

### Reviewers
 * [Erik Joelsson](https://openjdk.org/census#erikj) (@erikj79 - **Reviewer**) ⚠️ Review applies to [95d2c6e2](https://git.openjdk.org/skara/pull/1467/files/95d2c6e27c1f2f3c8acd7d6f3a8077b2fcda9b5b)
 * [Guoxiong Li](https://openjdk.org/census#gli) (@lgxbslgx - Committer) ⚠️ Review applies to [959eb93b](https://git.openjdk.org/skara/pull/1467/files/959eb93b4f0fdf831669d81c215ce03df0658947)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1467/head:pull/1467` \
`$ git checkout pull/1467`

`$ git checkout pull/1467` \
`$ git pull https://git.openjdk.org/skara pull/1467/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1467`

`$ git pr show -t 1467`

</details>

<a href="https://git.openjdk.org/skara/pull/1467.diff">https://git.openjdk.org/skara/pull/1467.diff</a>

</details>

### DIFF
--- a/bots/checkout/src/main/java/org/openjdk/skara/bots/checkout/CheckoutBot.java
+++ b/bots/checkout/src/main/java/org/openjdk/skara/bots/checkout/CheckoutBot.java
@@ -64,9 +64,7 @@ public class CheckoutBot implements Bot, WorkItem {
     }
 
     private URI webURI() {
-        var webURI = from.remoteUrl().toString();
-
-        return URI.create(webURI);
+        return from.remoteUrl();
     }
 
     private URI uri() {
--- a/bots/checkout/src/main/java/org/openjdk/skara/bots/checkout/CheckoutBot.java
+++ b/bots/checkout/src/main/java/org/openjdk/skara/bots/checkout/CheckoutBot.java
@@ -63,10 +63,6 @@ public class CheckoutBot implements Bot, WorkItem {
         return URLEncoder.encode(uri.toString(), StandardCharsets.UTF_8);
     }
 
-    private URI webURI() {
-        return from.remoteUrl();
-    }
-
     private URI uri() {
         var uri = from.url().toString();
         if (!uri.endsWith(".git")) {
@@ -98,7 +94,7 @@ public class CheckoutBot implements Bot, WorkItem {
     @Override
     public Collection<WorkItem> run(Path scratch) {
         try {
-            var fromDir = storage.resolve(urlEncode(webURI()));
+            var fromDir = storage.resolve(urlEncode(from.remoteUrl()));
             Repository fromRepo = null;
             if (!Files.exists(fromDir)) {
                 Files.createDirectories(fromDir);
--- a/bots/checkout/src/main/java/org/openjdk/skara/bots/checkout/CheckoutBot.java
+++ b/bots/checkout/src/main/java/org/openjdk/skara/bots/checkout/CheckoutBot.java
@@ -64,10 +64,7 @@ public class CheckoutBot implements Bot, WorkItem {
     }
 
     private URI webURI() {
-        var webURI = from.webUrl().toString();
-        if (!webURI.endsWith(".git")) {
-            webURI += ".git";
-        }
+        var webURI = from.remoteUrl().toString();
 
         return URI.create(webURI);
     }
--- a/bots/checkout/src/main/java/org/openjdk/skara/bots/checkout/CheckoutBot.java
+++ b/bots/checkout/src/main/java/org/openjdk/skara/bots/checkout/CheckoutBot.java
@@ -32,7 +32,6 @@ import org.openjdk.skara.storage.StorageBuilder;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.util.*;
-import java.util.stream.Collectors;
 import java.nio.file.*;
 import java.nio.charset.StandardCharsets;
 import java.net.URI;
@@ -64,7 +63,7 @@ public class CheckoutBot implements Bot, WorkItem {
     }
 
     private URI uri() {
-        var uri = from.url().toString();
+        var uri = from.authenticatedUrl().toString();
         if (!uri.endsWith(".git")) {
             uri += ".git";
         }
@@ -94,7 +93,7 @@ public class CheckoutBot implements Bot, WorkItem {
     @Override
     public Collection<WorkItem> run(Path scratch) {
         try {
-            var fromDir = storage.resolve(urlEncode(from.remoteUrl()));
+            var fromDir = storage.resolve(urlEncode(from.url()));
             Repository fromRepo = null;
             if (!Files.exists(fromDir)) {
                 Files.createDirectories(fromDir);
--- a/bots/checkout/src/main/java/org/openjdk/skara/bots/checkout/CheckoutBot.java
+++ b/bots/checkout/src/main/java/org/openjdk/skara/bots/checkout/CheckoutBot.java
@@ -62,15 +62,6 @@ public class CheckoutBot implements Bot, WorkItem {
         return URLEncoder.encode(uri.toString(), StandardCharsets.UTF_8);
     }
 
-    private URI uri() {
-        var uri = from.authenticatedUrl().toString();
-        if (!uri.endsWith(".git")) {
-            uri += ".git";
-        }
-
-        return URI.create(uri);
-    }
-
     @Override
     public boolean concurrentWith(WorkItem other) {
         if (!(other instanceof CheckoutBot)) {
@@ -98,7 +89,7 @@ public class CheckoutBot implements Bot, WorkItem {
             if (!Files.exists(fromDir)) {
                 Files.createDirectories(fromDir);
                 log.info("Cloning Git repo " + from + " to " + fromDir);
-                fromRepo = Repository.clone(uri(), fromDir);
+                fromRepo = Repository.clone(from.authenticatedUrl(), fromDir);
             } else {
                 log.info("Getting existing Git repo repository from " + fromDir);
                 fromRepo = Repository.get(fromDir).orElseThrow(() ->
--- a/bots/hgbridge/src/main/java/org/openjdk/skara/bots/hgbridge/JBridgeBot.java
+++ b/bots/hgbridge/src/main/java/org/openjdk/skara/bots/hgbridge/JBridgeBot.java
@@ -64,7 +64,7 @@ public class JBridgeBot implements Bot, WorkItem {
     }
 
     private void pushMarks(Path markSource, String destName, Path markScratchPath) throws IOException {
-        var marksRepo = Repository.materialize(markScratchPath, exporterConfig.marksRepo().url(),
+        var marksRepo = Repository.materialize(markScratchPath, exporterConfig.marksRepo().authenticatedUrl(),
                                                "+" + exporterConfig.marksRef() + ":hgbridge_marks");
 
         // We should never change existing marks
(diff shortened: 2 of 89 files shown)


erikj79 (changes requested):
I don't think we should just add ".git" to the webUrl. What we need is a new method on `HostedRepository` that returns the URL in a format suitable for the command line client. Basically the form you get in the "Code" drop down on the repository main page on GitHub. There is already a method `HostedRepository::url` which is almost this but not quite. It doesn't perform the transformation, and it adds authentication, which we don't want. Maybe `HostedRepository::transformedUrl`, not sure.

zhaosongzs:
> I don't think we should just add ".git" to the webUrl. What we need is a new method on `HostedRepository` that returns the URL in a format suitable for the command line client. Basically the form you get in the "Code" drop down on the repository main page on GitHub. There is already a method `HostedRepository::url` which is almost this but not quite. It doesn't perform the transformation, and it adds authentication, which we don't want. Maybe `HostedRepository::transformedUrl`, not sure.

Ok, I will add a new method. Thanks.

forge/src/main/java/org/openjdk/skara/forge/HostedRepository.java
```diff
@@ -83,6 +83,7 @@ PullRequest createPullRequest(HostedRepository target,
     URI webUrl(String baseRef, String headRef);
     URI diffUrl(String prId);
     VCS repositoryType();
+    URI remoteUrl();
```
erikj79, line 86:
```suggestion
    /**
     * Returns a URL suitable for CLI interactions with the repository
     */
    URI remoteUrl();
```

erikj79 (commented):
This is better. There are more places that need to use the new method. The fallback instructions for /backport is one example where I just saw URLs with no .git in example command lines.

zhaosongzs:
> This is better. There are more places that need to use the new method. The fallback instructions for /backport is one example where I just saw URLs with no .git in example command lines.

I would use it in suitable places.

bots/checkout/src/main/java/org/openjdk/skara/bots/checkout/CheckoutBot.java
```diff
@@ -64,10 +64,7 @@ private static String urlEncode(URI uri) {
     }
 
     private URI webURI() {
-        var webURI = from.webUrl().toString();
-        if (!webURI.endsWith(".git")) {
-            webURI += ".git";
-        }
+        var webURI = from.remoteUrl().toString();
```
erikj79, line 67: Converting from URI to String and then back to URI doesn't seem like a good idea. This method is called in one place, so should probably just get rid of it.

zhaosongzs: Oh, I didn't find this redundant operation.

bots/checkout/src/main/java/org/openjdk/skara/bots/checkout/CheckoutBot.java
```diff
@@ -64,12 +64,7 @@ private static String urlEncode(URI uri) {
     }
 
     private URI webURI() {
-        var webURI = from.webUrl().toString();
-        if (!webURI.endsWith(".git")) {
-            webURI += ".git";
-        }
-
-        return URI.create(webURI);
+        return from.remoteUrl();
```
erikj79, line 67: I meant that this whole method is redundant now and can just be inlined.

zhaosongzs: Sorry, fixed it now.

lgxbslgx (approved):
Looks good. The name `remoteUrl` doesn't seem very good, but with the comment, it is OK.

erikj79:
> Looks good. The name `remoteUrl` doesn't seem very good, but with the comment, it is OK.

I agree, it's hard to come up with a good name for this. Ideally I think this one should be just `url` and the existing `url` should be renamed something like `authenticatedUrl`.

zhaosongzs:
> > Looks good. The name `remoteUrl` doesn't seem very good, but with the comment, it is OK.
> 
> I agree, it's hard to come up with a good name for this. Ideally I think this one should be just `url` and the existing `url` should be renamed something like `authenticatedUrl`.

Yes, I found it difficult to come up with a name at that time. I will rename the methods.

zhaosongzs:
This pr is still waiting for review.

bots/notify/src/test/java/org/openjdk/skara/bots/notify/issue/IssueNotifierTests.java
```diff
@@ -1883,8 +1883,8 @@ void testIssueOriginalRepo(TestInfo testInfo) throws IOException {
             var comment = comments.get(0);
             assertTrue(comment.body().contains(editHash.toString()));
             // Verify that the 'original' repo URL is used in the comment and not the main one
-            assertTrue(comment.body().contains(originalRepo.url().toString()));
-            assertFalse(comment.body().contains(repo.url().toString()));
+            assertTrue(comment.body().contains(originalRepo.authenticatedUrl().toString()));
+            assertFalse(comment.body().contains(repo.authenticatedUrl().toString()));
```
erikj79, line 1887: I think we should be checking for `.url()` here.

bots/notify/src/test/java/org/openjdk/skara/bots/notify/issue/IssueNotifierTests.java
```diff
@@ -1986,7 +1986,7 @@ void testAltFixVersionsMatch(TestInfo testInfo) throws IOException {
             assertEquals(1, comments.size());
             var comment = comments.get(0);
             assertTrue(comment.body().contains(editHash.toString()));
-            assertTrue(comment.body().contains(repo.url().toString()));
+            assertTrue(comment.body().contains(repo.authenticatedUrl().toString()));
```
erikj79, line 1989: Same here.

bots/notify/src/test/java/org/openjdk/skara/bots/notify/issue/IssueNotifierTests.java
```diff
@@ -2034,7 +2034,7 @@ void testAltFixVersionsMatchRegex(TestInfo testInfo) throws IOException {
...
```
erikj79, line 2037: And again.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -1213,7 +1213,7 @@ private void checkStatus() {
 
             if (pr.sourceRepository().isPresent()) {
                 var branchNames = pr.repository().branches().stream().map(HostedBranch::name).collect(Collectors.toSet());
-                if (!pr.repository().url().equals(pr.sourceRepository().get().url()) && branchNames.contains(pr.sourceRef())) {
+                if (!pr.repository().authenticatedUrl().equals(pr.sourceRepository().get().authenticatedUrl()) && branchNames.contains(pr.sourceRef())) {
```
erikj79, line 1216: I think we should be comparing `.url()` here. If we compare `authenticatedUrl()` there is a risk that different user authentications would trigger a false negative.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/LimitedCensusInstance.java
```diff
@@ -115,7 +114,7 @@ private static Optional<JCheckConfiguration> jCheckConfiguration(
     }
 
     private static Path getRepoFolder(HostedRepositoryPool hostedRepositoryPool, HostedRepository censusRepo, String censusRef, Path folder) {
-        var repoName = censusRepo.url().getHost() + "/" + censusRepo.name();
+        var repoName = censusRepo.authenticatedUrl().getHost() + "/" + censusRepo.name();
```
erikj79, line 117: Only need `.url()` to extract host.

bots/tester/src/test/java/org/openjdk/skara/bots/tester/InMemoryHostedRepository.java
```diff
@@ -234,6 +234,11 @@ public URI webUrl(Tag tag) {
         return null;
     }
 
+    @Override
+    public URI url() {
+        return null;
```
erikj79, line 239: It probably doesn't matter much, but seems like this should return `url` just like `authenticatedUrl()` does.

cli/src/main/java/org/openjdk/skara/cli/pr/GitPrCreate.java
```diff
@@ -305,7 +305,7 @@ public static void main(String[] args) throws IOException, InterruptedException
         }
 
         var mailingLists = new ArrayList<String>();
-        var parentProject = ForgeUtils.projectName(parentRepo.url());
+        var parentProject = ForgeUtils.projectName(parentRepo.authenticatedUrl());
```
erikj79, line 308: Should probably be `.url()`.

forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabMergeRequest.java
```diff
@@ -530,7 +530,7 @@ private String metadataFor(Check check) {
     }
 
     private String linkToDiff(String path, Hash hash, int line) {
-        return "[" + path + " line " + line + "](" + URIBuilder.base(repository.url())
+        return "[" + path + " line " + line + "](" + URIBuilder.base(repository.authenticatedUrl())
```
erikj79, line 533: This looks like another case of `.url()`. The authentication is even removed below. This string is built to be presented as text somewhere.

test/src/main/java/org/openjdk/skara/test/TestHostedRepository.java
```diff
@@ -56,7 +56,7 @@ public TestHostedRepository(TestHost host, String projectName, Repository localR
         this.host = host;
         this.projectName = projectName;
         this.localRepository = localRepository;
-        pullRequestPattern = Pattern.compile(url().toString() + "/pr/" + "(\\d+)");
+        pullRequestPattern = Pattern.compile(authenticatedUrl().toString() + "/pr/" + "(\\d+)");
```
erikj79, line 59: Doesn't matter much as they resolve to the same value, but logically I think `.webUrl()` fits better here.

test/src/main/java/org/openjdk/skara/test/TestPullRequest.java
```diff
@@ -214,7 +214,7 @@ public boolean isDraft() {
     @Override
     public URI webUrl() {
         try {
-            return new URI(targetRepository.url().toString() + "/pr/" + id());
+            return new URI(targetRepository.authenticatedUrl().toString() + "/pr/" + id());
```
erikj79, line 217: Logically this should be `.webUrl()`.

erikj79 (commented):
This turned into a big patch. I looked through all the `url()` -> `authenticatedUrl()` changes and found some that should be reverted or changed in some way.

bots/hgbridge/src/main/java/org/openjdk/skara/bots/hgbridge/JBridgeBot.java
```diff
@@ -120,16 +120,16 @@ public Collection<WorkItem> run(Path scratchPath) {
 
             IOException lastException = null;
             for (var destination : exporterConfig.destinations()) {
-                var markerBase = destination.url().getHost() + "/" + destination.name();
+                var markerBase = destination.authenticatedUrl().getHost() + "/" + destination.name();
```
zhaosongzs, line 123: I think here should be `.url()` too.

zhaosongzs:
@erikj79 Thank you very much for reviewing this big patch and identifying some issues!

erikj79 (approved):

PR #1469, opened 2023-02-09
1588: Bridge messages should not be sent for PRs in draft state

lgxbslgx:
Hi all,

This patch prevents the `mailling bridge bot` from sending emails when a PR is draft.
And the related test case is added.

Best Regards,
-- Guoxiong

---------
### Progress

### Issue
 * [SKARA-1588](https://bugs.openjdk.org/browse/SKARA-1588): Bridge messages should not be sent for PRs in draft state

### Reviewers
 * [Zhao Song](https://openjdk.org/census#zsong) (@zhaosongzs - Committer) ⚠️ Review applies to [cdcfe380](https://git.openjdk.org/skara/pull/1469/files/cdcfe38057166bd94d6b834cf3e7dd13b5d86f33)
 * [Erik Joelsson](https://openjdk.org/census#erikj) (@erikj79 - **Reviewer**) ⚠️ Review applies to [4d50eaad](https://git.openjdk.org/skara/pull/1469/files/4d50eaade87cfec1218e7341d563ed7eb853f85b)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1469/head:pull/1469` \
`$ git checkout pull/1469`

`$ git checkout pull/1469` \
`$ git pull https://git.openjdk.org/skara pull/1469/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1469`

`$ git pr show -t 1469`

</details>

<a href="https://git.openjdk.org/skara/pull/1469.diff">https://git.openjdk.org/skara/pull/1469.diff</a>

</details>

### DIFF
--- a/bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveWorkItem.java
+++ b/bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveWorkItem.java
@@ -117,7 +117,7 @@ class ArchiveWorkItem implements WorkItem {
         }
     }
 
-    private boolean ignoreComment(HostUser author, String body) {
+    private boolean ignoreComment(HostUser author, String body, ZonedDateTime createdTime) {
         if (pr.repository().forge().currentUser().equals(author)) {
             return true;
         }
@@ -135,6 +135,14 @@ class ArchiveWorkItem implements WorkItem {
         for (var ignoredCommentPattern : bot.ignoredComments()) {
             var ignoredCommentMatcher = ignoredCommentPattern.matcher(body);
             if (ignoredCommentMatcher.find()) {
+                return true;
+            }
+        }
+        // If the pull request was converted to draft, the comments
+        // after the last converted time should be ignored.
+        if (pr.isDraft()) {
+            var lastDraftTime = pr.lastMarkedAsDraftTime();
+            if (lastDraftTime.isPresent() && lastDraftTime.get().isBefore(createdTime)) {
                 return true;
             }
         }
@@ -338,7 +346,7 @@ class ArchiveWorkItem implements WorkItem {
 
             // Regular comments
             for (var comment : comments) {
-                if (ignoreComment(comment.author(), comment.body())) {
+                if (ignoreComment(comment.author(), comment.body(), comment.createdAt())) {
                     archiver.addIgnored(comment);
                 } else {
                     archiver.addComment(comment);
@@ -348,7 +356,7 @@ class ArchiveWorkItem implements WorkItem {
             // Review comments
             var reviews = pr.reviews();
             for (var review : reviews) {
-                if (ignoreComment(review.reviewer(), review.body().orElse(""))) {
+                if (ignoreComment(review.reviewer(), review.body().orElse(""), review.createdAt())) {
                     continue;
                 }
                 archiver.addReview(review);
@@ -360,7 +368,7 @@ class ArchiveWorkItem implements WorkItem {
                                    .sorted(Comparator.comparing(ReviewComment::path))
                                    .collect(Collectors.toList());
             for (var reviewComment : reviewComments) {
-                if (ignoreComment(reviewComment.author(), reviewComment.body())) {
+                if (ignoreComment(reviewComment.author(), reviewComment.body(), reviewComment.createdAt())) {
                     continue;
                 }
                 archiver.addReviewComment(reviewComment);
--- a/bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveWorkItem.java
+++ b/bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveWorkItem.java
@@ -117,7 +117,7 @@ class ArchiveWorkItem implements WorkItem {
         }
     }
 
-    private boolean ignoreComment(HostUser author, String body, ZonedDateTime createdTime) {
+    private boolean ignoreComment(HostUser author, String body, ZonedDateTime createdTime, ZonedDateTime lastDraftTime) {
         if (pr.repository().forge().currentUser().equals(author)) {
             return true;
         }
@@ -141,8 +141,7 @@ class ArchiveWorkItem implements WorkItem {
         // If the pull request was converted to draft, the comments
         // after the last converted time should be ignored.
         if (pr.isDraft()) {
-            var lastDraftTime = pr.lastMarkedAsDraftTime();
-            if (lastDraftTime.isPresent() && lastDraftTime.get().isBefore(createdTime)) {
+            if (lastDraftTime != null && lastDraftTime.isBefore(createdTime)) {
                 return true;
             }
         }
@@ -343,10 +342,11 @@ class ArchiveWorkItem implements WorkItem {
             var webrevPath = scratchPath.resolve("mlbridge-webrevs");
             var listServer = MailingListServerFactory.createMailmanServer(bot.listArchive(), bot.smtpServer(), bot.sendInterval());
             var archiver = new ReviewArchive(pr, bot.emailAddress());
+            var lastDraftTime = pr.lastMarkedAsDraftTime().orElse(null);
 
             // Regular comments
             for (var comment : comments) {
-                if (ignoreComment(comment.author(), comment.body(), comment.createdAt())) {
+                if (ignoreComment(comment.author(), comment.body(), comment.createdAt(), lastDraftTime)) {
                     archiver.addIgnored(comment);
                 } else {
                     archiver.addComment(comment);
@@ -356,7 +356,7 @@ class ArchiveWorkItem implements WorkItem {
             // Review comments
             var reviews = pr.reviews();
             for (var review : reviews) {
-                if (ignoreComment(review.reviewer(), review.body().orElse(""), review.createdAt())) {
+                if (ignoreComment(review.reviewer(), review.body().orElse(""), review.createdAt(), lastDraftTime)) {
                     continue;
                 }
                 archiver.addReview(review);
@@ -368,7 +368,7 @@ class ArchiveWorkItem implements WorkItem {
                                    .sorted(Comparator.comparing(ReviewComment::path))
                                    .collect(Collectors.toList());
             for (var reviewComment : reviewComments) {
-                if (ignoreComment(reviewComment.author(), reviewComment.body(), reviewComment.createdAt())) {
+                if (ignoreComment(reviewComment.author(), reviewComment.body(), reviewComment.createdAt(), lastDraftTime)) {
                     continue;
                 }
                 archiver.addReviewComment(reviewComment);
--- a/bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/MailingListBridgeBot.java
+++ b/bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/MailingListBridgeBot.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2022, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -203,7 +203,6 @@ public class MailingListBridgeBot implements Bot {
 
         List<PullRequest> prs = poller.updatedPullRequests();
         prs.stream()
-                .filter(pr -> !pr.isDraft())
                 .map(pr -> new ArchiveWorkItem(pr, this,
                         e -> poller.retryPullRequest(pr),
                         r -> poller.quarantinePullRequest(pr, r)))
--- a/bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/MailingListBridgeBot.java
+++ b/bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/MailingListBridgeBot.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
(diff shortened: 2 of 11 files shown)


zhaosongzs (approved):
I think this patch could successfully help us postpone the sending of emails when the pr is in draft mode and some users will be happy to see this update.

bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/MailingListBridgeBot.java
```diff
@@ -203,6 +203,7 @@ public List<WorkItem> getPeriodicItems() {
 
         List<PullRequest> prs = poller.updatedPullRequests();
         prs.stream()
+                .filter(pr -> !pr.isDraft())
```
kevinrushforth, line 206: Are you sure this is the right place to do this? The intention of this fix is to pause notifications for Draft PRs until it becomes "rfr" again, just like it currently does for PRs initially in Draft that become "rfr" for the first time. Will it do this or will it just lose the notifications permanently?

kevinrushforth (commented):
Not sure whether this is sufficient. See my question below.

bots/mlbridge/src/test/java/org/openjdk/skara/bots/mlbridge/MailingListBridgeBotTests.java
```diff
@@ -3437,4 +3437,120 @@ void commentWithQuoteFromBridged(TestInfo testInfo) throws IOException {
...
+            pr.makeNotDraft();
+
+            // Run another archive pass.
+            TestBotRunner.runPeriodicItems(mlBot);
+
+            // The archive should now contain the new comment.
+            Repository.materialize(archiveFolder.path(), archive.url(), "master");
+            assertEquals(3, archiveContainsCount(archiveFolder.path(), "RFR: 1234: This is a pull request"));
+            assertEquals(1, archiveContainsCount(archiveFolder.path(), "This is a comment"));
+            assertEquals(1, archiveContainsCount(archiveFolder.path(), "This is a new comment"));
```
zhaosongzs, line 3553: @kevinrushforth Hi Kevin, see Li's test case here. This pr was 'rfr' and he converted it to draft. When the pr was in draft mode, he added a comment 'This is a new comment' and run the bot. As expected, there is no email sent. After that, he converted the pr to 'rfr' again and run the bot. This time, he got the notification about the comment 'This is a new comment'. So we won't lose the notification permanently. It's just delayed.

kevinrushforth: OK, thanks.

erikj79 (commented):
This will indeed pause emails while a PR is in draft state. When it gets back from draft, I believe everything that happened while in draft will be sent. That is certainly the easiest way to handle this and I'm good with this change. We will just have to see how it plays out in practice.

erikj79:
> This will indeed pause emails while a PR is in draft state. When it gets back from draft, I believe everything that happened while in draft will be sent. That is certainly the easiest way to handle this and I'm good with this change. We will just have to see how it plays out in practice.

Thinking about this some more, this simple solution has a pretty big drawback. There is no actual synchronization between events. All bot actions are asynchronous. Imagine the following scenario.

1. User posts comment "I'm putting this PR in draft to work on the feedback"
2. User puts PR in draft
3. mlbridge bot processes PR, finds it in draft and does nothing
4. User removes draft state
5. mlbridge bot processes PR, finds new comment from 1 and posts it.

For this to work correctly, draft PRs still need to be processed by mlbridge, and any events up until the draft state change need to be processed as usual.

zhaosongzs:
> > This will indeed pause emails while a PR is in draft state. When it gets back from draft, I believe everything that happened while in draft will be sent. That is certainly the easiest way to handle this and I'm good with this change. We will just have to see how it plays out in practice.
> 
> Thinking about this some more, this simple solution has a pretty big drawback. There is no actual synchronization between events. All bot actions are asynchronous. Imagine the following scenario.
> 
> 1. User posts comment "I'm putting this PR in draft to work on the feedback"
> 2. User puts PR in draft
> 3. mlbridge bot processes PR, finds it in draft and does nothing
> 4. User removes draft state
> 5. mlbridge bot processes PR, finds new comment from 1 and posts it.
> 
> For this to work correctly, draft PRs still need to be processed by mlbridge, and any events up until the draft state change need to be processed as usual.
> 
> This is still possibly better than what we have, but I'm not sure. Opinions?

I'm not sure I understand it correctly. Is the problem that the notification 'I'm putting this PR in draft to work on the feedback' won't be sent until the user makes the PR out of draft mode? Maybe it's not a big problem, the window is not very big, just few minutes.

erikj79:
> I'm not sure I understand it correctly. Is the problem that the notification 'I'm putting this PR in draft to work on the feedback' won't be sent until the user makes the PR out of draft mode? Maybe it's not a big problem, the window is not very big, just few minutes.

Yes, the problem is that the last comments or events before changing a PR to draft could be delayed until the PR is taken out of draft. The size of the window for this to happen is usually just a few minutes, but depending on the state of the bots, it could be much longer. My example scenario seems like a very plausible series of events however, a user comments about making it draft and then puts it in draft, more or less at the same time. In that case it's very likely that the last comment hits the window. Also delaying that particular message would be very confusing to users, and I can guarantee that if it happened, we would get a lot of complaints. A PR could be moved to draft for weeks, and suddenly receiving week old comments from before it was put in draft is pretty bad.

zhaosongzs:
> > I'm not sure I understand it correctly. Is the problem that the notification 'I'm putting this PR in draft to work on the feedback' won't be sent until the user makes the PR out of draft mode? Maybe it's not a big problem, the window is not very big, just few minutes.
> 
> Yes, the problem is that the last comments or events before changing a PR to draft could be delayed until the PR is taken out of draft. The size of the window for this to happen is usually just a few minutes, but depending on the state of the bots, it could be much longer. My example scenario seems like a very plausible series of events however, a user comments about making it draft and then puts it in draft, more or less at the same time. In that case it's very likely that the last comment hits the window. Also delaying that particular message would be very confusing to users, and I can guarantee that if it happened, we would get a lot of complaints. A PR could be moved to draft for weeks, and suddenly receiving week old comments from before it was put in draft is pretty bad.

Ok, I think the solution is that we don't filter in `MailingListBridgeBot#getPeriodicItems`. Instead, in `ArchiveWorkItem#run`, we can check if the pr is in draft mode and record the time. This recorded time can then be used to filter the comments.

erikj79:
> Ok, I think the solution is that we don't filter in `MailingListBridgeBot#getPeriodicItems`. Instead, in `ArchiveWorkItem#run`, we can check if the pr is in draft mode and record the time. This recorded time can then be used to filter the comments.

There should be events that we can query for with timestamps for when the state change happened. The check would be something like this: If pr.isDraft(), find last event that put it in draft and use the timestamp of that as filter for any email generation.

lgxbslgx:
> Thinking about this some more, this simple solution has a pretty big drawback. There is no actual synchronization between events. All bot actions are asynchronous. Imagine the following scenario.
> 
> 1. User posts comment "I'm putting this PR in draft to work on the feedback"
> 2. User puts PR in draft
> 3. mlbridge bot processes PR, finds it in draft and does nothing
> 4. User removes draft state
> 5. mlbridge bot processes PR, finds new comment from 1 and posts it.
> 
> For this to work correctly, draft PRs still need to be processed by mlbridge, and any events up until the draft state change need to be processed as usual.
> 
> This is still possibly better than what we have, but I'm not sure. Opinions?

Good Caught. This is a case we need to solve.

I updated the code just now, which ignores the comments in method `ArchiveWorkItem#ignoreComment`
instead of filtering the draft PRs in method `MailingListBridgeBot#getPeriodicItems`.

In order to get the time when the PR was last marked as draft, I add a new API `lastMarkedAsDraftTime` to `PullRequest` 
and implement it in the sub-classes. The related tests about this new case and the new API are also added.

zhaosongzs:
Now, I could remember that my implementation of [SKARA-1683](https://github.com/openjdk/skara/pull/1426) shares the same issue with Li's first version.  I had previously mentioned in that pr "My current implementation has a bug. If a user force-pushes or rebases when the pr is in ready state, and then he converts the pr to draft state very quickly, in such a case, our pullRequestBot won't have enough time to send warning, and then user will never get the notification of this force-push or rebase.". I will fix this issue later this week .

bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveWorkItem.java
```diff
@@ -138,6 +138,14 @@ private boolean ignoreComment(HostUser author, String body) {
                 return true;
             }
         }
+        // If the pull request was converted to draft, the comments
+        // after the last converted time should be ignored.
+        if (pr.isDraft()) {
+            var lastDraftTime = pr.lastMarkedAsDraftTime();
```
erikj79, line 144: This method gets called for every comment on the PR, every time `ArchiveWorkItem` is run for a PR. The result of `pr.lastMarkedAsDraftTime()` needs to be retrieved once and reused.

forge/src/main/java/org/openjdk/skara/forge/github/GitHubPullRequest.java
```diff
@@ -703,6 +703,20 @@ public void makeNotDraft() {
             .execute();
     }
 
+    @Override
+    public Optional<ZonedDateTime> lastMarkedAsDraftTime() {
+        var lastMarkedAsDraftTime = request.get("issues/" + json.get("number").toString() + "/timeline")                .execute().stream()
+                .map(JSONValue::asObject)
+                .filter(obj -> obj.contains("event"))
+                .filter(obj -> obj.get("event").asString().equals("convert_to_draft"))
+                .reduce((a, b) -> b)
```
erikj79, line 712: Are you sure of the order events are returned in? I couldn't find a definition of the order. Maybe safer to map to date and reduce to the highest one?

forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabMergeRequest.java
```diff
@@ -781,6 +781,22 @@ public void makeNotDraft() {
...
 
+    @Override
+    public Optional<ZonedDateTime> lastMarkedAsDraftTime() {
+        var draftMessage = "marked this merge request as **draft**";
+        var notes = request.get("notes").execute();
+        var lastMarkedAsDraftTime = notes.stream()
+                .map(JSONValue::asObject)
+                .filter(obj -> obj.get("system").asBoolean())
+                .filter(obj -> draftMessage.equals(obj.get("body").asString()))
+                .reduce((a, b) -> a)
```
erikj79, line 792: Same here as for GitHub, maybe reduce to highest date just to be safe?

erikj79 (commented):
This looks pretty good, but see comments. Were you able to run the manual tests to verify the new methods on PullRequest?

bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveWorkItem.java
lgxbslgx: I move `pr.lastMarkedAsDraftTime()` to the method `ArchiveWorkItem#run` to avoid the duplicated invocation.

forge/src/main/java/org/openjdk/skara/forge/github/GitHubPullRequest.java
lgxbslgx: I can't find the related document too. I tried the rest api to get the related data and found the order personally. Now I adjust the code to find the highest one to avoid the API result changes in the future.

forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabMergeRequest.java
lgxbslgx: Same as above and also adjust it.

lgxbslgx:
> Were you able to run the manual tests to verify the new methods on PullRequest?

I had run the manual tests locally. Everything looks fine.

forge/src/main/java/org/openjdk/skara/forge/github/GitHubPullRequest.java
```diff
@@ -703,6 +703,20 @@ public void makeNotDraft() {
...
     }
 
+    @Override
+    public Optional<ZonedDateTime> lastMarkedAsDraftTime() {
+        var lastMarkedAsDraftTime = request.get("issues/" + json.get("number").toString() + "/timeline")                .execute().stream()
+                .map(JSONValue::asObject)
+                .filter(obj -> obj.contains("event"))
+                .filter(obj -> obj.get("event").asString().equals("convert_to_draft"))
+                .map(obj -> ZonedDateTime.parse(obj.get("created_at").asString()))
+                .reduce((a, b) -> a.isBefore(b) ? b : a);
```
erikj79, line 713: I think this could be expressed as `max()`.

lgxbslgx: Fixed.

erikj79 (approved):

lgxbslgx:
It seems a new bug:

When a PR has become ready and then another PR, of which code conflicts with the former one, was integrated, the former PR remains ready. But actually, it should have a `merge-conflict label and should not be ready.

This PR-1469 (SKARA-1588) has became ready and then PR-1472 (SKARA-1817) was integrated. This PR is still ready now. But actually, this PR should have a `merge-conflict` label and should not be ready (the `ready` label should be removed and the related comment aboud conflict should be added).

I suspect it is because PR bot have not been triggerred in this PR (since this PR has not been updated). This comment can update this PR and can verify my idea.

lgxbslgx:
> I suspect it is because PR bot have not been triggered in this PR (since this PR has not been updated). This comment can update this PR and can verify my idea.

This suspicion looks wrong now.

lgxbslgx:
> It seems a new bug:
> 
> When a PR has become ready and then another PR, of which code conflicts with the former one, was integrated, the former PR remains ready. But actually, it should have a `merge-conflict` label and should not be ready.
> 
> This PR-1469 (SKARA-1588) has became ready and then PR-1472 (SKARA-1817) was integrated. This PR is still ready now. But actually, this PR should have a `merge-conflict` label and should not be ready (the `ready` label should be removed and the related comment aboud conflict should be added).
> 
> I suspect it is because PR bot have not been triggered in this PR (since this PR has not been updated). This comment can update this PR and can verify my idea.

Filed https://bugs.openjdk.org/browse/SKARA-1820 to record it.

PR #1470, opened 2023-02-09
1706: Store github checks in PR unique way

zhaosongzs:
Currently, checks in GitHub are stored in the commit. So if multiple PRs share the same source commit, then they will compete with the check-runs and store their different checksum in there. This in turn means that all, or most of those PRs will always think the check is outdated when a CheckRun is performed next.
 
This patch adds repository name and pull request id to the name of check in Github. Therefore, even if multiple PRs share the same source commit, every PR will store a unique check in the commit and there is no competition exists.

---------
### Progress

### Issue
 * [SKARA-1706](https://bugs.openjdk.org/browse/SKARA-1706): Store github checks in PR unique way

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1470/head:pull/1470` \
`$ git checkout pull/1470`

`$ git checkout pull/1470` \
`$ git pull https://git.openjdk.org/skara pull/1470/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1470`

`$ git pr show -t 1470`

</details>

<a href="https://git.openjdk.org/skara/pull/1470.diff">https://git.openjdk.org/skara/pull/1470.diff</a>

</details>

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -1085,8 +1085,12 @@ class CheckRun {
         pr.addComment(message);
     }
 
+    static String getJcheckName(PullRequest pr) {
+        return pr.repository().forge().name().equals("GitHub") ? "jcheck-" + pr.repository().name() + "-" + pr.id() : "jcheck";
+    }
+
     private void checkStatus() {
-        var checkBuilder = CheckBuilder.create("jcheck", pr.headHash());
+        var checkBuilder = CheckBuilder.create(getJcheckName(pr), pr.headHash());
         var censusDomain = censusInstance.configuration().census().domain();
         Exception checkException = null;
 
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/IntegrateCommand.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/IntegrateCommand.java
@@ -61,6 +61,9 @@ public class IntegrateCommand implements CommandHandler {
         final var inProgress = "the status check `" + checkName + "` is still in progress";
         final var outdated = "the status check `" + checkName + "` has not been performed on commit %s yet";
 
+        checkName = checkName.equals("jcheck") && pr.repository().forge().name().equals("GitHub") ?
+                "jcheck" + pr.repository().name() + pr.id() : "jcheck";
+
         if (performedChecks.containsKey(checkName)) {
             var check = performedChecks.get(checkName);
             if (check.status() == CheckStatus.SUCCESS) {
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/IntegrateCommand.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/IntegrateCommand.java
@@ -61,9 +61,6 @@ public class IntegrateCommand implements CommandHandler {
         final var inProgress = "the status check `" + checkName + "` is still in progress";
         final var outdated = "the status check `" + checkName + "` has not been performed on commit %s yet";
 
-        checkName = checkName.equals("jcheck") && pr.repository().forge().name().equals("GitHub") ?
-                "jcheck" + pr.repository().name() + pr.id() : "jcheck";
-
         if (performedChecks.containsKey(checkName)) {
             var check = performedChecks.get(checkName);
             if (check.status() == CheckStatus.SUCCESS) {
@@ -170,7 +167,7 @@ public class IntegrateCommand implements CommandHandler {
             return;
         }
 
-        var problem = checkProblem(pr.checks(pr.headHash()), "jcheck", pr);
+        var problem = checkProblem(pr.checks(pr.headHash()), CheckRun.getJcheckName(pr), pr);
         if (problem.isPresent()) {
             reply.print("Your integration request cannot be fulfilled at this time, as ");
             reply.println(problem.get());
--- a/forge/src/main/java/org/openjdk/skara/forge/github/GitHubPullRequest.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/github/GitHubPullRequest.java
@@ -461,7 +461,7 @@ public class GitHubPullRequest implements PullRequest {
     @Override
     public void updateCheck(Check check) {
         var completedQuery = JSON.object();
-        completedQuery.put("name", check.name().equals("jcheck") ? check.name() + repository().name() + id() : check.name());
+        completedQuery.put("name", check.name());
         completedQuery.put("head_branch", json.get("head").get("ref"));
         completedQuery.put("head_sha", check.hash().hex());
 
--- a/forge/src/main/java/org/openjdk/skara/forge/github/GitHubPullRequest.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/github/GitHubPullRequest.java
@@ -461,7 +461,7 @@ public class GitHubPullRequest implements PullRequest {
     @Override
     public void updateCheck(Check check) {
         var completedQuery = JSON.object();
-        completedQuery.put("name", check.name());
+        completedQuery.put("name", check.name().equals("jcheck") ? check.name() + repository().name() + id() : check.name());
         completedQuery.put("head_branch", json.get("head").get("ref"));
         completedQuery.put("head_sha", check.hash().hex());
 
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
@@ -120,7 +120,7 @@ class CheckWorkItem extends PullRequestWorkItem {
         var hash = pr.headHash();
         var metadata = getMetadata(censusInstance, pr.title(), pr.body(), comments, reviews, labels, pr.targetRef(), pr.isDraft(), null);
         var currentChecks = pr.checks(hash);
-        var jcheckName = pr.repository().forge().name().equals("GitHub") ? "jcheck" + pr.repository().name() + pr.id() : "jcheck";
+        var jcheckName = CheckRun.getJcheckName(pr);
 
         if (currentChecks.containsKey(jcheckName)) {
             var check = currentChecks.get(jcheckName);
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
@@ -120,9 +120,10 @@ class CheckWorkItem extends PullRequestWorkItem {
         var hash = pr.headHash();
         var metadata = getMetadata(censusInstance, pr.title(), pr.body(), comments, reviews, labels, pr.targetRef(), pr.isDraft(), null);
         var currentChecks = pr.checks(hash);
+        var jcheckName = pr.repository().forge().name().equals("GitHub") ? "jcheck" + pr.repository().name() + pr.id() : "jcheck";
 
-        if (currentChecks.containsKey("jcheck")) {
-            var check = currentChecks.get("jcheck");
+        if (currentChecks.containsKey(jcheckName)) {
+            var check = currentChecks.get(jcheckName);
             if (check.completedAt().isPresent() && check.metadata().isPresent()) {
                 var previousMetadata = check.metadata().get();
                 if (previousMetadata.contains(":")) {


zhaosongzs:
Checked the [checkrun](https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#get-a-check-run) document in GitHub and I found the document says there will be an `id` for every check-run. The document also says if we want to update the check-run, `id` is required in the header. But in `GitHubPullRequest#updateCheck`, I don't think we have anything related with `id`. Therefore I am wondering how could we update the check in GitHub, do we update the checks by the name?

erikj79:
> Checked the [checkrun](https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#get-a-check-run) document in GitHub and I found the document says there will be an `id` for every check-run. The document also says if we want to update the check-run, `id` is required in the header. But in `GitHubPullRequest#updateCheck`, I don't think we have anything related with `id`. Therefore I am wondering how could we update the check in GitHub, do we update the checks by the name?

See comment at:
https://github.com/openjdk/skara/blob/51956054485fce5cb42629c9696a3131cb098178/forge/src/main/java/org/openjdk/skara/forge/github/GitHubPullRequest.java#L454

zhaosongzs:
> > Checked the [checkrun](https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#get-a-check-run) document in GitHub and I found the document says there will be an `id` for every check-run. The document also says if we want to update the check-run, `id` is required in the header. But in `GitHubPullRequest#updateCheck`, I don't think we have anything related with `id`. Therefore I am wondering how could we update the check in GitHub, do we update the checks by the name?
> 
> See comment at:
> 
> https://github.com/openjdk/skara/blob/51956054485fce5cb42629c9696a3131cb098178/forge/src/main/java/org/openjdk/skara/forge/github/GitHubPullRequest.java#L454

Thanks!

bots/pr/src/main/java/org/openjdk/skara/bots/pr/IntegrateCommand.java
```diff
@@ -61,6 +61,9 @@ private Optional<String> checkProblem(Map<String, Check> performedChecks, String
         final var inProgress = "the status check `" + checkName + "` is still in progress";
         final var outdated = "the status check `" + checkName + "` has not been performed on commit %s yet";
 
+        checkName = checkName.equals("jcheck") && pr.repository().forge().name().equals("GitHub") ?
```
erikj79, line 64: There is just one caller of this method, which explicitly calls with `"jcheck"`. The caller should generate the new check name instead of rewriting it here.

forge/src/main/java/org/openjdk/skara/forge/github/GitHubPullRequest.java
```diff
@@ -461,7 +461,7 @@ public void createCheck(Check check) {
     @Override
     public void updateCheck(Check check) {
         var completedQuery = JSON.object();
-        completedQuery.put("name", check.name());
+        completedQuery.put("name", check.name().equals("jcheck") ? check.name() + repository().name() + id() : check.name());
```
erikj79, line 464: I don't understand this change. Why would we need that? I don't think GitHubPullRequest should be changing check names.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/IntegrateCommand.java
zhaosongzs: But in that way, the checkName will be "jcheck-repo#pull" and the error message showed to user would be "the status check jcheck-repo#pull is still in progress".

forge/src/main/java/org/openjdk/skara/forge/github/GitHubPullRequest.java
zhaosongzs: As you said in the comment above "The check is created in CheckRun, so the name needs to be changed there too.", I didn't change the check name when it is created, so I changed it here. It's my bad, I will fix it.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/IntegrateCommand.java
erikj79: I think that's what we want. That's the name the check will have in the GitHub UI as well.

erikj79 (commented):
I think this looks ok. Could you run it in staging and verify that it works and looks ok before integrating?

zhaosongzs:
> I think this looks ok. Could you run it in staging and verify that it works and looks ok before integrating?

Sure. I will run my own fork on staging to verify it.

zhaosongzs:
> I think this looks ok. Could you run it in staging and verify that it works and looks ok before integrating?

I think it works fine. Currently, each pr would have a unique jcheck. https://github.com/openjdk/playground/pull/126/checks?check_run_id=11239907703

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -1085,8 +1085,12 @@ private void addMergeCommitWarningComment() {
         pr.addComment(message);
     }
 
+    static String getJcheckName(PullRequest pr) {
+        return pr.repository().forge().name().equals("GitHub") ? "jcheck-" + pr.repository().name() + "-" + pr.id() : "jcheck";
+    }
+
```
lgxbslgx, line 1091: Why does the `GitLab merge request` not need to change such name? I think GitLab also needs.

This method is now put in class `CheckRun` and is used in both `CheckRun` and `CheckWorkItem`. It seems not a good design. We can move the method to the interface `PullRequest` with a default implementation, just like below. Then we can use `pr.getJcheckName()` in both `CheckRun` and `CheckWorkItem`.

```
// file PullRequest.java
    default String getJcheckName() {
        return "jcheck-" + repository().name() + "-" + id();
    }
```

lgxbslgx:
> I think it works fine. Currently, each pr would have a unique jcheck. https://github.com/openjdk/playground/pull/126/checks?check_run_id=11239907703

From the link you provided, I notice there are three items `jcheck`, `jcheck-openjdk/playground-125` and `jcheck-openjdk/playground-126` under the `openjdk`. And the `PR-125`, which is sooner than `PR-126`, also has these three items, please see https://github.com/openjdk/playground/pull/125/checks.

And the `checks`, please see https://github.com/openjdk/skara/pull/1470/checks, of this pull request, it contains both `jcheck` and `jcheck-openjdk/skara-1470`.  And under the `All checks have passed` of this page (this pull request page), it also has two items `jcheck` and `jcheck-openjdk/skara-1470`. It is so strange. I suspect the current patch is not the right solution.

erikj79:
> From the link you provided, I notice there are three items `jcheck`, `jcheck-openjdk/playground-125` and `jcheck-openjdk/playground-126` under the `openjdk`. And the `PR-125`, which is sooner than `PR-126`, also has these three items, please see https://github.com/openjdk/playground/pull/125/checks.

That's a good point and the kind of thing I wanted to catch by testing it in staging. This is caused by testinfo bot copying all checks from the source branch into the pull request. This copying will need to be filtered so that only the relevant jcheck checks are copied.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
erikj79: Because there is no concept of a "check" in GitLab. We just store the check results in a comment in the PR, so there is no reason to make the name of the check unique.

Until now I my opinion was that the name of the jcheck check was something only the pull request bot module needed to know about, so putting that on the `PullRequest` interface didn't seem right. It was an implementation detail. But now since we need to share this information with testinfo, it would make sense to push this logic into `PullRequest`. The default method should in that case return "jcheck" and `GitHubPullRequest` can have the special implementation needed only on GitHub.

zhaosongzs:
> > From the link you provided, I notice there are three items `jcheck`, `jcheck-openjdk/playground-125` and `jcheck-openjdk/playground-126` under the `openjdk`. And the `PR-125`, which is sooner than `PR-126`, also has these three items, please see https://github.com/openjdk/playground/pull/125/checks.
> 
> That's a good point and the kind of thing I wanted to catch by testing it in staging. This is caused by testinfo bot copying all checks from the source branch into the pull request. This copying will need to be filtered so that only the relevant jcheck checks are copied.

Yes, I will make testinfo bot only copy the right jcheck run.

zhaosongzs:
> > I think it works fine. Currently, each pr would have a unique jcheck. https://github.com/openjdk/playground/pull/126/checks?check_run_id=11239907703
> 
> From the link you provided, I notice there are three items `jcheck`, `jcheck-openjdk/playground-125` and `jcheck-openjdk/playground-126` under the `openjdk`. And the `PR-125`, which is sooner than `PR-126`, also has these three items, please see https://github.com/openjdk/playground/pull/125/checks.
> 
> And the `checks`, please see https://github.com/openjdk/skara/pull/1470/checks, of this pull request, it contains both `jcheck` and `jcheck-openjdk/skara-1470`. And under the `All checks have passed` of this page (this pull request page), it also has two items `jcheck` and `jcheck-openjdk/skara-1470`. It is so strange. I suspect the current patch is not the right solution.

For the PRs in playground, they contain `jcheck-openjdk/playground-125` and `jcheck-openjdk/playground-126` because PR-125 and PR-126 share the same source commit and testinfo bot will copy all the tests under the source commit. They also contain `jcheck` because after testing, I revert the staging bot, so it will have a normal `jcheck`. 

For this PR in SKARA,  it contains `jcheck-openjdk/skara-1470` because SKARA is also configured to pr bot in staging in our configuration. 

I think after implementing the filter logic in testinfo bot, the checks will look better.

Thank you!

zhaosongzs:
> > From the link you provided, I notice there are three items `jcheck`, `jcheck-openjdk/playground-125` and `jcheck-openjdk/playground-126` under the `openjdk`. And the `PR-125`, which is sooner than `PR-126`, also has these three items, please see https://github.com/openjdk/playground/pull/125/checks.
> 
> That's a good point and the kind of thing I wanted to catch by testing it in staging. This is caused by testinfo bot copying all checks from the source branch into the pull request. This copying will need to be filtered so that only the relevant jcheck checks are copied.

After testing, I realized that just filtering is insufficient for GitHub as checks in pull requests are same as checks in source commits(There is no need to copy). To display only relevant checks, we may need to store checks separately under the source commit.

erikj79:
> After testing, I realized that just filtering is insufficient for GitHub as checks in pull requests are same as checks in source commits(There is no need to copy). To display only relevant checks, we may need to store checks separately under the source commit.

Are you saying that the testinfo bot isn't needed?

zhaosongzs:
> > After testing, I realized that just filtering is insufficient for GitHub as checks in pull requests are same as checks in source commits(There is no need to copy). To display only relevant checks, we may need to store checks separately under the source commit.
> 
> Are you saying that the testinfo bot isn't needed?

Sorry, what I said is confusing. What I should say is that, jcheck is not triggered by the workflow, it is triggered by Checkrun. So it is not stored in the our own fork and then be copied to the pr.

erikj79:
> Sorry, what I said is confusing. What I should say is that, jcheck is not triggered by the workflow, it is triggered by Checkrun. So it is not stored in the our own fork and then be copied to the pr.

Right, but then how did we see all the different jcheck-<repo>-<pull> in the "checks" tab? Is that just github showing everything related to a commit? If so, I suppose it's ok, the naming should help telling them apart.

zhaosongzs:
> > Sorry, what I said is confusing. What I should say is that, jcheck is not triggered by the workflow, it is triggered by Checkrun. So it is not stored in the our own fork and then be copied to the pr.
> 
> Right, but then how did we see all the different jcheck-- in the "checks" tab? Is that just github showing everything related to a commit? If so, I suppose it's ok, the naming should help telling them apart.

I think so. And this problem would happen only when we create multiple prs with same source commit in the **same** repository.

zhaosongzs:
> > Sorry, what I said is confusing. What I should say is that, jcheck is not triggered by the workflow, it is triggered by Checkrun. So it is not stored in the our own fork and then be copied to the pr.
> 
> Right, but then how did we see all the different `jcheck-<repo>-<pull>` in the "checks" tab? Is that just github showing every check related to a commit? If so, I suppose it's ok, the naming should help telling them apart.

My test yesterday is not very good. I  created multiple pull requests with the same source commit within a single repository, whereas the problem this patch was intended to address was creating multiple pull requests with the same source commit in different repositories.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
lgxbslgx: > Because there is no concept of a "check" in GitLab. We just store the check results in a comment in the PR, so there is no reason to make the name of the check unique.

Thanks for the explanation.

> Until now my opinion was that the name of the jcheck check was something only the pull request bot module needed to know about, so putting that on the `PullRequest` interface didn't seem right. It was an implementation detail. But now since we need to share this information with testinfo, it would make sense to push this logic into `PullRequest`. The default method should in that case return "jcheck" and `GitHubPullRequest` can have the special implementation needed only on GitHub.

I understand what you considered. The name is indeed only used in PR bot. So now I think the current code is also okay.

zhaosongzs:
> > My test yesterday is not very good. I created multiple pull requests with the same source commit within a single repository, whereas the problem this patch was intended to address was creating multiple pull requests with the same source commit in different repositories.
> 
> I would say the intention is to handle both situations.

Yes, seems like that checks cannot be stored within a specific PR and must instead be stored in a specific commit. I also reviewed the GitHub documentation, it seems that it is not possible to hide certain checks within a PR. I am afraid that I am unable to improve the current code, however, I believe that it will work correctly. The names of the checks will indicate to the user which one is associated with the PR, and only the relevant check will have impact. Checks from other PRs will not be an integrationBlocker for this PR.

erikj79 (approved):

erikj79:
> > I would say the intention is to handle both situations.
> 
> Yes, seems like that checks cannot be stored within a specific PR and must instead be stored in a specific commit. I also reviewed the GitHub documentation, it seems that it is not possible to hide certain checks within a PR. I am afraid that I am unable to improve the current code, however, I believe that it will work correctly. The names of the checks will indicate to the user which one is associated with the PR, and only the relevant check will have impact. Checks from other PRs will not be an integrationBlocker for this PR.

Then I think this looks good.

lgxbslgx (approved):

PR #1475, opened 2023-02-16
1271: Webrev generation should be optional

zhaosongzs:
Currently, webrevs are generated automatically when a user pushes a new commit, and it's mandatory to choose either JSON or HTML when configuring the mlbridge bot. If both of them are disabled in the configuration, the bot will throw errors.

In this patch, generating a webrev will be optional. If both JSON and HTML are disabled, no webrev will be generated.

---------
### Progress

### Issue
 * [SKARA-1271](https://bugs.openjdk.org/browse/SKARA-1271): Webrev generation should be optional

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1475/head:pull/1475` \
`$ git checkout pull/1475`

`$ git checkout pull/1475` \
`$ git pull https://git.openjdk.org/skara pull/1475/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1475`

`$ git pr show -t 1475`

</details>

<a href="https://git.openjdk.org/skara/pull/1475.diff">https://git.openjdk.org/skara/pull/1475.diff</a>

</details>

### DIFF
--- a/bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveMessages.java
+++ b/bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveMessages.java
@@ -315,7 +315,7 @@ class ArchiveMessages {
                 "Commit messages:\n" +
                 formatCommitMessagesBrief(commits, commitsLink).orElse("") + "\n\n" +
                 "Changes: " + pr.changeUrl() + "\n" +
-                " Webrev: " + webrev.uri().toString() + "\n" +
+                (webrev.uri() == null ? "" : " Webrev: " + webrev.uri().toString()) + "\n" +
                 issueString +
                 "  Stats: " + stats(localRepo, base, head) + "\n" +
                 "  Patch: " + pr.diffUrl().toString() + "\n" +
@@ -328,18 +328,22 @@ class ArchiveMessages {
         var commitsLink = commitsLink(pr, base, head);
         String webrevLinks;
         if (webrevs.size() > 0) {
-            var containsConflicts = webrevs.stream().anyMatch(w -> w.type().equals(WebrevDescription.Type.MERGE_CONFLICT));
-            var containsMergeDiffs = webrevs.stream().anyMatch(w -> w.type().equals(WebrevDescription.Type.MERGE_TARGET) ||
-                    w.type().equals(WebrevDescription.Type.MERGE_SOURCE));
+            if (webrevs.stream().noneMatch(w -> w.uri() != null)) {
+                webrevLinks = "Webrev is disabled\n\n";
+            } else {
+                var containsConflicts = webrevs.stream().anyMatch(w -> w.type().equals(WebrevDescription.Type.MERGE_CONFLICT));
+                var containsMergeDiffs = webrevs.stream().anyMatch(w -> w.type().equals(WebrevDescription.Type.MERGE_TARGET) ||
+                        w.type().equals(WebrevDescription.Type.MERGE_SOURCE));
 
-            webrevLinks = "The webrev" + (webrevs.size() > 1 ? "s" : "") + " contain" + (webrevs.size() == 1 ? "s" : "") + " " +
-                    (containsConflicts ? "the conflicts with " + pr.targetRef() : "") +
-                    (containsConflicts && containsMergeDiffs ? " and " : "") +
-                    (containsMergeDiffs ? "the adjustments done while merging with regards to each parent branch" : "")
-                    +":\n" +
-                    webrevs.stream()
-                           .map(d -> String.format(" - %s: %s", d.shortLabel(), d.uri()))
-                           .collect(Collectors.joining("\n")) + "\n\n";
+                webrevLinks = "The webrev" + (webrevs.size() > 1 ? "s" : "") + " contain" + (webrevs.size() == 1 ? "s" : "") + " " +
+                        (containsConflicts ? "the conflicts with " + pr.targetRef() : "") +
+                        (containsConflicts && containsMergeDiffs ? " and " : "") +
+                        (containsMergeDiffs ? "the adjustments done while merging with regards to each parent branch" : "")
+                        + ":\n" +
+                        webrevs.stream()
+                                .map(d -> String.format(" - %s: %s", d.shortLabel(), d.uri()))
+                                .collect(Collectors.joining("\n")) + "\n\n";
+            }
         } else {
             webrevLinks = "The merge commit only contains trivial merges, so no merge-specific webrevs have been generated.\n\n";
         }
@@ -355,7 +359,7 @@ class ArchiveMessages {
 
     static String composeRebasedFooter(PullRequest pr, Repository localRepo, WebrevDescription fullWebrev, Hash base, Hash head) {
         return "Changes: " + pr.changeUrl() + "\n" +
-                " Webrev: " + fullWebrev.uri().toString() + "\n" +
+                (fullWebrev.uri() == null ? "" : " Webrev: " + fullWebrev.uri().toString() + "\n") +
                 "  Stats: " + stats(localRepo, base, head) + "\n" +
                 "  Patch: " + pr.diffUrl().toString() + "\n" +
                 "  Fetch: " + fetchCommand(pr) + "\n\n" +
@@ -366,9 +370,9 @@ class ArchiveMessages {
         return "Changes:\n" +
                 "  - all: " + pr.changeUrl() + "\n" +
                 "  - new: " + pr.changeUrl(lastHead) + "\n\n" +
-                "Webrevs:\n" +
-                " - full: " + fullWebrev.uri().toString() + "\n" +
-                " - incr: " + incrementalWebrev.uri().toString() + "\n\n" +
+                (fullWebrev.uri() == null ? "" : "Webrevs:\n") +
+                (fullWebrev.uri() == null ? "" : " - full: " + fullWebrev.uri().toString() + "\n") +
+                (incrementalWebrev.uri() == null ? "" : " - incr: " + incrementalWebrev.uri().toString() + "\n\n") +
                 "  Stats: " + stats(localRepo, lastHead, head) + "\n" +
                 "  Patch: " + pr.diffUrl().toString() + "\n" +
                 "  Fetch: " + fetchCommand(pr) + "\n\n" +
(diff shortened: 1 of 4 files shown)


zhaosongzs:
Above `ArchiveMessages#composeConversationFooter`, there is a comment says `// When changing this, ensure that the PR pattern in the notifier still matches`. I tried to figure out where the PR pattern is, but I couldn't find it.

bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveMessages.java
```diff
@@ -315,7 +315,7 @@ static String composeConversationFooter(PullRequest pr, URI issueProject, String
                 "Commit messages:\n" +
```
erikj79, line 315:
```
// When changing this, ensure that the PR pattern in the notifier still matches
```
Have you verified this?

bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveMessages.java
```diff
@@ -328,18 +328,22 @@ static String composeMergeConversationFooter(PullRequest pr, Repository localRep
...
-            webrevLinks = "The webrev" + (webrevs.size() > 1 ? "s" : "") + " contain" + (webrevs.size() == 1 ? "s" : "") + " " +
-                    (containsConflicts ? "the conflicts with " + pr.targetRef() : "") +
-                    (containsConflicts && containsMergeDiffs ? " and " : "") +
-                    (containsMergeDiffs ? "the adjustments done while merging with regards to each parent branch" : "")
-                    +":\n" +
-                    webrevs.stream()
-                           .map(d -> String.format(" - %s: %s", d.shortLabel(), d.uri()))
-                           .collect(Collectors.joining("\n")) + "\n\n";
+            if (webrevs.stream().noneMatch(w -> w.uri() != null)) {
+                webrevLinks = "Webrev is disabled\n\n";
```
erikj79, line 332: I think this should just be `""` or possibly `"\n"`, not sure.

erikj79 (commented):
This looks like it solves the stated problem, but the WebrevStorage is still initiated and will still do some work, like creating directories for webrev creation. Ideally I would like that whole part of the bot be optional, but it seems really hard to achieve without having to add null checks all over the place.

I'm not sure what the best solution is here. Maybe we can introduce another `WebrevDescription.Type.NONE` and if neither kind of webrev has been configured, then `WebrevGenerator::generate` would return a `Webrevdescription` of `NONE` type. 

Another possible solution would be to have an `enabled()` method on `WebrevStorage` and wrap all uses of `webrevGenerator` in a check for that. That may get messy though, but would definitely reduce unnecessary work when webrevs are disabled.

Have you looked at `updateWebrevComment`. When webrevs are disabled, we shouldn't post that comment either.

erikj79:
> Above `ArchiveMessages#composeConversationFooter`, there is a comment says `// When changing this, ensure that the PR pattern in the notifier still matches`. I tried to figure out where the PR pattern is, but I couldn't find it.

I believe it's the logic that parses email from the email list server and extracts the PR URL from the verify first email in the each thread. You are likely safe here as you aren't changing the line that contains the PR URL, but better be safe.

zhaosongzs:
> This looks like it solves the stated problem, but the WebrevStorage is still initiated and will still do some work, like creating directories for webrev creation. Ideally I would like that whole part of the bot be optional, but it seems really hard to achieve without having to add null checks all over the place.
> 
> I'm not sure what the best solution is here. Maybe we can introduce another `WebrevDescription.Type.NONE` and if neither kind of webrev has been configured, then `WebrevGenerator::generate` would return a `Webrevdescription` of `NONE` type.
> 
> Another possible solution would be to have an `enabled()` method on `WebrevStorage` and wrap all uses of `webrevGenerator` in a check for that. That may get messy though, but would definitely reduce unnecessary work when webrevs are disabled.
> 
> Have you looked at `updateWebrevComment`. When webrevs are disabled, we shouldn't post that comment either.

Yesterday, I was also thinking about what's the best way to solve this problem. And I thought my current change would have the least impact to current code. But your opinion is right, it still trigger some unnecessary work. I would think more about it.
And yes, I already made change in `updateWebrevComment`, it won't post comment in pr when webrev is disabled.

bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveMessages.java
```diff
@@ -315,7 +315,7 @@ static String composeConversationFooter(PullRequest pr, URI issueProject, String
                 "Commit messages:\n" +
                 formatCommitMessagesBrief(commits, commitsLink).orElse("") + "\n\n" +
                 "Changes: " + pr.changeUrl() + "\n" +
-                " Webrev: " + webrev.uri().toString() + "\n" +
+                (webrev.uri() == null ? "" : " Webrev: " + webrev.uri().toString()) + "\n" +
```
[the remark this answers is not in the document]
zhaosongzs: Sure, will fix it.

bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveMessages.java
zhaosongzs: I tried to find the related PR pattern again this morning, but unfortunately I still couldn't find it. I'm still confused about where to look for it. AFAIK, it's just the footer of the email.  Is there some place in the code where we would use this email and and parse it? I also tried resetting my head to the commit where the comment was added, but I couldn't find anything related to the notifier.

bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveMessages.java
zhaosongzs:
```
else {
            webrevLinks = "The merge commit only contains trivial merges, so no merge-specific webrevs have been generated.\n\n";
        }
```
I just tried to align with this.

zhaosongzs:
> > Above `ArchiveMessages#composeConversationFooter`, there is a comment says `// When changing this, ensure that the PR pattern in the notifier still matches`. I tried to figure out where the PR pattern is, but I couldn't find it.
> 
> I believe it's the logic that parses email from the email list server and extracts the PR URL from the verify first email in the each thread. You are likely safe here as you aren't changing the line that contains the PR URL, but better be safe.

Just see this comment. I will check it again.

bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveMessages.java
erikj79: Yes, I figured that, but there is an important difference. That message explains why there is no webrevs where the user would expect them to be, but for a repo where they are completely turned off, I think we shouldn't even mention them.

zhaosongzs:
> > Above `ArchiveMessages#composeConversationFooter`, there is a comment says `// When changing this, ensure that the PR pattern in the notifier still matches`. I tried to figure out where the PR pattern is, but I couldn't find it.
> 
> I believe it's the logic that parses email from the email list server and extracts the PR URL from the verify first email in the each thread. You are likely safe here as you aren't changing the line that contains the PR URL, but better be safe.

I couldn't find where this logic is. Could you please point it out to me?

zhaosongzs:
> > > Above `ArchiveMessages#composeConversationFooter`, there is a comment says `// When changing this, ensure that the PR pattern in the notifier still matches`. I tried to figure out where the PR pattern is, but I couldn't find it.
> > 
> > 
> > I believe it's the logic that parses email from the email list server and extracts the PR URL from the verify first email in the each thread. You are likely safe here as you aren't changing the line that contains the PR URL, but better be safe.
> 
> I couldn't find where this logic is. Could you please point it out to me?

I think I found it...

zhaosongzs:
> > Above `ArchiveMessages#composeConversationFooter`, there is a comment says `// When changing this, ensure that the PR pattern in the notifier still matches`. I tried to figure out where the PR pattern is, but I couldn't find it.
> 
> I believe it's the logic that parses email from the email list server and extracts the PR URL from the verify first email in the each thread. You are likely safe here as you aren't changing the line that contains the PR URL, but better be safe.

The logic you mentioned seems like has been deleted in SKARA-386. So I think we are safe.

erikj79:
> > I believe it's the logic that parses email from the email list server and extracts the PR URL from the verify first email in the each thread. You are likely safe here as you aren't changing the line that contains the PR URL, but better be safe.
> 
> The logic you mentioned seems like has been deleted in SKARA-386. So I think we are safe.

The logic I was thinking about is in `MailingListArchiveReaderBot` `PULL_REQUEST_LINK_PATTERN`.

zhaosongzs:
> > > I believe it's the logic that parses email from the email list server and extracts the PR URL from the verify first email in the each thread. You are likely safe here as you aren't changing the line that contains the PR URL, but better be safe.
> > 
> > 
> > The logic you mentioned seems like has been deleted in SKARA-386. So I think we are safe.
> 
> The logic I was thinking about is in `MailingListArchiveReaderBot` `PULL_REQUEST_LINK_PATTERN`.

You're right, it was not deleted, just moved.

zhaosongzs:
> This looks like it solves the stated problem, but the WebrevStorage is still initiated and will still do some work, like creating directories for webrev creation. Ideally I would like that whole part of the bot be optional, but it seems really hard to achieve without having to add null checks all over the place.
> 
> I'm not sure what the best solution is here. Maybe we can introduce another `WebrevDescription.Type.NONE` and if neither kind of webrev has been configured, then `WebrevGenerator::generate` would return a `Webrevdescription` of `NONE` type.
> 
> Another possible solution would be to have an `enabled()` method on `WebrevStorage` and wrap all uses of `webrevGenerator` in a check for that. That may get messy though, but would definitely reduce unnecessary work when webrevs are disabled.
> 
> Have you looked at `updateWebrevComment`. When webrevs are disabled, we shouldn't post that comment either.

Checked the logic again and I found although `WebrevStorage` is still initiated, it will just generate some paths but not create directories.

erikj79 (approved):

zhaosongzs:
Thanks for the review!
/integrate

PR #1476, opened 2023-02-16
1035: Race when creating backports for bug spanning multiple repos

vijayk22oracle:
SKARA-1035: Race when creating backports for bug spanning multiple repos

---------
### Progress

### Issue
 * [SKARA-1035](https://bugs.openjdk.org/browse/SKARA-1035): Race when creating backports for bug spanning multiple repos

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1476/head:pull/1476` \
`$ git checkout pull/1476`

`$ git checkout pull/1476` \
`$ git pull https://git.openjdk.org/skara pull/1476/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1476`

`$ git pr show -t 1476`

</details>

<a href="https://git.openjdk.org/skara/pull/1476.diff">https://git.openjdk.org/skara/pull/1476.diff</a>

</details>

### DIFF
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
@@ -305,7 +305,18 @@ class IssueNotifier implements Notifier, PullRequestListener, RepositoryListener
                             var existing = Backports.findIssue(issue, fixVersion);
                             if (existing.isEmpty()) {
                                 log.info("Creating new backport for " + issue.id() + " with fixVersion " + requestedVersion);
-                                issue = jbsBackport.createBackport(issue, requestedVersion, username.orElse(null), defaultSecurity(branch));
+                                try {
+                                    issue = jbsBackport.createBackport(issue, requestedVersion, username.orElse(null), defaultSecurity(branch));
+                                } catch (Exception exp) {
+                                    existing = Backports.findIssue(issue, fixVersion);
+                                    if (existing.isPresent()) {
+                                        log.info("Race condition occurred while creating the back port. So returning an existing backport for " + issue.id() + " and requested fixVersion "
+                                                + requestedVersion + " " + existing.get().id());
+                                        issue = existing.get();
+                                    } else {
+                                        throw exp;
+                                    }
+                                }
                             } else {
                                 log.info("Found existing backport for " + issue.id() + " and requested fixVersion "
                                         + requestedVersion + " " + existing.get().id());
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
@@ -31,6 +31,7 @@ import org.openjdk.skara.issuetracker.*;
 import org.openjdk.skara.jbs.*;
 import org.openjdk.skara.jcheck.JCheckConfiguration;
 import org.openjdk.skara.json.JSON;
+import org.openjdk.skara.network.UncheckedRestException;
 import org.openjdk.skara.vcs.*;
 import org.openjdk.skara.vcs.openjdk.*;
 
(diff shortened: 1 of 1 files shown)


bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
```diff
@@ -305,7 +305,18 @@ public void onNewCommits(HostedRepository repository, Repository localRepository
                             var existing = Backports.findIssue(issue, fixVersion);
                             if (existing.isEmpty()) {
                                 log.info("Creating new backport for " + issue.id() + " with fixVersion " + requestedVersion);
-                                issue = jbsBackport.createBackport(issue, requestedVersion, username.orElse(null), defaultSecurity(branch));
+                                try {
+                                    issue = jbsBackport.createBackport(issue, requestedVersion, username.orElse(null), defaultSecurity(branch));
+                                } catch (Exception exp) {
```
erikj79, line 310: Catching `Exception` is too wide a net. We need to catch `UncheckedRestException` and ideally check that the response code indicates this kind of error. Unfortunately I don't know what response that would be, so perhaps it's ok to just change the exception type for now.

Exception variables in catch statements are usually just called `e` in this codebase.

bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
```diff
@@ -305,7 +305,18 @@ public void onNewCommits(HostedRepository repository, Repository localRepository
...
+                                    existing = Backports.findIssue(issue, fixVersion);
+                                    if (existing.isPresent()) {
+                                        log.info("Race condition occurred while creating the back port. So returning an existing backport for " + issue.id() + " and requested fixVersion "
```
erikj79, line 313:
```suggestion
                                        log.info("Race condition occurred while creating backport issue, returning the existing backport for " + issue.id() + " and requested fixVersion "
```

vijayk22oracle: Sure, did the rewording as suggested, thanks.

bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
vijayk22oracle: Yes, I agree, used the more specific exception as suggested.

Regarding the specific exception code, yes I tried to run two test cases in parallel by putting the debug points to simulate the race condition but TestHost returned 2nd issue as a new one. I guess because of two instances of TestHost (one per test case). So could not verify the exception code.

zhaosongzs: If you are interested, I think you can add the status code to the log with method `UncheckedRestException#getStatusCode`.

zhaosongzs (approved):
LGTM

bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
erikj79: I think logging the statusCode as suggested by Zhao is a good idea.

erikj79: Or never mind, we already log when throwing that exception.

erikj79 (approved):

bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java
zhaosongzs: I don't think so, when the race condition occurs, the exception won't be thrown again.

erikj79: We log in `RestRequest::transformBadResponse` before throwing the exception.

zhaosongzs: Oh, right.

PR #1477, opened 2023-02-16
1825: BotTaskAggregationHandler causes OOME

erikj79:
The BotTaskAggregationHandler saves all log records in a list per thread and publishes them when a specific marker is detected in a record. This marker signals the end of a WorkItem run. If a large amount of log messages are generated by a single WorkItem, this can result in a very large amount of log messages being saved. We currently have a case where 18 million+ have been observed when processing the webrevs for https://github.com/openjdk/mobile/pull/16 in MailingListBridgeBot.

The irony here is that the only subclass of this abstract handler, the BotSlackHandler, throws away all log records that are lower level than what it's supposed to log, so we are saving all of these, only to log those of level SEVERE in the end. In the past, this class was also used for logging to logstash, and in that case, when we had one SEVERE log record, we actually wanted to log everything that lead up to it.

This fix makes it an option for the BotTaskAggregationHandler if it should save everything or just above the configured level, so a subclass can choose which behavior it expects. I also removed the filtering in BotSlackHandler as its no longer needed.

---------
### Progress

### Issue
 * [SKARA-1825](https://bugs.openjdk.org/browse/SKARA-1825): BotTaskAggregationHandler causes OOME

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1477/head:pull/1477` \
`$ git checkout pull/1477`

`$ git checkout pull/1477` \
`$ git pull https://git.openjdk.org/skara pull/1477/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1477`

`$ git pr show -t 1477`

</details>

<a href="https://git.openjdk.org/skara/pull/1477.diff">https://git.openjdk.org/skara/pull/1477.diff</a>

</details>

### DIFF
--- a/bot/src/main/java/org/openjdk/skara/bot/BotTaskAggregationHandler.java
+++ b/bot/src/main/java/org/openjdk/skara/bot/BotTaskAggregationHandler.java
@@ -46,8 +46,11 @@ public abstract class BotTaskAggregationHandler extends StreamHandler {
 
     private final Map<Long, ThreadLogs> threadLogs;
     private final Logger log;
+    // Should this class handle log level filtering or leave that to the subclass
+    private final boolean filterOnLevel;
 
-    public BotTaskAggregationHandler() {
+    public BotTaskAggregationHandler(boolean filterOnLevel) {
+        this.filterOnLevel = filterOnLevel;
         threadLogs = new ConcurrentHashMap<>();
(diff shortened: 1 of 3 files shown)


magicus (approved):
LGTM.

PR #1479, opened 2023-02-24
755: Anything starting with a "/" is treated as a command

vijayk22oracle:
755: Anything starting with a "/" is treated as a command

---------
### Progress

### Issue
 * [SKARA-755](https://bugs.openjdk.org/browse/SKARA-755): Anything starting with a "/" is treated as a command

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1479/head:pull/1479` \
`$ git checkout pull/1479`

`$ git checkout pull/1479` \
`$ git pull https://git.openjdk.org/skara pull/1479/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1479`

`$ git pr show -t 1479`

</details>

<a href="https://git.openjdk.org/skara/pull/1479.diff">https://git.openjdk.org/skara/pull/1479.diff</a>

</details>

### DIFF
--- a/bots/mlbridge/build.gradle
+++ b/bots/mlbridge/build.gradle
@@ -47,7 +47,7 @@ dependencies {
     implementation project(':version')
     implementation project(':metrics')
     implementation project(':bots:common')
-    implementation project(path: ':jbs')
+    implementation project(':jbs')
 
     testImplementation project(':test')
 }
--- a/bots/mlbridge/build.gradle
+++ b/bots/mlbridge/build.gradle
@@ -46,6 +46,8 @@ dependencies {
     implementation project(':webrev')
     implementation project(':version')
     implementation project(':metrics')
+    implementation project(':bots:common')
+    implementation project(path: ':jbs')
 
     testImplementation project(':test')
 }
--- a/bots/mlbridge/src/main/java/module-info.java
+++ b/bots/mlbridge/src/main/java/module-info.java
@@ -32,6 +32,8 @@ module org.openjdk.skara.bots.mlbridge {
     requires org.openjdk.skara.webrev;
     requires org.openjdk.skara.network;
     requires org.openjdk.skara.version;
+    requires org.openjdk.skara.jbs;
+    requires org.openjdk.skara.bots.common;
     requires java.logging;
     requires java.net.http;
 
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CommitCommandWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CommitCommandWorkItem.java
@@ -109,13 +109,8 @@ public class CommitCommandWorkItem implements WorkItem {
                 printer.print(command.name());
                 printer.println("` can only be used in pull requests.");
             }
-        } else {
-            printer.print("Unknown command `");
-            printer.print(command.name());
-            printer.println("` - for a list of valid commands use `/help`.");
+            bot.repo().addCommitComment(commitComment.commit(), writer.toString());
         }
-
-        bot.repo().addCommitComment(commitComment.commit(), writer.toString());
     }
 
     @Override
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CommitCommandWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CommitCommandWorkItem.java
@@ -28,6 +28,7 @@ import org.openjdk.skara.issuetracker.Comment;
 
 import java.io.*;
 import java.nio.file.*;
+import java.text.MessageFormat;
 import java.util.*;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -109,8 +110,14 @@ public class CommitCommandWorkItem implements WorkItem {
                 printer.print(command.name());
                 printer.println("` can only be used in pull requests.");
             }
-            bot.repo().addCommitComment(commitComment.commit(), writer.toString());
+        } else {
+            printer.printf(MessageFormat.format("""
+                    I see that you entered a comment with `/{0}` at the beginning of a line.
+                    `/{0}` is not a recognized Skara PR command, so it was ignored and treated like an ordinary comment.
+                    If you intended to issue a Skara command, use `/help` in a comment to see a complete list of Skara commands.
+                    """, command.name()));
         }
+        bot.repo().addCommitComment(commitComment.commit(), writer.toString());
     }
 
     @Override
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CommitCommandWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CommitCommandWorkItem.java
@@ -28,7 +28,6 @@ import org.openjdk.skara.issuetracker.Comment;
 
 import java.io.*;
 import java.nio.file.*;
-import java.text.MessageFormat;
 import java.util.*;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -111,12 +110,11 @@ public class CommitCommandWorkItem implements WorkItem {
                 printer.println("` can only be used in pull requests.");
             }
         } else {
-            printer.printf(MessageFormat.format("""
-                    I see that you entered a comment with `/{0}` at the beginning of a line.
-                    `/{0}` is not a recognized Skara PR command, so it was ignored and treated like an ordinary comment.
-                    If you intended to issue a Skara command, use `/help` in a comment to see a complete list of Skara commands.
-                    """, command.name()));
+            printer.print("Unknown command `");
+            printer.print(command.name());
+            printer.println("` - for a list of valid commands use `/help`.");
         }
+
         bot.repo().addCommitComment(commitComment.commit(), writer.toString());
     }
 
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
@@ -163,15 +163,10 @@ public class PullRequestCommandWorkItem extends PullRequestWorkItem {
                     printer.println("` can not be used in pull requests.");
                 }
             }
-        } else {
-            printer.print("Unknown command `");
-            printer.print(command.name());
-            printer.println("` - for a list of valid commands use `/help`.");
+            var newComment = pr.addComment(writer.toString());
+            var latency = Duration.between(command.createdAt(), newComment.createdAt());
+            log.log(Level.INFO, "Time from command '" + command.name() + "' to reply " + latency, latency);
         }
-
-        var newComment = pr.addComment(writer.toString());
-        var latency = Duration.between(command.createdAt(), newComment.createdAt());
-        log.log(Level.INFO, "Time from command '" + command.name() + "' to reply " + latency, latency);
     }
 
     @Override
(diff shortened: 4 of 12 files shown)


kevinrushforth:
This seem like too simplistic an approach to me, in that it trades one problem for another. Now a typo in a command name, such as `/intigrate` or similar, will show up as a review comment without the helpful "unrecognized command" comment and suggestion to type `/help`.

I might recommend the more targeted solution of not treating `/sometext` as a command when it appears in a code block.

As for other problematic cases, wasn't there an earlier change to not treat `/sometext` as a command unless it was at the beginning of a line (after trimming whitespace)? If not, then that would also be a good change.

I'll test that theory now: /integrate which Skara should ignore.

vijayk22oracle:
> This seem like too simplistic an approach to me, in that it trades one problem for another. Now a typo in a command name, such as `/intigrate` or similar, will show up as a review comment without the helpful "unrecognized command" comment and suggestion to type `/help`.
> 
> I might recommend the more targeted solution of not treating `/sometext` as a command when it appears in a code block.
> 
> As for other problematic cases, wasn't there an earlier change to not treat `/sometext` as a command unless it was at the beginning of a line (after trimming whitespace)? If not, then that would also be a good change.
> 
> I'll test that theory now: /integrate which Skara should ignore.

Thanks, for your feedback, yes, I agree that with this change, we won't be able to provide user feedback if there is a typo in the Skara command. But providing Skara feedback for the text mentioned only in the code block comes with assumptions that:

a) user won't type text starting with '/' outside the code block (e.g. /name-of-commenter)
b) user will always type Skara and non-Skara commands starting with '/' (e.g. windows/Linux commands) in the code block

If any of the above rules are not followed that would result in Skara bot feedback for invalid command. 

Would that be an acceptable outcome?

erikj79:
> This seem like too simplistic an approach to me, in that it trades one problem for another. Now a typo in a command name, such as `/intigrate` or similar, will show up as a review comment without the helpful "unrecognized command" comment and suggestion to type `/help`.

I think this change is good and I directed Vijay to tackle this issue. 
 
It's a difficult tradeoff for sure, but I've seen the command extractor trigger on unintentional command looking comments more often than I would like. Especially the filtering when sending emails is problematic. See for example this PR and corresponding email:

https://github.com/openjdk/jdk/pull/12695
https://mail.openjdk.org/pipermail/build-dev/2023-February/037829.html

I'm not sure why the PR bot didn't reply to that instance though.

> I might recommend the more targeted solution of not treating `/sometext` as a command when it appears in a code block.

That would be another approach, but it would require a more complicated parsing logic. I could also imagine adding some kind of fuzzy matcher on the existing commands so that obvious spelling errors were caught. I'm not sure it's worth the effort.

> As for other problematic cases, wasn't there an earlier change to not treat `/sometext` as a command unless it was at the beginning of a line (after trimming whitespace)? If not, then that would also be a good change.
> 
> I'll test that theory now: /integrate which Skara should ignore.

It's correct that we only try to match commands preceded by whitespace. I just don't think it's enough to avoid enough false positive matches.

erikj79 (approved):
The code looks good, but please hold off integrating until we have agreement.

kevinrushforth:
> I think this change is good and I directed Vijay to tackle this issue.
> 
> It's a difficult tradeoff for sure, but I've seen the command extractor trigger on unintentional command looking comments more often than I would like. Especially the filtering when sending emails is problematic. See for example this PR and corresponding email:
> 
> [openjdk/jdk#12695](https://github.com/openjdk/jdk/pull/12695) https://mail.openjdk.org/pipermail/build-dev/2023-February/037829.html
> 
> I'm not sure why the PR bot didn't reply to that instance though.
> 
> > I might recommend the more targeted solution of not treating `/sometext` as a command when it appears in a code block.
> 
> That would be another approach, but it would require a more complicated parsing logic. I could also imagine adding some kind of fuzzy matcher on the existing commands so that obvious spelling errors were caught. I'm not sure it's worth the effort.

OK, I can see your point. What do you think about also producing a comment with a warning (but not filtering the email) in case there is an unrecognized command with a `/` at the beginning of a line? Something like:

"I see that you entered a comment with `/sometext` at the beginning of a line. `/sometext` is not a recognized Skara PR command, so it was ignored and treated like an ordinary comment. If you intended to issue a Skara command, use `/help` in a comment to see a complete list of Skara commands."

erikj79:
I did a search on github `org:openjdk "for a list of valid commands use"` and from the first handful of hits, it seems about even between good and bad matches. The good matches do seem really useful. Common misspellings are "/labels" and "/intergrate". So maybe we still need some kind of suggestion posted. Another consequence of the proposed change is that any misspelled attempt will now also be emailed.

erikj79:
> See for example this PR and corresponding email:
> 
> [openjdk/jdk#12695](https://github.com/openjdk/jdk/pull/12695) https://mail.openjdk.org/pipermail/build-dev/2023-February/037829.html
> 
> I'm not sure why the PR bot didn't reply to that instance though.

I figured out why this was filtered in the email but not interpreted as a command in PR bot. The regex is different. In mlbridge, there is no matching of space between the command and the optional arguments. One way we could reduce false positives is to further restrict the regex. Stop accepting capital letters, as we only have lower case commands, and make sure the regex in the different bots match (maybe even unify them into one location in the common module).

Another pattern I saw in the search was a user using `/keepalive` in a comment to stop the bot from auto closing the PR. He probably knows it's an invalid command, but by using an invalid command, there is no email sent, which is actually a feature in this case.

vijayk22oracle:
Thanks for the above feedback, so as per the suggestions I will
    1. consolidate the patterns and move those to the common module
    2. modify the regex to verify that the command starts with a lowercase

**Question**: In case the command starts with '/' but does not match with the standard command regex, do we need to provide suggestions based on the mistyped command in the message (e.g. Do you mean 'integrate' instead of 'intergate'?) or a generic warning message suggested by Kevin well enough?

erikj79:
> Thanks for the above feedback, so as per the suggestions I will 1. consolidate the patterns and move those to the common module 2. modify the regex to verify that the command starts with a lowercase

We don't need to match any capital letters in command names as we don't have any such commands.

> **Question**: In case the command starts with '/' but does not match with the standard command regex, do we need to provide suggestions based on the mistyped command in the message (e.g. Do you mean 'integrate' instead of 'intergate'?) or a generic warning message suggested by Kevin well enough?

The generic message is enough.

magicus:
An idea might be to compare if the entered text is "roughly" equals to a Skara command. This can be measured e.g. by comparing e.g. a simplistic soundex-conversion of the command, or possibly calculating the levenshtein distance; these algorithms sound a bit scary but should be rather simple to implement, if Java does not have anything similar already.

erikj79:
> OK, I can see your point. What do you think about also producing a comment with a warning (but not filtering the email) in case there is an unrecognized command with a `/` at the beginning of a line?

I forgot to respond to this. After having reviewed how users are actually interacting with the current functionality (bot replying to "bad" commands and emails being filtered, I think we absolutely must filter and react uniformly. That means if we react to something as being a command, then we should also filter the email, and vice versa.

erikj79:
> An idea might be to compare if the entered text is "roughly" equals to a Skara command. This can be measured e.g. by comparing e.g. a simplistic soundex-conversion of the command, or possibly calculating the levenshtein distance; these algorithms sound a bit scary but should be rather simple to implement, if Java does not have anything similar already.

Yes, that kind of matching would be even better. Looking into such a solution could be an interesting followup to this fix. We currently don't use any external dependencies for Skara, and breaking that tradition shouldn't be done lightly.

vijayk22oracle:
Implemented suggested changes to 
1. unify the command identification check for the bot and archive
2. tightened the regex pattern to identify Skara-like commands more accurately. (lowercase and separated by white space).

bots/mlbridge/build.gradle
```diff
@@ -46,6 +46,8 @@ dependencies {
     implementation project(':webrev')
     implementation project(':version')
     implementation project(':metrics')
+    implementation project(':bots:common')
+    implementation project(path: ':jbs')
```
erikj79, line 50: What does the `path:` prefix mean here? If it's not significant, better stay with the same pattern as the other dependencies.

bots/mlbridge/src/main/java/module-info.java
```diff
@@ -32,6 +32,8 @@
     requires org.openjdk.skara.webrev;
     requires org.openjdk.skara.network;
     requires org.openjdk.skara.version;
+    requires org.openjdk.skara.jbs;
```
erikj79, line 35: Do we really require jbs now?

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CommitCommandWorkItem.java
```diff
@@ -110,11 +111,12 @@ private void processCommand(Path scratchPath, HostedCommit commit, LimitedCensus
...
             }
         } else {
-            printer.print("Unknown command `");
-            printer.print(command.name());
-            printer.println("` - for a list of valid commands use `/help`.");
+            printer.printf(MessageFormat.format("""
+                    I see that you entered a comment with `/{0}` at the beginning of a line.
+                    `/{0}` is not a recognized Skara PR command, so it was ignored and treated like an ordinary comment.
+                    If you intended to issue a Skara command, use `/help` in a comment to see a complete list of Skara commands.
+                    """, command.name()));
```
erikj79, line 118: Since we are keeping the mail filtering rules the same as command execution, I don't think we should change this message.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
```diff
@@ -164,11 +165,12 @@ private void processCommand(PullRequest pr, CensusInstance censusInstance, Path
...
             }
         } else {
-            printer.print("Unknown command `");
-            printer.print(command.name());
-            printer.println("` - for a list of valid commands use `/help`.");
+            printer.printf(MessageFormat.format("""
+                    I see that you entered a comment with `/{0}` at the beginning of a line.
+                    `/{0}` is not a recognized Skara PR command, so it was ignored and treated like an ordinary comment
+                    If you intended to issue a Skara command, use `/help` in a comment to see a complete list of Skara commands.
+                    """, command.name()));
```
erikj79, line 172: Same here, keep the message as is.

vijayk22oracle: Sure, I reverted this change.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CommitCommandWorkItem.java
vijayk22oracle: Sure, I reverted this change.

bots/mlbridge/src/main/java/module-info.java
vijayk22oracle: I tried to remove it but got an error while running the test cases. I tried to add jbs as a test gradle dependency but it still complained during compiling code before running the tests.
So I guess, it is needed along with the common module. Hence I did not make any changes.

Here is the error message if removed.
`> Task :bots:mlbridge:compileJava FAILED
error: module not found: org.openjdk.skara.jbs
1 error`

bots/mlbridge/build.gradle
vijayk22oracle: Yes sure, I removed it to be consistent.

erikj79 (approved):

PR #1481, opened 2023-03-13
1839: SKARA-1480 broke BridgeBotTests

erikj79:
When changing all 'openjdk.java.net' references to 'openjdk.org', the BridgeBotTests broke, because we now have a different user mapping in those tests, which results in different Git hashes.

These tests have an assumeTrue condition in the @BeforeClass setup, which will skip if the HG server isn't available. The URL to hg.openjdk.org uses 'http', which hasn't worked since the migration to the new URL, but some time during the last few days, it started responding to 'http', which caused the tests to start running again, and subsequently fail.

This patch updates the Git hash to the new one being produced with the new user mapping, and changes the protocol to https for hg.openjdk.org, since that is the preferred one.

---------
### Progress

### Issue
 * [SKARA-1839](https://bugs.openjdk.org/browse/SKARA-1839): SKARA-1480 broke BridgeBotTests

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara pull/1481/head:pull/1481` \
`$ git checkout pull/1481`

`$ git checkout pull/1481` \
`$ git pull https://git.openjdk.org/skara pull/1481/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1481`

`$ git pr show -t 1481`

</details>

<a href="https://git.openjdk.org/skara/pull/1481.diff">https://git.openjdk.org/skara/pull/1481.diff</a>

</details>

### DIFF
--- a/bots/hgbridge/src/test/java/org/openjdk/skara/bots/hgbridge/BridgeBotTests.java
+++ b/bots/hgbridge/src/test/java/org/openjdk/skara/bots/hgbridge/BridgeBotTests.java
@@ -46,6 +46,9 @@ import static org.junit.jupiter.api.Assertions.*;
 
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 class BridgeBotTests {
+
+    private static final String FIRST_CONVERTED_HASH = "35b30f47db61035c479b18e0940fb6fdabc3944f";
+
     private List<String> runHgCommand(Repository repository, String... params) throws IOException {
         List<String> finalParams = new ArrayList<>();
         finalParams.add("hg");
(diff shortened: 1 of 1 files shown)


zhaosongzs:
This change looks good to me. But I don't know the meaning of 'user mapping' in the tests. Could you please explain it to me?

erikj79:
> This change looks good to me. But I don't know the meaning of 'user mapping' in the tests. Could you please explain it to me?

When converting from Mercurial to Git, we are also converting the format of users. In Mercurial, the user field in a changeset is generally on the form of a userId (erikj in my case). In Git the tradition is to use full name and email in every commit. The user mappings are used to help the conversion process translate from Mercurial users to Git users. In the test, there is a hardcoded mapping of two users (jjg and duke) which are the only ones represented in the changes being converted. Before SKARA-1480, the email addresses in these mappings where `*openjdk.java.net`, but in SKARA-1480 I changed them to `*openjdk.org`. This is causing the resulting Git hash to be different after the conversion.

zhaosongzs:
> > This change looks good to me. But I don't know the meaning of 'user mapping' in the tests. Could you please explain it to me?
> 
> When converting from Mercurial to Git, we are also converting the format of users. In Mercurial, the user field in a changeset is generally on the form of a userId (erikj in my case). In Git the tradition is to use full name and email in every commit. The user mappings are used to help the conversion process translate from Mercurial users to Git users. In the test, there is a hardcoded mapping of two users (jjg and duke) which are the only ones represented in the changes being converted. Before SKARA-1480, the email addresses in these mappings where `*openjdk.java.net`, but in SKARA-1480 I changed them to `*openjdk.org`. This is causing the resulting Git hash to be different after the conversion.

Thank you so much for this detailed explanation!

kevinrushforth (approved):

PR #1483, opened 2023-03-14
SKARA-1840

zhaosongzs:
This patch fixes MailingListBridgeBotTests::simpleArchive.


zhaosongzs:
> This patch looks good.
> 
> But when reading the code of this test case `MailingListBridgeBotTests::simpleArchive`, I notice the following code is wrong. ` pr.addLabel("rfr");` should be adjusted to ` pr.removeLabel("rfr");`. It is good to fix this little bug in this patch. What do you think?
> 
> ```
> // the line 279:
>             // Remove the rfr flag and post another comment
>             pr.addLabel("rfr");
>             pr.addComment("@" + pr.author().username() +" This is another comment");
> ```

Good catch! I think fix it in this patch is good.

lgxbslgx (approved):
Looks good.

zhaosongzs:
Thanks for the review!
/integrate

PR #1486, opened 2023-03-16
SKARA-1838

zhaosongzs:
Sometimes, users may want to run jcheck locally before creating PRs. However, Oracle employees who are connected to the Oracle VPN and have not set the https_proxy environment variable may encounter a "connect exception" error. 
In this patch, we will provide users with more information in the case of a "connect exception" error while running jcheck.

### DIFF
--- a/cli/src/main/java/org/openjdk/skara/cli/GitJCheck.java
+++ b/cli/src/main/java/org/openjdk/skara/cli/GitJCheck.java
@@ -349,6 +349,9 @@ public class GitJCheck {
                 }
             } catch (Exception e) {
                 System.err.println(String.format("error: exception thrown during jcheck: %s", e.getMessage()));
+                if (e.getMessage().equals("java.net.ConnectException")) {
+                    System.err.println("If you are connected to Oracle's VPN, please set the https_proxy environment variable and try again");
+                }
                 return 1;
             }
         }


cli/src/main/java/org/openjdk/skara/cli/GitJCheck.java
```diff
@@ -349,6 +349,9 @@ else if (confFile || confWorkingTree || workingTree) {
                 }
             } catch (Exception e) {
                 System.err.println(String.format("error: exception thrown during jcheck: %s", e.getMessage()));
+                if (e.getMessage().equals("java.net.ConnectException")) {
+                    System.err.println("If you are connected to Oracle's VPN, please set the https_proxy environment variable and try again");
```
erikj79, line 353:
```suggestion
                    System.err.println("If you are behind a firewall without direct access to the internet, make sure to configure any required proxy server through the https_proxy environment variable and try again");
```

cli/src/main/java/org/openjdk/skara/cli/GitJCheck.java
```diff
@@ -349,6 +349,9 @@ else if (confFile || confWorkingTree || workingTree) {
                 }
             } catch (Exception e) {
                 System.err.println(String.format("error: exception thrown during jcheck: %s", e.getMessage()));
+                if (e.getMessage().equals("java.net.ConnectException")) {
```
erikj79, line 352: I forgot to look at the rest of the code. I think this should be an `instanceof` check instead of comparing strings.

zhaosongzs: Sure

erikj79 (approved):

PR #1489, opened 2023-03-20
SKARA-1830

zhaosongzs:
In this patch, notifyBot would add wildcard branch protection for the repo, and when the bot tries to delete the pr/* branch, it will temporarily remove wildcard branch protection.

### DIFF
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/NotifyBot.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/NotifyBot.java
@@ -83,12 +83,14 @@ public class NotifyBot implements Bot, Emitter {
     @Override
     public List<WorkItem> getPeriodicItems() {
         var ret = new ArrayList<WorkItem>();
+
         if (!wildCardProtected && prListeners.stream()
                 .anyMatch(e -> e instanceof PullRequestBranchNotifier && ((PullRequestBranchNotifier) e).isProtectBranches())) {
             log.info("Protecting branch *");
             repository.protectBranchPattern("*");
             wildCardProtected = true;
         }
+
         if (!prListeners.isEmpty()) {
             // Pull request events
             List<PullRequest> prs = poller.updatedPullRequests();
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/NotifyBot.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/NotifyBot.java
@@ -28,7 +28,6 @@ import org.openjdk.skara.forge.*;
 import org.openjdk.skara.storage.StorageBuilder;
 
 import java.nio.file.Path;
-import java.time.*;
 import java.util.*;
 import java.util.logging.Logger;
 import java.util.regex.Pattern;
--- a/bots/notify/src/main/java/org/openjdk/skara/bots/notify/NotifyBot.java
+++ b/bots/notify/src/main/java/org/openjdk/skara/bots/notify/NotifyBot.java
@@ -23,6 +23,7 @@
 package org.openjdk.skara.bots.notify;
 
 import org.openjdk.skara.bot.*;
+import org.openjdk.skara.bots.notify.prbranch.PullRequestBranchNotifier;
 import org.openjdk.skara.forge.*;
 import org.openjdk.skara.storage.StorageBuilder;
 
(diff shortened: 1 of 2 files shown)


bots/notify/src/main/java/org/openjdk/skara/bots/notify/prbranch/PullRequestBranchNotifier.java
```diff
@@ -69,13 +65,17 @@ private void deleteBranch(PullRequest pr) {
         if (protectBranches) {
             log.info("Removing branch protection for " + branch);
             pr.repository().unprotectBranchPattern(branch);
```
zhaosongzs, line 67: We still need this because it's possible that we have some pr branch protected before we deploy this patch.

bots/notify/src/main/java/org/openjdk/skara/bots/notify/NotifyBot.java
```diff
@@ -83,6 +84,13 @@ public String toString() {
     public List<WorkItem> getPeriodicItems() {
         var ret = new ArrayList<WorkItem>();
 
+        if (!wildCardProtected && prListeners.stream()
+                .anyMatch(e -> e instanceof PullRequestBranchNotifier && ((PullRequestBranchNotifier) e).isProtectBranches())) {
+            log.info("Protecting branch *");
+            repository.protectBranchPattern("*");
+            wildCardProtected = true;
+        }
```
erikj79, line 92: I would suggest adding a new method on the `PullRequestListener` interface `initialize(HostedRepository)` and implement this method in `PullRequestBranchNotifier` so you can put this logic there instead. Then you can call `initialize` on all listeners when `getPeriodicItems` is called the first time.

bots/notify/src/main/java/org/openjdk/skara/bots/notify/prbranch/PullRequestBranchNotifier.java
```diff
@@ -69,13 +65,17 @@ private void deleteBranch(PullRequest pr) {
...
+            pr.repository().unprotectBranchPattern("*");
         }
         if (!branchExists) {
             log.info("Pull request pre-integration branch " + branch + " doesn't exist on remote - ignoring");
             return;
         }
         log.info("Deleting pull request pre-integration branch " + branch);
         pr.repository().deleteBranch(branch);
+        log.info("Protecting branch * after deleting branch " + branch);
+        pr.repository().protectBranchPattern("*");
```
erikj79, line 78: We should not be touching branch protection settings unless `protectBranches` is enabled.

bots/notify/src/main/java/org/openjdk/skara/bots/notify/prbranch/PullRequestBranchNotifier.java
erikj79: You should add a code comment explaining this so we may eventually remove that code later.

Have you checked how the call behaves if there is no protection to remove?

bots/notify/src/main/java/org/openjdk/skara/bots/notify/NotifyBot.java
zhaosongzs: Good idea. Will do it.

bots/notify/src/main/java/org/openjdk/skara/bots/notify/prbranch/PullRequestBranchNotifier.java
zhaosongzs: Exactly, will fix it.

bots/notify/src/main/java/org/openjdk/skara/bots/notify/prbranch/PullRequestBranchNotifier.java
zhaosongzs: Sure, will add the comment. No, I didn't check it by myself because I saw your test in `GitLabRestApiTest::branchProtection`. I will check it again later.

zhaosongzs: Checked it again, if there is no protection to remove, nothing would happen. The bot works fine.

erikj79 (approved):

zhaosongzs:
Thanks for the review!
/integrate

PR #1493, opened 2023-03-28
1859: Add "tags only" option to MirrorBot

edvbld:
Hi all,

please review this patch that adds an option to the `MirrorBot` to mirror only tags from one repository to another. I also added some additional logging to `MirrorBot` and cleaned up the logic a bit for selecting the "mode" (tags, mirror, selected branches) that the `MirrorBot` uses.

### Testing
- [x] Added one additional unit test

Thanks,
Erik

---------
### Progress

### Issue
 * [SKARA-1859](https://bugs.openjdk.org/browse/SKARA-1859): Add "tags only" option to MirrorBot

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1493/head:pull/1493` \
`$ git checkout pull/1493`

`$ git checkout pull/1493` \
`$ git pull https://git.openjdk.org/skara.git pull/1493/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1493`

`$ git pr show -t 1493`

</details>

<a href="https://git.openjdk.org/skara/pull/1493.diff">https://git.openjdk.org/skara/pull/1493.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1493#issuecomment-1486533385)

### DIFF
--- a/bots/mirror/src/main/java/org/openjdk/skara/bots/mirror/MirrorBotFactory.java
+++ b/bots/mirror/src/main/java/org/openjdk/skara/bots/mirror/MirrorBotFactory.java
@@ -96,9 +96,6 @@ public class MirrorBotFactory implements BotFactory {
             if (onlyTags && !branchPatterns.isEmpty()) {
                 throw new IllegalStateException("Branches cannot be mirrored when only tags are mirrored");
             }
-            if (includeTags && !branchPatterns.isEmpty()) {
-                throw new IllegalStateException("Cannot include tags with only selected branches (tags are *not* per branch)");
-            }
 
             log.info("Setting up mirroring from " + fromRepo.name() + "to " + toRepo.name());
             bots.add(new MirrorBot(storage, fromRepo, toRepo, branchPatterns, includeTags, onlyTags));
--- a/bots/mirror/src/main/java/org/openjdk/skara/bots/mirror/MirrorBotFactory.java
+++ b/bots/mirror/src/main/java/org/openjdk/skara/bots/mirror/MirrorBotFactory.java
@@ -83,8 +83,8 @@ public class MirrorBotFactory implements BotFactory {
             var onlyTags = false;
             if (repo.contains("tags")) {
                 var tags = repo.get("tags").asString().toLowerCase().strip();
-                if (!Set.of("include", "only", "exclude").contains(tags)) {
-                    throw new IllegalStateException("\"tags\" field can only have values \"include\", \"only\" or \"exclude\"");
+                if (!Set.of("include", "only").contains(tags)) {
+                    throw new IllegalStateException("\"tags\" field can only have value \"include\" or \"only\"");
                 }
                 onlyTags = tags.equals("only");
                 includeTags = tags.equals("include");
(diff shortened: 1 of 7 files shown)


erikj79 (commented):
Could you expand the `MirrorBotFactory` test to verify the new configuration options? It's also worth pointing out that existing configurations will need to be updated when this change goes live.

bots/mirror/src/main/java/org/openjdk/skara/bots/mirror/MirrorBotFactory.java
```diff
@@ -79,10 +79,23 @@ public List<Bot> create(BotConfiguration configuration) {
...
+                              repo.get("tags").get("include").asBoolean();
+                onlyTags = repo.get("tags").contains("only") &&
+                           repo.get("tags").get("only").asBoolean();
+            }
+            if (onlyTags && !includeTags) {
+                throw new IllegalStateException("Must include tags if only tags are mirrored");
+            }
+            if (onlyTags && !branchPatterns.isEmpty()) {
+                throw new IllegalStateException("Branches cannot be mirrored when only tags are mirrored");
+            }
```
zhaosongzs, line 95: Originally, if `branchPatterns.isEmpty()`, `includeTags` would always be true. But now, it is possible that when `branchPatterns.isEmpty()`, `includeTags` is false.
So I think we could also throw exception when `branchPatterns.isEmpty() && !includeTags`. It is illegal.

edvbld: Thanks @zhaosongzs , I reworked this a bit in the latest version and fixed this as well

edvbld:
@erikj79 thans, I added a bunch of tests in `MirrorBotFactoryTest.java`. As I wrote the tests I realized it was more natural to use a string value for the `"tags"` field, see the tests in the latest version for examples.

@zhaosongzs I think the latest version also resolves your comment, please have a look.

bots/mirror/src/test/java/org/openjdk/skara/bots/mirror/MirrorBotFactoryTest.java
```diff
@@ -74,24 +76,203 @@ public void testCreate() {
...
+                        {
+                          "from": "from3",
+                          "to": "to3",
+                          "tags": "only"
+                        },
+                        {
+                          "from": "from4",
+                          "to": "to4",
+                          "tags": "exclude"
+                        },
```
zhaosongzs, line 234: I am wondering what's the expected behavior for the mirror bot configured like this? Intuitively speaking, I think the bot should mirror every branch but excluding tags. But currently, the bot would not mirror anything. if i am wrong please correct me. Thank you!

erikj79: I believe you are correct, this config will mirror nothing, but I agree that the expectation would be to mirror all branches but no tags.

edvbld: Yep, realized this after I pushed. I decided to remove `"exclude"` in the latest commit since implementing it is quite cumbersome (the `git` CLI client doesn't have a `--no-tags` option for `git push`).

erikj79 (approved):

zhaosongzs:
> @zhaosongzs @erikj79 I decided to remove `"exclude"` (see inline comment for details). The latest version should now be fully backwards compatible and only features `"include"` and `"only"` for tags.

LGTM! Thank you!

erikj79 (approved):

edvbld:
Thanks @erikj79!

/integrate

PR #1500, opened 2023-04-11
SKARA-1854

zhaosongzs:
In this patch, PR bot would add clean label to Merge-style PRs if one of the following conditions is met.

1. There isn't a merge commit at HEAD. 
To verify this condition, we can simply check the number of parents that the HEAD commit has. If the HEAD commit has one parent, it is not a merge commit. However, if it has two parents, then it is considered a merge commit.

2. The merge commit at HEAD is empty(has no merge resolutions)
To verify this condition, we should be able to tell the difference between an empty merge commit and a non-empty merge commit.

Using `git show <Hash>` would help us differentiate between this two types of commits.

(1) git show output for empty merge commit
```
commit 9752b806edafbdc210db00f3099404cd8bd11edf
Merge: 982b5ad5d8b 99f8d05da93
Author: Zhao Song <zhao.song@oracle.com>
Date:   Tue Apr 11 14:26:18 2023 -0700

    Merge branch 'feature'

```

(2) git show output for non-empty merge commit
```
commit 242a0c07cf3204c662a2f1d55a2ca3e494f955fd (HEAD -> master)
Merge: af35791d093 0ebf3bd5fe1
Author: Zhao Song <zhao.song@oracle.com>
Date:   Tue Apr 11 14:38:05 2023 -0700

    Merge branch 'feature2'

diff --cc test.java
index 7c736ab8f08,09647d12eda..a3c5725e39d
--- a/test.java
+++ b/test.java
@@@ -3,4 -3,4 +3,5 @@@
  aaaaa
  +++++
  11111
 +33333
+ 22222

```

We could find that the difference between the two is whether the commit message body is empty or not.

Since we only care about commit message body, we can use the `git show --pretty=format:%b <Hash>` command to extract solely the commit message body.

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -443,6 +443,23 @@ class CheckRun {
         return isClean || isCleanLabelManuallyAdded;
     }
 
+    private void updateMergeClean(Commit commit) {
+        boolean isClean = !commit.isMerge();
+        if (!isClean) {
+            var commitMessageBody = localRepo.commitMessageBody(commit.hash());
+            if (commitMessageBody.isPresent()) {
+                if (commitMessageBody.get().equals("")) {
+                    isClean = true;
+                }
+            }
+        }
+        if (isClean) {
+            newLabels.add("clean");
+        } else {
+            newLabels.remove("clean");
+        }
+    }
+
     private Optional<HostedCommit> backportedFrom() {
         var hash = checkablePullRequest.findOriginalBackportHash();
         if (hash == null) {
@@ -1125,6 +1142,10 @@ class CheckRun {
                 commitHash = mergedHash.get();
             } else {
                 rebasePossible = false;
+            }
+
+            if (rebasePossible && PullRequestUtils.isMerge(pr)) {
+                localRepo.lookup(pr.headHash()).ifPresent(this::updateMergeClean);
             }
 
             var original = backportedFrom();
--- a/vcs/src/main/java/org/openjdk/skara/vcs/Repository.java
+++ b/vcs/src/main/java/org/openjdk/skara/vcs/Repository.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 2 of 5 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -443,6 +443,28 @@ private boolean updateClean(Commit commit) {
...
+            var commitMessageBody = localRepo.commitMessageBody(commit.hash());
+            if (commitMessageBody.isPresent()) {
+                var lines = commitMessageBody.get();
+                for (int i = 0; i < lines.size() - 1; i++) {
+                    if (lines.get(i).startsWith("diff") && lines.get(i + 1).startsWith("index")) {
+                        isClean = false;
+                        break;
+                    }
+                }
+            }
```
erikj79, line 459: I think this is doing the right thing, but the abstraction isn't right. If we were to ever implement this for HG, then we can't know what the output would look like here, and we can't assume it will be the same as for Git. The knowledge about how to interpret the output of Git needs to be encapsulated in `GitRepository`. Most of this method should move into a new method in `Repository`/`GitRepository`.

Thinking some more about this, don't we have enough information in the `Commit` instance to figure this out? If the commit doesn't have any patches, then we should be good to go right? If so, we don't need any new method on `Repository`.

vcs/src/main/java/org/openjdk/skara/vcs/Repository.java
```diff
@@ -183,6 +183,8 @@ default void updateSubmodule(Submodule s) throws IOException {
         updateSubmodule(s.path());
     }
 
+    boolean isEmptyCommit(Hash hash);
```
erikj79, line 186: We should add a javadoc comment describing the expected behavior for this method, specifically for merge commits.

vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
```diff
@@ -1727,4 +1727,21 @@ private Diff diffStaged() throws IOException {
             throw t;
         }
     }
+
+    @Override
+    public boolean isEmptyCommit(Hash hash) {
+        try (var p = capture("git", "show", "--pretty=format:%b", hash.hex())) {
```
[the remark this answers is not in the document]
zhaosongzs: Yes, it is a good idea. Will fix it.

vcs/src/main/java/org/openjdk/skara/vcs/Repository.java
zhaosongzs: Sure. But I think it also works for simple empty commit(not merge commit)

zhaosongzs: At first, I name this method as `isEmptyMergeCommit` and later I thought it also works for simple empty commit, so I changed the method name to `isEmptyCommit`

zhaosongzs: NVM, I misunderstood your comment

erikj79 (approved):

zhaosongzs:
Thanks for the review!
/integrate

PR #1510, opened 2023-04-26
1860: Make it possible to configure PR bot to jcheck all commits in "Merge PR"

zhaosongzs:
In this patch, PR bot has been added with the capability to control whether jcheck every commit in a merge PR.

By default, jcheckMerge is enabled.

However, if it is desired to enable jcheckMerge in a repo, the configuration can be added to the PR bot configuration as follows:
```
{
  "pr": {
      "repositories": {
          "repo1": {
               "jcheckMerge": true
          }
       }
  }
}
```
If jcheckMerge is true, then the pr bot would jcheck all the commits in the merge PR and if any jcheck doesn't pass, the error message will be treated as Integration blocker.

---------
### Progress

### Issue
 * [SKARA-1860](https://bugs.openjdk.org/browse/SKARA-1860): Make it possible to configure PR bot to jcheck all commits in "Merge PR"

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1510/head:pull/1510` \
`$ git checkout pull/1510`

`$ git checkout pull/1510` \
`$ git pull https://git.openjdk.org/skara.git pull/1510/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1510`

`$ git pr show -t 1510`

</details>

<a href="https://git.openjdk.org/skara/pull/1510.diff">https://git.openjdk.org/skara/pull/1510.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1510#issuecomment-1524007706)

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -1136,8 +1136,36 @@ class CheckRun {
                 rebasePossible = false;
             }
 
-            if (rebasePossible && PullRequestUtils.isMerge(pr)) {
-                localRepo.lookup(pr.headHash()).ifPresent(this::updateMergeClean);
+            List<String> mergeJCheckMessage = new ArrayList<>();
+
+            if (PullRequestUtils.isMerge(pr)) {
+                if (rebasePossible) {
+                    localRepo.lookup(pr.headHash()).ifPresent(this::updateMergeClean);
+                }
+
+                // JCheck all commits in "Merge PR"
+                if (workItem.bot.jcheckMerge()) {
+                    var commits = localRepo.commitMetadata(localRepo.mergeBase(PullRequestUtils.targetHash(localRepo), pr.headHash()), commitHash);
+                    var commitHashes = commits.stream()
+                            .map(CommitMetadata::hash)
+                            .collect(Collectors.toSet());
+                    commitHashes.remove(pr.headHash());
+                    for (Hash hash : commitHashes) {
+                        try {
+                            PullRequestCheckIssueVisitor visitor = checkablePullRequest.createVisitor(hash);
+                            checkablePullRequest.executeChecks(hash, censusInstance, visitor, List.of(), hash);
+                            mergeJCheckMessage.addAll(visitor.messages().stream()
+                                    .map(StringBuilder::new)
+                                    .map(e -> e.append(" (failed when running jcheck with commit " + hash.hex() + ")"))
+                                    .map(StringBuilder::toString)
+                                    .toList());
+                        } catch (Exception e) {
+                            var message = e.getMessage() + " (exception thrown when running jcheck with commit " + hash.hex() + ")";
+                            log.warning(message);
+                            mergeJCheckMessage.add(message);
+                        }
+                    }
+                }
             }
 
             var original = backportedFrom();
@@ -1158,7 +1186,7 @@ class CheckRun {
                 additionalErrors = List.of(e.getMessage());
                 localHash = baseHash;
             }
-            PullRequestCheckIssueVisitor visitor = checkablePullRequest.createVisitor();
+            PullRequestCheckIssueVisitor visitor = checkablePullRequest.createVisitor(checkablePullRequest.targetHash());
             boolean needUpdateAdditionalProgresses = false;
             boolean sourceBranchJCheckConfValid = true;
             if (localHash.equals(baseHash)) {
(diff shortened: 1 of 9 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -1136,8 +1136,36 @@ private void checkStatus() {
...
+                    for (Hash hash : commitHashes) {
+                        try {
+                            PullRequestCheckIssueVisitor visitor = checkablePullRequest.createVisitor(hash);
+                            checkablePullRequest.executeChecks(hash, censusInstance, visitor, List.of(), hash);
+                            mergeJCheckMessage.addAll(visitor.messages().stream()
+                                    .map(StringBuilder::new)
+                                    .map(e -> e.append(" (failed when running jcheck with commit " + hash.hex() + ")"))
+                                    .map(StringBuilder::toString)
+                                    .toList());
+                        } catch (Exception e) {
```
erikj79, line 1162: I think instead of catching Exception here, let the outer try-catch take it as that will trigger a retry and a suitable message to be printed.

bots/pr/src/test/java/org/openjdk/skara/bots/pr/MergeTests.java
```diff
@@ -1729,4 +1729,66 @@ void mergeSourceInvalid(TestInfo testInfo) throws IOException {
...
+            TestBotRunner.runPeriodicItems(mergeBot);
+
+            // There is a merge commit at HEAD, but the merge commit is empty
+            assertTrue(pr.store().labelNames().contains("clean"));
+
+            // Push it
+            pr.addComment("/integrate");
+            TestBotRunner.runPeriodicItems(mergeBot);
+
+            // The bot should reply with an ok message
```
erikj79, line 1784: This looks like the opposite of "ok".  😄

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -1136,8 +1136,36 @@ private void checkStatus() {
...
+                            .map(CommitMetadata::hash)
+                            .collect(Collectors.toSet());
+                    commitHashes.remove(pr.headHash());
+                    for (Hash hash : commitHashes) {
+                        try {
+                            PullRequestCheckIssueVisitor visitor = checkablePullRequest.createVisitor(hash);
+                            checkablePullRequest.executeChecks(hash, censusInstance, visitor, List.of(), hash);
+                            mergeJCheckMessage.addAll(visitor.messages().stream()
+                                    .map(StringBuilder::new)
+                                    .map(e -> e.append(" (failed when running jcheck with commit " + hash.hex() + ")"))
```
erikj79, line 1159:
```suggestion
                                    .map(e -> e.append(" (in commit " + hash.hex() + ")"))
```

bots/pr/src/test/java/org/openjdk/skara/bots/pr/MergeTests.java
zhaosongzs: Will fix it!

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
zhaosongzs: Ah, right. And I will also remove the try catch when running "second jcheck".

erikj79: Good, do that please. 

For the second jcheck, there may be a usecase for printing the exception, because the user has made edits to .jcheck/conf. On the other hand, I don't think we should ever print `e.getMessage()` on a caught `Exception` instance to the user unless we know exactly what would have thrown it. It's enough that it gets logged for now. If jcheck throws exceptions that the user needs to see, then we need to fix jcheck to present that as a valid failure instead.

bots/pr/src/test/java/org/openjdk/skara/bots/pr/CheckTests.java
```diff
@@ -1350,7 +1350,7 @@ void excessiveFailures(TestInfo testInfo) throws IOException {
     }
 
     @Test
-    void useJCheckConfFromTargetBranch(TestInfo testInfo) throws IOException {
+    void invalidUpdatedJCheckConf(TestInfo testInfo) throws IOException {
```
zhaosongzs, line 1353: Now, this test duplicates `testRunJcheckTwiceWithBadConfiguration`, but I don't think it would hurt to keep it.

erikj79 (approved):

zhaosongzs:
Thanks for the review!
/integrate

PR #1519, opened 2023-05-11
SKARA-1909

erikj79:
Over the last week, we have had issues with JBS returning weird errors seemingly randomly, which has caused a lot of problems with the bots. It turns out our hosting provider has become more stringent with handling http/2 requests and no longer likes GET requests with a request body. Skara currently always fills in the body of a request, regardless of type, with an empty JSON document '{}' when the body is empty. This seems to be faulty behavior and we should let empty GET calls be empty instead.

I have been running staging with this patch over night and there were no errors.


kevinrushforth (approved):
Looks good.

tbell29552 (approved):

PR #1521, opened 2023-05-12
1910: Basic support for Bitbucket as a Forge

erikj79:
This patch adds very basic support for Bitbucket as a Forge, just enough to be able to use a Bitbucket repo as the source for mirroring.

In order to support ssh for bitbucket, I moved the `configureSshKey` method from `GitLabForgeFactory` to a new utility class, so that it could be shared more easily.

---------
### Progress

### Issue
 * [SKARA-1910](https://bugs.openjdk.org/browse/SKARA-1910): Basic support for Bitbucket as a Forge

### Reviewers
 * [Zhao Song](https://openjdk.org/census#zsong) (@zhaosongzs - Committer) ⚠️ Review applies to [0a81c6f6](https://git.openjdk.org/skara/pull/1521/files/0a81c6f6537e60707c38e9663393341836170433)
 * [Erik Helin](https://openjdk.org/census#ehelin) (@edvbld - **Reviewer**) ⚠️ Review applies to [0a81c6f6](https://git.openjdk.org/skara/pull/1521/files/0a81c6f6537e60707c38e9663393341836170433)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1521/head:pull/1521` \
`$ git checkout pull/1521`

`$ git checkout pull/1521` \
`$ git pull https://git.openjdk.org/skara.git pull/1521/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1521`

`$ git pr show -t 1521`

</details>

<a href="https://git.openjdk.org/skara/pull/1521.diff">https://git.openjdk.org/skara/pull/1521.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1521#issuecomment-1546242996)

### DIFF
--- a/forge/src/main/java/org/openjdk/skara/forge/bitbucket/BitbucketForgeFactory.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/bitbucket/BitbucketForgeFactory.java
@@ -31,6 +31,7 @@ import org.openjdk.skara.host.Credential;
 import org.openjdk.skara.json.JSONObject;
 
 public class BitbucketForgeFactory implements ForgeFactory {
+
     @Override
     public String name() {
         return "bitbucket";
@@ -48,18 +49,17 @@ public class BitbucketForgeFactory implements ForgeFactory {
             name = configuration.get("name").asString();
         }
         var useSsh = false;
-        if (configuration != null && configuration.contains("sshkey") && credential != null) {
+        if (configuration != null && configuration.contains("sshkey")) {
+            if (credential == null) {
+                throw new RuntimeException("Cannot use SSH without credentials");
+            }
             ForgeUtils.configureSshKey(credential.username(), uri.getHost(), configuration.get("sshkey").asString());
             useSsh = true;
         }
-        int sshport = 22;
+        int sshport = BitbucketHost.DEFAULT_SSH_PORT;
         if (configuration != null && configuration.contains("sshport")) {
             sshport = configuration.get("sshport").asInt();
         }
-        if (credential != null) {
-            return new BitbucketHost(name, uri, useSsh, sshport, credential);
-        } else {
-            return new BitbucketHost(name, uri, useSsh, sshport);
-        }
+        return new BitbucketHost(name, uri, useSsh, sshport, credential);
     }
 }
--- a/forge/src/main/java/org/openjdk/skara/forge/bitbucket/BitbucketForgeFactory.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/bitbucket/BitbucketForgeFactory.java
@@ -1,0 +1,43 @@
+package org.openjdk.skara.forge.bitbucket;
+
+import java.net.URI;
+import java.util.Set;
+import org.openjdk.skara.forge.Forge;
+import org.openjdk.skara.forge.ForgeFactory;
+import org.openjdk.skara.forge.ForgeUtils;
+import org.openjdk.skara.host.Credential;
+import org.openjdk.skara.json.JSONObject;
+
+public class BitbucketForgeFactory implements ForgeFactory {
+    @Override
+    public String name() {
+        return "bitbucket";
+    }
+
+    @Override
+    public Set<String> knownHosts() {
+        return Set.of();
+    }
+
+    @Override
+    public Forge create(URI uri, Credential credential, JSONObject configuration) {
+        var name = "Bitbucket";
+        if (configuration != null && configuration.contains("name")) {
+            name = configuration.get("name").asString();
+        }
+        var useSsh = false;
+        if (configuration != null && configuration.contains("sshkey") && credential != null) {
+            ForgeUtils.configureSshKey(credential.username(), uri.getHost(), configuration.get("sshkey").asString());
+            useSsh = true;
+        }
+        int sshport = 22;
+        if (configuration != null && configuration.contains("sshport")) {
+            sshport = configuration.get("sshport").asInt();
+        }
+        if (credential != null) {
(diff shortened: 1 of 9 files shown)


zhaosongzs (commented):
In general, this patch looks good to me!  Just a few minor problems. First, I think the new files need copyright header. Then, the GHA failed due to compile error, we need to import `org.openjdk.skara.cli.ForgeUtils` in `GitPrCreate` explicitly. Additionally, have you attempted to run a mirror bot using a BitBucket repo as the source on your local machine? If so, I believe this patch will work fine!

erikj79:
> In general, this patch looks good to me! Just a few minor problems. First, I think the new files need copyright header. Then, the GHA failed due to compile error, we need to import `org.openjdk.skara.cli.ForgeUtils` in `GitPrCreate` explicitly. Additionally, have you attempted to run a mirror bot using a BitBucket repo as the source on your local machine? If so, I believe this patch will work fine!

Thanks the feedback! Added copyright, completely forgot about that. To fix ForgeUtils, I moved it to a new package called "internal" instead. It's not meant to be exported to other modules.

Yes, I have run a local mirror configuration to verify the functionality. There isn't much of this that we have a model for automatic testing for.

forge/src/main/java/org/openjdk/skara/forge/bitbucket/BitbucketHost.java
```diff
@@ -0,0 +1,115 @@
...
+        this.credential = credential;
+    }
+
+    public BitbucketHost(String name, URI uri, boolean useSsh, int sshPort) {
+        this.name = name;
+        this.uri = uri;
+        this.useSsh = useSsh;
+        this.sshPort = sshPort;
+        this.credential = null;
+    }
```
edvbld, line 55: It seems like credential cannot be `null` when `useSsh` is `true` (see the method `sshHostString`). Maybe the constructors should reflect this? Or use static factory methods like `BitBucketHost.withSSH` and/or `BitBucketHost.withPAT`?

forge/src/main/java/org/openjdk/skara/forge/bitbucket/BitbucketRepository.java
```diff
@@ -0,0 +1,306 @@
...
+import org.openjdk.skara.json.JSONValue;
+import org.openjdk.skara.network.URIBuilder;
+import org.openjdk.skara.vcs.Branch;
+import org.openjdk.skara.vcs.Hash;
+import org.openjdk.skara.vcs.ReadOnlyRepository;
+import org.openjdk.skara.vcs.Tag;
+import org.openjdk.skara.vcs.VCS;
+
+public class BitbucketRepository implements HostedRepository {
+    private final BitbucketHost bitbucketHost;
```
edvbld, line 51: I would have named the field just `host`.

forge/src/main/java/org/openjdk/skara/forge/bitbucket/BitbucketRepository.java
```diff
@@ -0,0 +1,306 @@
...
+    private final String repositoryName;
```
edvbld, line 52: And here I would have just used the field `name` 😄

edvbld (approved):
A general comment regarding `ForgeUtils.configureSshKey`: I know this existing code, but I think a cleaner (and safer) solution would have been to stored the configuration in a separate file per host and then use `ssh -F /path/to/hosts/ssh/config/file`. Not worth thinking in this patch, but could be something for the future.

Otherwise just stylistic nits and minor style suggestions. Overall nice work and cool to see it how easy it was to add a minimal new forge!

forge/src/main/java/org/openjdk/skara/forge/bitbucket/BitbucketForgeFactory.java
```diff
@@ -0,0 +1,65 @@
...
+        var useSsh = false;
+        if (configuration != null && configuration.contains("sshkey") && credential != null) {
+            ForgeUtils.configureSshKey(credential.username(), uri.getHost(), configuration.get("sshkey").asString());
+            useSsh = true;
+        }
+        int sshport = 22;
+        if (configuration != null && configuration.contains("sshport")) {
+            sshport = configuration.get("sshport").asInt();
+        }
+        if (credential != null) {
```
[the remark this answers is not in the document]
erikj79: I did think of that, but decided against changing the config format in this change.

forge/src/main/java/org/openjdk/skara/forge/bitbucket/BitbucketHost.java
erikj79: I more or less copy pasted the existing gitlab implementation here so this is following the same existing patterns. The same relationship between useSsh and credential are present there too. I agree it could be made neater with a builder pattern like most other factories do, or static factory methods.

forge/src/main/java/org/openjdk/skara/forge/bitbucket/BitbucketRepository.java
erikj79: Right, again copy paste and just mindlessly modifying the field names. This I will adjust.

forge/src/main/java/org/openjdk/skara/forge/bitbucket/BitbucketHost.java
erikj79: I decided to make it simpler by just removing the second constructor and let the factory deal with requiring credentials for ssh. Having two constructors didn't really make sense anymore as no API calls to bitbucket were needed.

PR #1522, opened 2023-05-19
1914: IssuePoller always returns the most recently updated issue even if it has already been handled

zhaosongzs:
When I was testing with [SKARA-1912](https://bugs.openjdk.org/browse/SKARA-1912), I observed an issue in the IssuePoller. Although some updated csr issues has been handled in the previous round, the most recently updated csr issue would be returned again in the next round. Therefore, this would trigger the CSRIssueWorkItem and CheckWorkItem to update the PR very frequently. 

Erik investigated it and found the root cause. 

In the IssuePoller#updatedIssues, the poller first queries for updated issues (which may include some handled issues due to the imprecise timestamp, measured in minutes). Subsequently, the poller checks whether the issues have been really updated by comparing them with their copies using JiraIssue#equals. In this method, we compare the json of the issue with our own copy. If the issues are really updated, the poller returns them. 

Originally, the JiraIssue#equals method worked well. However, now, when the poller uses the REST API to retrieve the JSON of the issue, it includes a field called "customfield_11700," which represents the "Development" field used by Jira to display data quickly and easily in the issues. Within this field, there are various objects, such as "[SummaryItemBean@5e105688](mailto:SummaryItemBean@5e105688)." These objects change each time the poller makes a query. Therefore, even if the issue has not actually been updated, the JiraIssue#equals method will return false.

In this patch, some fields will be filtered out before comparing the jsons.

---------
### Progress

### Issue
 * [SKARA-1914](https://bugs.openjdk.org/browse/SKARA-1914): IssuePoller always returns the most recently updated issue even if it has already been handled

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1522/head:pull/1522` \
`$ git checkout pull/1522`

`$ git checkout pull/1522` \
`$ git pull https://git.openjdk.org/skara.git pull/1522/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1522`

`$ git pr show -t 1522`

</details>

<a href="https://git.openjdk.org/skara/pull/1522.diff">https://git.openjdk.org/skara/pull/1522.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1522#issuecomment-1555064783)

### DIFF
--- a/issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraHost.java
+++ b/issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraHost.java
@@ -55,7 +55,7 @@ public class JiraHost implements IssueTracker {
     /**
      * This constructor is only used by the manual test code.
      */
-    JiraHost(URI uri, String cookie) {
+    JiraHost(URI uri, String header, String value) {
         this.uri = uri;
         this.visibilityRole = null;
         this.securityLevel = null;
@@ -63,7 +63,7 @@ public class JiraHost implements IssueTracker {
         var baseApi = URIBuilder.base(uri)
                                 .appendPath("/rest/api/2/")
                                 .build();
-        request = new RestRequest(baseApi, "test", (r) -> Arrays.asList("Cookie", cookie));
+        request = new RestRequest(baseApi, "test", (r) -> Arrays.asList(header, value));
     }
 
     JiraHost(URI uri, JiraVault jiraVault) {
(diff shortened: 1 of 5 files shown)


erikj79 (commented):
I think this looks ok, but do you think you could write at least a manual test in `JiraProjectTests`? Let me know if you need helping setting up the environment to get such a test running.

zhaosongzs:
> I think this looks ok, but do you think you could write at least a manual test in `JiraProjectTests`? Let me know if you need helping setting up the environment to get such a test running.

I have tried this patch in staging, but I think you are right, it's better to add a manual test.

erikj79 (approved):

zhaosongzs:
Thanks for the review!
/integrate

PR #1523, opened 2023-05-22
1912: Show priority for bugs in pull request body

zhaosongzs:
The initial purpose of this issue is to show priority for bugs in the issue list in the pull request body. However, with current pr bot, if the user changes the type of the issue or the priority of the issue in the JBS, the changes will not be reflected in the PR body unless there is something could trigger the CheckWorkItem(like user edits the PR title and user issues some commands).

As Erik.J suggested, we could maintain a map(from JBS issue to the PRs related with this issue) in memory, so if the issue gets updated, we could know which pr needs to be updated.

To query updated issues and handle them, a new bot called `IssueBot` is introduced. This bot would query for updated issues(exclude CSR and JEP) in JBS, and it will read the in memory map to know whether there is any pr needs to be updated. If so, it will generate `CheckWorkItem` for that pr.

So currently, there are two ways to generate `CheckWorkItem`. Updated issues and updated pull requests. Previously, in the `CheckWorkItem`, we would check the metadata of the pull request and if the metadata is up to date, `jcheck` would not be triggered.  Now, we could check the metadata of the pull request and the issues related to this pr, but it would be too expensive because if only the pull request is updated, we also need to fetch all the issues from JBS to just generate the metadata. Therefore, in this patch, metadata is split to two parts, one part for pull request and one part for issues. If the `CheckWorkItem` is spawned from an updated pull request, we will only check pr metadata. On the other hand, if the `CheckWorkItem` is spawned from an updated issue, we will only check issues metadata.

Besides, since the map is in-memory, so if the bot restarts, the map needs to be initialized. When bot restarts, a `CheckWorkItem` would be generated for each pr, so the initialization is in `CheckWorkItem`.

---------
### Progress

### Issue
 * [SKARA-1912](https://bugs.openjdk.org/browse/SKARA-1912): Show priority for bugs in pull request body

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1523/head:pull/1523` \
`$ git checkout pull/1523`

`$ git checkout pull/1523` \
`$ git pull https://git.openjdk.org/skara.git pull/1523/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1523`

`$ git pr show -t 1523`

</details>

<a href="https://git.openjdk.org/skara/pull/1523.diff">https://git.openjdk.org/skara/pull/1523.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1523#issuecomment-1558141867)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -241,6 +320,23 @@ private boolean updateTitle() {
         return false;
     }
 
+    private void initializeIssuePRMap() {
+        // When bot restarts, the issuePRMap needs to get updated with this pr
```
erikj79, line 324: I'm not sure about this comment. This mapping needs to be updated every time there is a change to the associations, or if the map is currently missing associations for this PR. I don't think it makes sense to maintain an `initializedPRs` map, we should just always update the map.

This update process needs to be synchronized. The map is a `ConcurrentHashMap`, which is good, but we need to synchronize on the list value before trying to manipulate it.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/IssueWorkItem.java
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        if (!bot.equals(otherItem.bot)) {
+            return true;
+        }
+        return false;
+    }
+
+    @Override
+    public Collection<WorkItem> run(Path scratchPath) {
```
erikj79, line 82: I don't think we need a separate IssueWorkItem. The reason for the `CSRIssueWorkItem` is that we need to do potentially expensive work for each CSR issue found to translate them into `CheckWorkItem` (follow multiple issue links through queries). In this IssueWorkItem, all we need to do is look up the issue ID in an in-memory map, which is very cheap. That can easily be done in a loop in `bot.getPeriodicItems` without risking it taking too long.

erikj79 (commented):
This is a big change that will likely take a few rounds of review before we get it right. I have made a first pass and left comments mostly about higher level structural issues.

In addition to the below, I think it may be worth thinking about how we can minimize the amount of times we fetch the same issue from an IssueProject. Can we perhaps cache them in CheckWorkItem to avoid repeated fetching?

erikj79 (commented):
This is a big change that will likely take a few rounds of review before we get it right. I have made a first pass and left comments mostly about higher level structural issues.

In addition to the below, I think it may be worth thinking about how we can minimize the amount of times we fetch the same issue from an IssueProject. Can we perhaps cache them in CheckWorkItem to avoid repeated fetching?

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -681,15 +682,31 @@ private String getStatusMessage(PullRequestCheckIssueVisitor visitor,
...
+                                    } else {
+                                        progressBody.append(" - `" + issuePriority + "`)");
+                                    }
+                                    currentIssues.add(iss.get().id());
+                                } else if ("Enhancement".equals(issueType.asString())) {
+                                    progressBody.append(" (**Enhancement**)");
+                                    currentIssues.add(iss.get().id());
+                                } else {
+                                    progressBody.append(" (⚠️ Uncommon issue type: " + issueType.asString() + ")");
+                                    currentIssues.add(iss.get().id());
```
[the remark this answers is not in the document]
zhaosongzs: Ah, I do this because in the bug description, Erik D's demo code only show priority for `Bug`. I will fix it.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
zhaosongzs: I make the bot maintaining the` initializedPRs` map because I think there is no need to always update the map here. We  only need to update the map here when the bot restarts and it's the first time for the bot to see this pr. Otherwise, the update is extra work. 

The real update for the map located in `CheckRun#getStatusMessage` from line 744 to 768. 

And yes, I need to synchronize on the list value before trying to manipulate it. Thanks for catching!

bots/pr/src/main/java/org/openjdk/skara/bots/pr/IssueWorkItem.java
zhaosongzs: Yes, it makes sense. Will fix it.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
erikj79: Ah, I missed that part, that's good. With multiple locations doing updates to the map, I would recommend adding methods `addIssuePrMapping` and `removeIssuePrMapping` to `PullRequestBot` so you can encapsulate the dealing with lists and synchronization.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestBot.java
```diff
@@ -72,6 +73,8 @@ class PullRequestBot implements Bot {
     private final boolean jcheckMerge;
     private final Set<String> mergeSources;
     private final boolean enableBackport;
+    private final Map<String, List<String>> issuePRMap;
+    private final Map<String, Boolean> initializedPRs = new ConcurrentHashMap<>();
```
erikj79, line 77: This could just be a `Set`. Also it doesn't need to be concurrent as it's local to just this PullRequestBot, and we don't run WorkItems for the same PR in parallel.

zhaosongzs: Ah, right! Thanks!

zhaosongzs:
> This is a big change that will likely take a few rounds of review before we get it right. I have made a first pass and left comments mostly about higher level structural issues.
> 
> In addition to the below, I think it may be worth thinking about how we can minimize the amount of times we fetch the same issue from an IssueProject. Can we perhaps cache them in CheckWorkItem to avoid repeated fetching?

I think I have resolved most of the issues mentioned by you. And yes, it's possible for us to cache the issues, but I need more time to think about it, also, I think this patch is big enough. If you don't mind, I prefer to file a separate issue for this.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestBot.java
```diff
@@ -324,6 +329,33 @@ public Set<String> integrators() {
...
     }
 
+    public Map<String, List<PRRecord>> issuePRMap() {
+        return issuePRMap;
+    }
+
+    public void addIssuePRMapping(String issueId, PRRecord prRecord) {
+        issuePRMap.putIfAbsent(issueId, new LinkedList<>());
+        synchronized (issuePRMap.get(issueId)) {
+            List<PRRecord> prRecords = issuePRMap.get(issueId);
```
erikj79, line 339: I would suggest only fetching the list once and then synchronizing on it.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -723,6 +735,25 @@ private String getStatusMessage(PullRequestCheckIssueVisitor visitor,
...
+                    workItem.bot.addIssuePRMapping(issueId, prRecord);
+                }
+            }
+            // Delete associations
+            for (String oldIssueId : previousIssues) {
+                if (!currentIssues.contains(oldIssueId)) {
+                    workItem.bot.removeIssuePRMapping(oldIssueId, prRecord);
+                }
+            }
+            log.info("Map after updated: " + workItem.bot.issuePRMap());
```
erikj79, line 756: Logging the whole map here is probably not a good idea in production.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -241,6 +319,19 @@ private boolean updateTitle() {
...
+        var prRecord = new PRRecord(pr.repository().name(), pr.id());
+        if (!bot.initializedPRs().contains(prId)) {
+            var issueIds = BotUtils.parseIssues(pr.body());
+            for (String issueId : issueIds) {
+                bot.addIssuePRMapping(issueId, prRecord);
+            }
+            bot.initializedPRs().add(prId);
+            log.info("Map after initialization with pr " + pr.id() + " : " + bot.issuePRMap());
```
erikj79, line 331: Same here regarding logging. The map will get huge in production.

erikj79:
> > This is a big change that will likely take a few rounds of review before we get it right. I have made a first pass and left comments mostly about higher level structural issues.
> > In addition to the below, I think it may be worth thinking about how we can minimize the amount of times we fetch the same issue from an IssueProject. Can we perhaps cache them in CheckWorkItem to avoid repeated fetching?
> 
> I think I have resolved most of the issues mentioned by you. And yes, it's possible for us to cache the issues, but I need more time to think about it, also, I think this patch is big enough. If you don't mind, I prefer to file a separate issue for this.

That's ok, we can leave that as an optimization for later.

bots/common/src/main/java/org/openjdk/skara/bots/common/BotUtils.java
```diff
@@ -22,15 +22,34 @@
...
 public class BotUtils {
+    private static final String lineSep = "(?:\\n|\\r|\\r\\n|\\n\\r)";
+    private static final Pattern issuesBlockPattern = Pattern.compile(lineSep + lineSep + "###? Issues?((?:" + lineSep + "(?: \\* )?\\[.*)+)", Pattern.MULTILINE);
+    private static final Pattern issuePattern = Pattern.compile("^(?: \\* )?\\[(\\S+)]\\(.*\\): (.*$)", Pattern.MULTILINE);
 
     public static String escape(String s) {
         return s.replace("&", "&amp;").replace("<", "&lt;").replace(">", "&gt;");
     }
+
+    public static Set<String> parseIssues(String body) {
```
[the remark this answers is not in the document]
zhaosongzs: Sure. Will add some comments

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PRRecord.java
```diff
@@ -0,0 +1,65 @@
...
+    @Override
+    public int hashCode() {
+        return Objects.hash(repoName, prId);
+    }
+
+    @Override
+    public String toString() {
+        return repoName + "#" + prId;
+    }
+}
```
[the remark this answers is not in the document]
zhaosongzs: Oh, I didn't know Java had this feature before. I will use it! Thanks!

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestBot.java
zhaosongzs: Sure

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
zhaosongzs: Ok, I will remove it.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestBot.java
zhaosongzs: Think again about this, It's possible that CheckWorkItems for different PRs in the same repo run in parallel. Every CheckWorkItem is possible to modify `initializedPRs`. So we still need synchronize it.

erikj79: You are right, the map will be concurrently modified, just not the same element, and that requires a `ConcurrentHashMap` or synchronization to avoid corrupting data. I think `ConcurrentHashMap` is preferred here.

zhaosongzs: Yes, I will change it back

erikj79 (approved):

PR #1524, opened 2023-06-01
SKARA-1925

zhaosongzs:
A user reported that there is no email generated for this comment. 
https://github.com/openjdk/jdk/pull/14114#issuecomment-1562126987 

It is caused by the initial `/csr needed` command. 

The logic about filtering out commands from a comment differs between `ArchiveWorkItem#ignoreComment` and `CommandExtractor#extractCommands`.

In this patch, I make the logic consistent in this two methods.

1. Any line starts with '/' followed by lowercase characters will be recognized as a command line. The command line will be stripped. 
2. Check whether this command is a multiline command. If so, the following lines will be considered as arguments of the command and the argument lines will be stripped until the bot found another command line.

### DIFF
--- a/bots/common/src/main/java/org/openjdk/skara/bots/common/CommandNameEnum.java
+++ b/bots/common/src/main/java/org/openjdk/skara/bots/common/CommandNameEnum.java
@@ -1,3 +1,25 @@
+/*
+ * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
 package org.openjdk.skara.bots.common;
 
 import java.util.stream.Collectors;
@@ -11,7 +33,7 @@ public enum CommandNameEnum {
     integrate,
     sponsor,
     contributor,
-    summary,
+    summary(true),
     issue,
     solves,
     reviewers,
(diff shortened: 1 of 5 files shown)


bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveWorkItem.java
```diff
@@ -118,21 +119,41 @@ private Repository materializeArchive(Path scratchPath) {
...
+            } else {
+                if (!readingMultiLineCommandArgs) {
+                    filteredBody.append(line).append(System.lineSeparator());
+                }
+            }
+        }
+        return filteredBody.toString().strip();
+    }
+
+    private boolean ignoreComment(HostUser author, String body, ZonedDateTime createdTime, ZonedDateTime lastDraftTime, boolean isComment) {
```
zhaosongzs, line 143: Introduced argument `isComment` to this method because I think in any case, no review should be ignored. But according to the previous logic, if a review whose body only contains command, the review would be ignored.

bots/common/src/main/java/org/openjdk/skara/bots/common/CommandNameEnum.java
```diff
@@ -25,6 +47,19 @@ public enum CommandNameEnum {
...
+    }
+
+    CommandNameEnum(boolean isMultiLine) {
+        this.isMultiLine = isMultiLine;
+    }
+
+    public boolean isMultiLine() {
+        return isMultiLine;
+    }
+
```
zhaosongzs, line 62: Introduced a field called` isMultiLine` here because `CommandHandler` is coupled in PR bot and it is very difficult to extract it to the bots::common package.

bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/ArchiveWorkItem.java
erikj79: I'm not sure about this. We already send email when someone approves a PR. I don't think we need to send emails with an empty review.

zhaosongzs: According to the previous logic, for example, if we approve a PR and the review body only contains command like "/reviewers 2", then this review will be ignored. This issue is what I want to resolve.

But now I also think something in my logic is wrong.

erikj79: See this PR that I reviewed this morning: https://github.com/openjdk/jdk/pull/14261

I didn't put any text in the review comment/body, but we still had this email sent: https://mail.openjdk.org/pipermail/build-dev/2023-June/039655.html

zhaosongzs: Yes, empty body is good. But if and only if the review body contains command, it will be ignored. 

Just like when you are approving the pr, you put "/reviewers 2"(or any other command) in the review body.

erikj79: Ah, that's pretty bad, but I see what you mean now after having studied the code.

erikj79 (approved):

PR #1528, opened 2023-06-07
1936: IssueCommand would change pr title in GitLab

zhaosongzs:
In GitLab, when a user wants to create a PR in draft mode, they need to include the prefix "Draft: " before the title(Checked the GitLab Document, "WIP" is already deprecated and removed). However, if the user wants to use the "/issue" command to add another issue to this PR, the pr bot will not be able to parse the issue within the PR title. Therefore, the bot will modify the PR title to match the new issue the user wants to add, resulting in the removal of the draft state of the PR.

---------
### Progress

### Issue
 * [SKARA-1936](https://bugs.openjdk.org/browse/SKARA-1936): IssueCommand would change pr title in GitLab (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1528/head:pull/1528` \
`$ git checkout pull/1528`

`$ git checkout pull/1528` \
`$ git pull https://git.openjdk.org/skara.git pull/1528/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1528`

`$ git pr show -t 1528`

</details>

<a href="https://git.openjdk.org/skara/pull/1528.diff">https://git.openjdk.org/skara/pull/1528.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1528#issuecomment-1581627685)

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/IssueCommand.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/IssueCommand.java
@@ -162,7 +162,7 @@ public class IssueCommand implements CommandHandler {
             return;
         }
 
-        var titleIssue = Issue.fromStringRelaxed(pr.title());
+        var titleIssue = Issue.fromStringRelaxed(removeDraftPrefix(pr.title()));
         for (var issue : validatedIssues) {
             if (titleIssue.isEmpty()) {
                 reply.print("The primary solved issue for a PR is set through the PR title. Since the current title does ");
@@ -183,6 +183,11 @@ public class IssueCommand implements CommandHandler {
                 reply.println("Adding additional issue to " + name + " list: `" + issue.toShortString() + "`.");
             }
         }
+    }
+
+    private String removeDraftPrefix(String title) {
+        String pattern = "(?i)^draft:?\\s*";
+        return title.replaceAll(pattern, "");
     }
 
     private void removeIssue(PullRequestBot bot, String args, Set<String> currentSolved, PrintWriter reply) throws InvalidIssue {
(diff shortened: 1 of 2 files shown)


kevinrushforth:
Is there a better place to remove the `Draft:` prefix? I'm wondering if it would be possible to generalize it by removing the "Draft:" prefix similarly to how the optional `JDK-` prefix is removed? That way it would be parsed as a valid issue for the purpose of listing the BugID + Bug tltle in the Description, checking for a CSR, etc.

zhaosongzs:
> Is there a better place to remove the `Draft:` prefix? I'm wondering if it would be possible to generalize it by removing the "Draft:" prefix similarly to how the optional `JDK-` prefix is removed? That way it would be parsed as a valid issue for the purpose of listing the BugID + Bug tltle in the Description, checking for a CSR, etc.

I will think about whether we could remove "Draft:" prefix in the method `pr.title()`. And it's a good idea that we could also remove "Draft: " prefix when generating pr body for draft prs. In that way, as you said, the issue could be listed in the pr body and checking for CSR can be done.  Thanks for the advice!

zhaosongzs:
I have updated the solution of this issue. Now, `GitLabMergeRequest#title` would return title without draft prefix and `GitLabMergeRequest#setTitle` would check whether the pr is in draft mode, if so, it will add draft prefix to the title.

Previous, when the user creates a pr in GitLab and set the title to "Draft: 123", the pr bot will not be able to update the title and checking for CSR etc. Now, all the behavior of draft pr in GitLab will be the same as in GitHub. 

Thanks to @kevinrushforth 's advice again!

erikj79 (approved):

zhaosongzs:
Thanks for the review!
/integrate

PR #1531, opened 2023-06-12
SKARA-1937

zhaosongzs:
As Erik said in the issue: "When we bump the version in .jcheck/conf from N to N+1 in the jdk master branch, PRs with CSRs are often not updated to reflect this."

This root cause of this is that changing of .jcheck/conf in target ref of the pr will not be able to trigger checkWorkItem for this pr.

In this patch,
1. The prBot will monitor the changes to the .jcheck/conf file in any branch(except for PR branches).
2. If the prBot detects an update to the .jcheck/conf file in a branch, it will trigger a checkWorkItem for all open PRs whose target ref is the branch where the .jcheck/conf file was updated.
3. To solve the problem related with bot restarting, the contents of the .jcheck/conf file in the target branch of a PR will be included in the metadata hash.

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -1335,7 +1335,7 @@ class CheckRun {
 
             // Calculate current metadata to avoid unnecessary future checks
             var metadata = workItem.getMetadata(workItem.getPRMetadata(censusInstance, title, updatedBody, pr.comments(), activeReviews,
-                                                newLabels, pr.targetRef(), pr.isDraft()), workItem.getIssueMetadata(updatedBody), expiresIn);
+                    newLabels, pr.targetRef(), pr.isDraft(), pr.targetRefJCheckConf()), workItem.getIssueMetadata(updatedBody), expiresIn);
             checkBuilder.metadata(metadata);
         } catch (Exception e) {
             log.throwing("CommitChecker", "checkStatus", e);
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestBot.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestBot.java
@@ -35,6 +35,7 @@ import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.logging.*;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 class PullRequestBot implements Bot {
     private final HostedRepository remoteRepo;
@@ -75,6 +76,7 @@ class PullRequestBot implements Bot {
     private final boolean enableBackport;
     private final Map<String, List<PRRecord>> issuePRMap;
     private final Map<String, Boolean> initializedPRs = new ConcurrentHashMap<>();
+    private final Map<String, String> jCheckConfMap = new HashMap<>();
 
     private Instant lastFullUpdate;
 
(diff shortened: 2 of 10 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestBot.java
```diff
@@ -165,11 +167,43 @@ public List<WorkItem> getPeriodicItems() {
...
         return workItems;
     }
 
+    private List<PullRequest> getJCheckConfUpdateRelatedPRs() {
+        var ret = new ArrayList<PullRequest>();
+        // Get all branches except 'pr' branch
+        var allTargetRefs = remoteRepo.branches().stream()
+                .map(HostedBranch::name)
+                .filter(name -> !name.startsWith("pr/"))
+                .toList();
```
zhaosongzs, line 189: Currently, fetching all existing branches is the main overhead in this feature. It takes about 2s for JDK repo. If this overhead is not acceptable, the other way to solve this is store the branches in bot config.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -1335,7 +1335,7 @@ private void checkStatus() {
 
             // Calculate current metadata to avoid unnecessary future checks
             var metadata = workItem.getMetadata(workItem.getPRMetadata(censusInstance, title, updatedBody, pr.comments(), activeReviews,
-                                                newLabels, pr.targetRef(), pr.isDraft()), workItem.getIssueMetadata(updatedBody), expiresIn);
+                    newLabels, pr.targetRef(), pr.isDraft(), pr.targetRefJCheckConf()), workItem.getIssueMetadata(updatedBody), expiresIn);
```
erikj79, line 1338: The .jcheck/conf file is already read when creating the `CensusInstance` (and may sometimes even be overridden and located in another repository). I think we should store the raw .jcheck/conf string in `LimitedCensusInstance` and get it from there in the rest of `CheckWorkItem`/`CheckRun`.

zhaosongzs: Yes, it's a good point. Will fix it

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestBot.java
zhaosongzs: Yes, it will be much better. No extra overhead will be added.

jcheck/src/main/java/org/openjdk/skara/jcheck/JCheckConfiguration.java
```diff
@@ -149,12 +155,13 @@ private static INI convert(INI old) {
 
     public static JCheckConfiguration parse(List<String> lines) {
         var ini = INI.parse(lines);
+        var rawConf = String.join("", lines);
```
erikj79, line 158: This isn't actually the raw conf anymore, since it's been split by lines and then put together, but the original line breaks are lost. If we are calling this the "rawConf" then we should at least put `\n` back in there. Alternatively, we could store just a hash of the lines to make it clear that it can't be used to compare to another raw source of .jcheck/conf in the future.

I think just putting linebreaks in is enough for now.

zhaosongzs: Will add the line breaks back. Thanks!

erikj79 (approved):

zhaosongzs:
Thanks for the review!
/integrate

PR #1534, opened 2023-06-16
1951: Refactor Issue class hierarchy

erikj79:
The class hierarchy around the (org.openjdk.skara.issuetracker.)Issue interface is weird and cumbersome to work with. This interface is representing an Issue as returned from an IssueTracker. It's implemented by `JiraIssue` (and `TestIssue` for testing purposes). It's also extended by `PullRequest`, which makes it awkward to add Jira/bug tracker specific functionality to it since we must then also implement that for all `PullRequest` implementations (usually by just throwing a runtime exception). It's rare to use `Issue` as something that can be either a `PullRequest` or an actual issue, so sharing this interface isn't adding much value.

In addition to this, there is also an unrelated class called Issue (org.openjdk.skara.vcs.openjdk.Issue) in the vcs module. This class just represents the OpenJDK interpretation of an issue in the context of a commit message. These classes are however used together sometimes, which is causing a lot of confusion.

I would like to improve this situation. My proposal is to introduce a new interface `IssueTrackerIssue` which is a sub type of `Issue` that contains methods specific to bug trackers (or at least Jira). In practice right now this is links and properties. The name reflects that this is the kind of issue returned from an `IssueTracker` as opposed to an Issue from just any kind of `Forge`. Having that in an interface enables us to also define a specific test implementation `TestIssueTrackerIssue` that mimics our `JiraIssue` implementation without clashing with `PullRequest` and its implementations. With this new type name, we can also get around the clash when having to deal with the vcs Issue type. This solution retains the shared super interface `Issue` between `IssueTrackerIssue` and `PullRequest`. I'm also refactoring `TestIssue` and its subclasses to reflect this new hierarchy.

The patch is pretty big, which is why I'm doing this in isolation. No functionality should change.

---------
### Progress

### Issue
 * [SKARA-1951](https://bugs.openjdk.org/browse/SKARA-1951): Refactor Issue class hierarchy (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1534/head:pull/1534` \
`$ git checkout pull/1534`

`$ git checkout pull/1534` \
`$ git pull https://git.openjdk.org/skara.git pull/1534/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1534`

`$ git pr show -t 1534`

</details>

<a href="https://git.openjdk.org/skara/pull/1534.diff">https://git.openjdk.org/skara/pull/1534.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1534#issuecomment-1595372139)

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/JEPCommand.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/JEPCommand.java
@@ -25,7 +25,6 @@ package org.openjdk.skara.bots.pr;
 
 import org.openjdk.skara.forge.PullRequest;
 import org.openjdk.skara.issuetracker.Comment;
-import org.openjdk.skara.issuetracker.Issue;
 
 import java.io.PrintWriter;
 import java.util.List;
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/JEPCommand.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/JEPCommand.java
@@ -30,6 +30,7 @@ import org.openjdk.skara.issuetracker.Issue;
 import java.io.PrintWriter;
 import java.util.List;
 import java.util.Optional;
+import org.openjdk.skara.issuetracker.IssueTrackerIssue;
 
 import static org.openjdk.skara.bots.common.CommandNameEnum.jep;
 import static org.openjdk.skara.issuetracker.jira.JiraProject.JEP_NUMBER;
@@ -60,8 +61,8 @@ public class JEPCommand implements CommandHandler {
                 """);
     }
 
-    private Optional<Issue> getJepIssue(String args, PullRequestBot bot) {
-        Optional<Issue> jbsIssue = Optional.empty();
+    private Optional<IssueTrackerIssue> getJepIssue(String args, PullRequestBot bot) {
+        Optional<IssueTrackerIssue> jbsIssue = Optional.empty();
         var upperArgs = args.toUpperCase();
         if (upperArgs.startsWith("JEP-")) {
             // Handle the JEP ID with `JEP` prefix
--- a/bots/pr/src/test/java/org/openjdk/skara/bots/pr/IssueTests.java
+++ b/bots/pr/src/test/java/org/openjdk/skara/bots/pr/IssueTests.java
@@ -432,7 +432,7 @@ class IssueTests {
             // Make a change with a corresponding PR
             var editHash = CheckableRepository.appendAndCommit(localRepo);
             localRepo.push(editHash, author.authenticatedUrl(), "edit", true);
-            var issue1 = (TestIssue)issues.createIssue("First", List.of("Hello"), Map.of());
+            var issue1 = (TestIssueTrackerIssue)issues.createIssue("First", List.of("Hello"), Map.of());
             issue1.setState(Issue.State.CLOSED);
             issue1.store().properties().put("resolution", JSON.object().put("name", JSON.of("Not an Issue")));
             var pr = credentials.createPullRequest(author, "master", "edit",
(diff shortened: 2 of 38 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/JEPCommand.java
```diff
@@ -30,6 +30,7 @@
 import java.io.PrintWriter;
 import java.util.List;
 import java.util.Optional;
+import org.openjdk.skara.issuetracker.IssueTrackerIssue;
```
zhaosongzs, line 33: `import org.openjdk.skara.issuetracker.Issue;` in line 28 can be removed

bots/pr/src/test/java/org/openjdk/skara/bots/pr/IssueTests.java
```diff
@@ -432,7 +432,7 @@ void closedIssue(TestInfo testInfo) throws IOException {
             // Make a change with a corresponding PR
             var editHash = CheckableRepository.appendAndCommit(localRepo);
             localRepo.push(editHash, author.authenticatedUrl(), "edit", true);
-            var issue1 = (TestIssue)issues.createIssue("First", List.of("Hello"), Map.of());
+            var issue1 = (TestIssueTrackerIssue)issues.createIssue("First", List.of("Hello"), Map.of());
```
zhaosongzs, line 435:
```suggestion
            var issue1 = (TestIssueTrackerIssue) issues.createIssue("First", List.of("Hello"), Map.of());
```

bots/pr/src/test/java/org/openjdk/skara/bots/pr/IssueTests.java
```diff
@@ -473,7 +473,7 @@ void resolvedIssue(TestInfo testInfo) throws IOException {
...
```
zhaosongzs, line 476:
```suggestion
            var issue1 = (TestIssueTrackerIssue) issues.createIssue("First", List.of("Hello"), Map.of());
```

bots/pr/src/test/java/org/openjdk/skara/bots/pr/IssueTests.java
zhaosongzs: It's my fault, I forgot to add the whitespace when writing the tests

zhaosongzs (commented):
I think this patch is good, and your design of the class hierarchy makes a lot of sense. I just have these minor suggestions. Additionally, copyrights in some updated files need to be updated.

Furthermore, regarding setting or getting the state of an IssueTrackerIssue or a Pull Request, we currently utilize `Issue.State.XXX`. However, I think it might be better to use `IssueTrackerIssue.State.XXX` and `PullRequest.State.XXX`.

erikj79:
> Furthermore, regarding setting or getting the state of an IssueTrackerIssue or a Pull Request, we currently utilize `Issue.State.XXX`. However, I think it might be better to use `IssueTrackerIssue.State.XXX` and `PullRequest.State.XXX`.

Not sure if you mean that we should have separate state enums for `IssueTrackerIssue` and `PullRequest` or if you are just wanting to save an import by referencing the enum through the sub interfaces. If the former, then if we need to express different states in issues vs pull requests, then yes we should have different enums, but I don't think we do right now at least. If the later, I don't think optimizing or reducing imports is something to strive for. They are handled automatically by the IDE.

Another variant of this change would be to remove the common super interface between issue and pull request completely. It really is a false abstraction to begin with.

zhaosongzs:
> > Furthermore, regarding setting or getting the state of an IssueTrackerIssue or a Pull Request, we currently utilize `Issue.State.XXX`. However, I think it might be better to use `IssueTrackerIssue.State.XXX` and `PullRequest.State.XXX`.
> 
> Not sure if you mean that we should have separate state enums for `IssueTrackerIssue` and `PullRequest` or if you are just wanting to save an import by referencing the enum through the sub interfaces. If the former, then if we need to express different states in issues vs pull requests, then yes we should have different enums, but I don't think we do right now at least. If the later, I don't think optimizing or reducing imports is something to strive for. They are handled automatically by the IDE.
> 
> Another variant of this change would be to remove the common super interface between issue and pull request completely. It really is a false abstraction to begin with.

What I mean was the later one. I was thinking using a specific subtype of `Issue` would be better. It could help us to distinguish between an Issue and a pull request. And also, if later we want to separate state enums for `IssueTrackerIssue` and `PullRequest`, it could save us some work.

zhaosongzs (approved):
Looks good now

erikj79:
Thanks for review!

/integrate

PR #1537, opened 2023-06-27
1199: Enforce maintainer approval in JBS before allowing to integrate backports into updates projects

zhaosongzs:
As Erik said in the description of this issue, currently, this issue only cares about tracking approval labels in the related bugs.

If a repository is set up with the "approval" configuration, pull requests in that repository will require the maintainer's approval in JBS. Otherwise, the pull request will not be considered ready. 

Erik has also provided a design outlining how to configure the "approval" for a repository.

```
The simple case, where the labels are the same for every branch in a repository: 

"approval": { 
  "request": "jdk17u-fix-request", 
  "approved": "jdk17u-fix-yes", 
  "rejected": "jdk17u-fix-no", 
} 

To reduce the need for changing multiple strings when copying a configuration for a new repository, there is an optional "prefix" field: 

"approval": { 
  "prefix": "jdk17u-fix-", 
  "request": "request", 
  "approved": "yes", 
  "rejected": "no", 
} 

When there are multiple branches with different labels, having the prefix set per branch can help reduce the size of the configuration significantly: 

"approval": { 
  "request": "-critical-request", 
  "approved": "-critical-approved", 
  "rejected": "-critical-rejected", 
  "branches": [ 
    "jdk20\\.0\\.1": { "prefix": "CPU23_04" } 
  ] 
} 
```

---------
### Progress

### Issue
 * [SKARA-1199](https://bugs.openjdk.org/browse/SKARA-1199): Enforce maintainer approval in JBS before allowing to integrate backports into updates projects (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1537/head:pull/1537` \
`$ git checkout pull/1537`

`$ git checkout pull/1537` \
`$ git pull https://git.openjdk.org/skara.git pull/1537/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1537`

`$ git pr show -t 1537`

</details>

<a href="https://git.openjdk.org/skara/pull/1537.diff">https://git.openjdk.org/skara/pull/1537.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1537#issuecomment-1610305331)

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -275,7 +275,7 @@ class CheckRun {
                     var issue = issueOpt.get();
                     var labelNames = issue.labelNames();
                     if (labelNames.contains(approval.approvedLabel(pr.targetRef()))) {
-                        ret.put("[" + issue.id() + "](" + issue.webUrl() + ") needs maintainer approval in JBS", true);
+                        ret.put("[" + issue.id() + "](" + issue.webUrl() + ") needs maintainer approval", true);
                     } else {
                         ret.put("[" + issue.id() + "](" + issue.webUrl() + ") needs maintainer approval in JBS", false);
                     }
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -1514,7 +1514,7 @@ class CheckRun {
             return;
         }
         String message = "⚠️  @" + pr.author().username() +
-                " This change has now been reviewed and requires maintainer [approval](" + approval.documentLink() + ")." +
+                " This change is now ready for maintainer [approval](" + approval.documentLink() + ")." +
                 APPROVAL_NEEDED_MARKER;
         pr.addComment(message);
     }
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -70,6 +70,7 @@ class CheckRun {
     private static final String EMPTY_PR_BODY_MARKER = "<!--\nReplace this text with a description of your pull request (also remove the surrounding HTML comment markers).\n" +
             "If in doubt, feel free to delete everything in this edit box first, the bot will restore the progress section as needed.\n-->";
     private static final String FULL_NAME_WARNING_MARKER = "<!-- PullRequestBot full name warning comment -->";
+    private static final String APPROVAL_NEEDED_MARKER = "<!-- PullRequestBot approval needed comment -->";
     private static final Set<String> PRIMARY_TYPES = Set.of("Bug", "New Feature", "Enhancement", "Task", "Sub-task");
     private final Set<String> newLabels;
     private final boolean reviewCleanBackport;
@@ -269,7 +270,7 @@ class CheckRun {
                                                        IssueTrackerIssue jepIssue, JdkVersion version) {
         var ret = new HashMap<String, Boolean>();
 
-        if (approval != null && approval.needsApproval(pr.targetRef())) {
+        if (approvalNeeded()) {
             for (var issueOpt : regularIssuesMap.values()) {
                 if (issueOpt.isPresent()) {
                     var issue = issueOpt.get();
@@ -698,7 +699,7 @@ class CheckRun {
                             if (issuePriority != null) {
                                 progressBody.append(" - P").append(issuePriority.asString());
                             }
-                            if (approval != null && approval.needsApproval(pr.targetRef())) {
+                            if (approvalNeeded()) {
                                 String status = "";
                                 String targetRef = pr.targetRef();
                                 var labels = issueTrackerIssue.get().labelNames();
@@ -1318,7 +1319,9 @@ class CheckRun {
                                       integrationBlockers.isEmpty() &&
                                       !statusMessage.contains(TEMPORARY_ISSUE_FAILURE_MARKER);
             }
-
+            if (approvalNeeded()) {
+                updateApprovalNeededComment(additionalProgresses);
+            }
             updateMergeReadyComment(readyForIntegration, commitMessage, rebasePossible);
             if (readyForIntegration && rebasePossible) {
                 newLabels.add("ready");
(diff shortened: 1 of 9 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -677,11 +695,25 @@ private String getStatusMessage(PullRequestCheckIssueVisitor visitor,
...
+                                String targetRef = pr.targetRef();
+                                var labels = issueTrackerIssue.get().labelNames();
+                                if (labels.contains(approval.rejectedLabel(targetRef))) {
+                                    status = "Rejected";
+                                } else if (labels.contains(approval.approvedLabel(targetRef))) {
+                                    status = "Approved";
+                                } else if (labels.contains(approval.requestedLabel(targetRef))) {
+                                    status = "Requested";
+                                }
+                                if (!status.isEmpty() && !status.isBlank()) {
```
erikj79, line 712: Checking just for empty should be enough here.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -179,6 +179,7 @@ String getIssueMetadata(String prBody) {
                             issueData.append(properties.get("priority").asString());
                             issueData.append(properties.get("issuetype").asString());
                         }
+                        issueData.append(String.join("", issue.labelNames()));
```
erikj79, line 182: This is currently adding another query to the issue tracker, but that can actually be fixed, and I think it's worth doing that in this patch. You can override `Issue::labelNames` in `JiraIssue` and just read the names from the json. There is a labels field with the names in it. I had this in a local patch for a while but it didn't fit with the thing I was working on at the time, and then I forgot.

erikj79:
When a PR is ready with everything but approval, we should also update the status comment with a helpful message linking to the approval process. This is the same comment where the "ready to integrate" message is posted. That link needs to be configurable.

erikj79:
> > When a PR is ready with everything but approval, we should also update the status comment with a helpful message linking to the approval process. This is the same comment where the "ready to integrate" message is posted. That link needs to be configurable.
> 
> Do you mean the `mergeReadyComment`?(I saw your comment in the issue description) But currently, the maintainer approval is a `progress`. The PR would not be considered ready without maintainer approval label in the related issue, and the mergeReadyComment would not be posted unless the PR is considered ready.

That's what I meant. This would mean that we post the comment earlier and that we should rename the concept of that comment to something like `nextStepComment`. But maybe we shouldn't use the same comment, but instead introduce a new `approvalNeededComment`. Otherwise there is risk that it gets lost in the list of comments after a while. This new comment would be added when only the approval is left to do. Later in [SKARA-1949](https://bugs.openjdk.org/browse/SKARA-1949), the contents of this comment would be expanded to contain instructions for how to apply for maintainer approval using Skara commands. For now it can only reference the process documentation.

zhaosongzs:
> > > When a PR is ready with everything but approval, we should also update the status comment with a helpful message linking to the approval process. This is the same comment where the "ready to integrate" message is posted. That link needs to be configurable.
> > 
> > 
> > Do you mean the `mergeReadyComment`?(I saw your comment in the issue description) But currently, the maintainer approval is a `progress`. The PR would not be considered ready without maintainer approval label in the related issue, and the mergeReadyComment would not be posted unless the PR is considered ready.
> 
> That's what I meant. This would mean that we post the comment earlier and that we should rename the concept of that comment to something like `nextStepComment`. But maybe we shouldn't use the same comment, but instead introduce a new `approvalNeededComment`. Otherwise there is risk that it gets lost in the list of comments after a while. This new comment would be added when only the approval is left to do. Later in [SKARA-1949](https://bugs.openjdk.org/browse/SKARA-1949), the contents of this comment would be expanded to contain instructions for how to apply for maintainer approval using Skara commands. For now it can only reference the process documentation.

Ah, got it! Would add the comment.

zhaosongzs:
> > > When a PR is ready with everything but approval, we should also update the status comment with a helpful message linking to the approval process. This is the same comment where the "ready to integrate" message is posted. That link needs to be configurable.
> > 
> > 
> > Do you mean the `mergeReadyComment`?(I saw your comment in the issue description) But currently, the maintainer approval is a `progress`. The PR would not be considered ready without maintainer approval label in the related issue, and the mergeReadyComment would not be posted unless the PR is considered ready.
> 
> That's what I meant. This would mean that we post the comment earlier and that we should rename the concept of that comment to something like `nextStepComment`. But maybe we shouldn't use the same comment, but instead introduce a new `approvalNeededComment`. Otherwise there is risk that it gets lost in the list of comments after a while. This new comment would be added when only the approval is left to do. Later in [SKARA-1949](https://bugs.openjdk.org/browse/SKARA-1949), the contents of this comment would be expanded to contain instructions for how to apply for maintainer approval using Skara commands. For now it can only reference the process documentation.

And now I am thinking why we only post `approvalNeededComment` when a PR is ready with everything but approval? What about posting `approvalNeededComment` when approval is missing?

erikj79:
> And now I am thinking why we only post `approvalNeededComment` when a PR is ready with everything but approval? What about posting `approvalNeededComment` when approval is missing? In this way, user could see the instructions earlier and they would have the time to apply for approval.

I think we should refrain from posting it early because the formal approval process shouldn't start until the PR is reviewed. Posting the comment early might encourage users to post approval requests early, which will annoy the maintainers.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -1490,4 +1493,38 @@ private String describe(PullRequest pr) {
...
+        }
+        StringBuilder messageBuilder = new StringBuilder();
+        if (existingUnapproved) {
+            messageBuilder.append("⚠️  @").append(pr.author().username())
+                    .append(" There are still some issues that have not received maintainer approval.")
+                    .append("\n")
+                    .append("Please follow the instruction here to get maintainer approval: ")
+                    .append("[Requesting push approval for fixes](https://openjdk.org/projects/jdk-updates/approval.html)");
+        } else {
+            messageBuilder.append("@").append(pr.author().username()).append(" All the issues have already got the maintainer approval!");
```
erikj79, line 1517: When looking at this, I think I'm changing my mind a bit around this. I think we should just post the "approval needed" comment once and not edit it with progress. It's enough to have progress in the pr body.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -1497,34 +1508,15 @@ private boolean approvalNeeded() {
...
         if (existing.isPresent()) {
-            if (!existing.get().body().equals(message)) {
-                pr.updateComment(existing.get().id(), message);
-            }
-        } else {
-            pr.addComment(message);
+            return;
         }
+        String message = "⚠️  @" + pr.author().username() +
+                " This change has now been reviewed and requires maintainer [approval](" + approval.documentLink() + ")." +
```
erikj79, line 1517: I did some more thinking about how this should be formulated. I don't think we should mention the review state here either. 
```suggestion
                " This change is now ready for maintainer [approval](" + approval.documentLink() + ")." +
```

bots/pr/src/test/java/org/openjdk/skara/bots/pr/ApprovalTests.java
```diff
@@ -31,19 +31,20 @@
 public class ApprovalTests {
     @Test
     void simple() {
-        Approval approval = new Approval("", "jdk17u-fix-request", "jdk17u-fix-yes", "jdk17u-fix-no");
+        Approval approval = new Approval("", "jdk17u-fix-request", "jdk17u-fix-yes", "jdk17u-fix-no", "https://test.openjdk.org");
```
erikj79, line 34: I recommend using something with `example.com` for test URL.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
erikj79:
```suggestion
                " This change is now ready for you to apply for maintainer [approval](" + approval.documentLink() + ")." +
```

bots/pr/src/test/java/org/openjdk/skara/bots/pr/ApprovalTests.java
zhaosongzs: Sure, will fix it.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
erikj79: Did you see my last suggestion here? I think we need to be clear that the ball is in the hands of the PR owner.

zhaosongzs: Sorry, I missed it. Will fix it soon.

erikj79 (approved):

zhaosongzs:
Thanks for the review!
/integrate

PR #1546, opened 2023-08-21
SKARA-1999

zhaosongzs:
In [SKARA-1199](https://bugs.openjdk.org/browse/SKARA-1199), I made a bug about posting approvalNeededComment. 

Before the bot posting ApprovalNeededComment, we need to make sure this pr could pass all the checks.

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -1332,7 +1332,7 @@ class CheckRun {
                                       !statusMessage.contains(TEMPORARY_ISSUE_FAILURE_MARKER);
             }
 
-            if (approvalNeeded()) {
+            if (approvalNeeded() && visitor.messages().isEmpty()) {
                 var readyButMaintainerApproval = true;
                 for (var entry : additionalProgresses.entrySet()) {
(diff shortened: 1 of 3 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -1332,7 +1332,7 @@ private void checkStatus() {
                                       !statusMessage.contains(TEMPORARY_ISSUE_FAILURE_MARKER);
             }
 
-            if (approvalNeeded()) {
+            if (approvalNeeded() && visitor.messages().isEmpty()) {
```
erikj79, line 1335: Is this enough? Shouldn't we check `readyForReview` as well. What about the other things in the list for `readyForIntegration`?

zhaosongzs: You are right. I will fix the problem soon. Thank you.

zhaosongzs: I also refactored the code about `readyForIntegration`.

erikj79 (approved):

zhaosongzs:
Thanks for the review!
/integrate

PR #1548, opened 2023-08-22
SKARA-2000

zhaosongzs:
In this patch, the bot would escape `@` in the issue title in pr body.

I tried to escape `@` with `&#64;`, but failed, the text would still mention the link.

So according to this [page](https://github.com/github/markup/issues/1168#issuecomment-494946168), I escape `@` with `@<!---->`.


zhaosongzs:
`&#64;zhaosong` &#64;zhaosong 
vs 
`@<!---->zhaosong` @zhaosong

pavelrappo (approved):
Not a Skara reviewer, but the change looks good. It's sad to see that HTML entities, `&#64;` or `&commat;`, whose work is to help avoid unneeded interpretation, do not work in this case.

erikj79 (approved):
That's a creative solution. Nice that you found something that works.

zhaosongzs:
Thanks for the review!
/integrate

PR #1549, opened 2023-08-22
2002: GitToHgConverter fails on merge with ancestor

erikj79:
Sometimes through weird circumstances we can end up with Git repositories where merges between a commit and an ancestor to that commit. Converting such a commit to Mercurial can be tricky, and in one case we are currently failing.

The GitToHgConverter is already trying to solve this issue, but only detects the ancestor relationship in one direction between the commits to be merged, and not the other. Specifically if the first parent is an ancestor of the second parent, the conversion will identify the situation and apply the workaround. But, if instead the second parent is an ancestor of first parent, it's not detected. The first case is quite commonly achieved by merging a branch into the current branch using --no-ff, but the second seems rarer. The same workaround should apply fine in both cases however.

To verify this behavior I implemented two new tests, one for each of the possible ancestor relationships. The first test already passed, but the second was fixed by this change.

The FF.Auto case isn't tested, and is unlikely to ever be used, but I believe my change there makes it more correct than it was before.

---------
### Progress

### Issue
 * [SKARA-2002](https://bugs.openjdk.org/browse/SKARA-2002): GitToHgConverter fails on merge with ancestor (**Bug** - P2)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1549/head:pull/1549` \
`$ git checkout pull/1549`

`$ git checkout pull/1549` \
`$ git pull https://git.openjdk.org/skara.git pull/1549/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1549`

`$ git pr show -t 1549`

</details>

<a href="https://git.openjdk.org/skara/pull/1549.diff">https://git.openjdk.org/skara/pull/1549.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1549#issuecomment-1689051957)

### DIFF
--- a/vcs/src/main/java/org/openjdk/skara/vcs/hg/HgRepository.java
+++ b/vcs/src/main/java/org/openjdk/skara/vcs/hg/HgRepository.java
@@ -806,6 +806,15 @@ public class HgRepository implements Repository {
         return mergeBase(ancestor, descendant).equals(ancestor);
     }
 
+    /**
+     * Check if either a is an ancestor of b, or b is an ancestor of a.
+     * @return true if a and b are related
+     */
+    private boolean isRelated(Hash a, Hash b) throws IOException {
+        var base = mergeBase(a, b);
+        return base.equals(a) || base.equals(b);
+    }
+
     @Override
     public void rebase(Hash hash, String committerName, String committerEmail) throws IOException {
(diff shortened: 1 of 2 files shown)


zhaosongzs (approved):
Looks good!

PR #1554, opened 2023-09-06
2015: Initial run of PR IssueBot misses issue updates

zhaosongzs:
Currently, if pullRequestBot is down for more than 10 minutes and during that time, if users make changes to some issues, than the update activities would not trigger CheckWorkItem after the bot restarts.

To solve this problem, In this patch, after the bot restarts, in the initial run, pullRequestBot would make CheckWorkItem compare both prMetaData and issueMetaData.

---------
### Progress

### Issue
 * [SKARA-2015](https://bugs.openjdk.org/browse/SKARA-2015): Initial run of PR IssueBot misses issue updates (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1554/head:pull/1554` \
`$ git checkout pull/1554`

`$ git checkout pull/1554` \
`$ git pull https://git.openjdk.org/skara.git pull/1554/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1554`

`$ git pr show -t 1554`

</details>

<a href="https://git.openjdk.org/skara/pull/1554.diff">https://git.openjdk.org/skara/pull/1554.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1554#issuecomment-1708882636)

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
@@ -66,27 +66,15 @@ class CheckWorkItem extends PullRequestWorkItem {
 
     private final boolean forceUpdate;
     private final boolean spawnedFromIssueBot;
+    private final boolean initialRun;
     private final Map<String, Optional<IssueTrackerIssue>> issues = new HashMap<>();
 
     CheckWorkItem(PullRequestBot bot, String prId, Consumer<RuntimeException> errorHandler, ZonedDateTime triggerUpdatedAt,
-                  boolean needsReadyCheck, boolean forceUpdate) {
-        super(bot, prId, errorHandler, triggerUpdatedAt, needsReadyCheck);
-        this.forceUpdate = forceUpdate;
-        this.spawnedFromIssueBot = false;
-    }
-
-    CheckWorkItem(PullRequestBot bot, String prId, Consumer<RuntimeException> errorHandler, ZonedDateTime triggerUpdatedAt,
-                  boolean needsReadyCheck, boolean forceUpdate, boolean spawnedFromIssueBot) {
+                  boolean needsReadyCheck, boolean forceUpdate, boolean spawnedFromIssueBot, boolean initialRun) {
         super(bot, prId, errorHandler, triggerUpdatedAt, needsReadyCheck);
         this.forceUpdate = forceUpdate;
         this.spawnedFromIssueBot = spawnedFromIssueBot;
-    }
-
-    CheckWorkItem(PullRequestBot bot, String prId, Consumer<RuntimeException> errorHandler, ZonedDateTime triggerUpdatedAt,
-                  boolean needsReadyCheck) {
-        super(bot, prId, errorHandler, triggerUpdatedAt, needsReadyCheck);
-        this.forceUpdate = false;
-        this.spawnedFromIssueBot = false;
+        this.initialRun = initialRun;
     }
 
     private String encodeReviewer(HostUser reviewer, CensusInstance censusInstance) {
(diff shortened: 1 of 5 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -223,50 +211,54 @@ private boolean currentCheckValid(CensusInstance censusInstance, List<Comment> c
...
 
-                // triggered by issue update
-                if (spawnedFromIssueBot) {
+                // triggered by issue update or initial run when bot restarts
+                if (initialRun || spawnedFromIssueBot) {
                     var currIssueMetadata = getIssueMetadata(pr.body());
                     if (expiresAt != null) {
                         if (previousIssueMetadata.equals(currIssueMetadata) && expiresAt.isAfter(Instant.now())) {
                             log.finer("[Issue]Metadata with expiration time is still valid, not checking again");
-                            return true;
```
erikj79, line 232: I think this method would be easier to read if we kept all return `true`/`false` inside the conditional blocks. If we know we have a terminal state, lets be clear about it so the reader doesn't need to check the rest of the method.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -66,27 +66,15 @@ class CheckWorkItem extends PullRequestWorkItem {
...
     CheckWorkItem(PullRequestBot bot, String prId, Consumer<RuntimeException> errorHandler, ZonedDateTime triggerUpdatedAt,
-                  boolean needsReadyCheck, boolean forceUpdate) {
-        super(bot, prId, errorHandler, triggerUpdatedAt, needsReadyCheck);
-        this.forceUpdate = forceUpdate;
-        this.spawnedFromIssueBot = false;
-    }
-
-    CheckWorkItem(PullRequestBot bot, String prId, Consumer<RuntimeException> errorHandler, ZonedDateTime triggerUpdatedAt,
-                  boolean needsReadyCheck, boolean forceUpdate, boolean spawnedFromIssueBot) {
+                  boolean needsReadyCheck, boolean forceUpdate, boolean spawnedFromIssueBot, boolean initialRun) {
```
erikj79, line 73: Having 4 booleans in the constructor is becoming really hard to read and understand. Perhaps we should introduce factory methods instead so we only allow for valid combinations of the booleans. Something like:
```
fromIssue
fromPr
initialRun
forceUpdate
```

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -223,50 +211,54 @@ private boolean currentCheckValid(CensusInstance censusInstance, List<Comment> c
...
                     } else {
                         if (previousIssueMetadata.equals(currIssueMetadata)) {
                             log.fine("[Issue]No activity since last check, not checking again.");
-                            return true;
                         } else {
                             log.fine("[Issue]Previous metadata: " + previousIssueMetadata + " - current: " + currIssueMetadata);
+                            return false;
                         }
                     }
                     // triggered by pr updates
```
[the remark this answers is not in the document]
zhaosongzs: Yes, will fix it

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
zhaosongzs: I removed `return true` here because we couldn't return early in this block. For the initial run case, `initialRun` would be true and `spawnedFromIssueBot` would be false. We need to check both issueMetaData and PRMetaData. 

But we could add some `return true` in `if (!spawnedFromIssueBot)` block.

zhaosongzs: Or I could switch the order of `if (initialRun || spawnedFromIssueBot)` and `if (!spawnedFromIssueBot)`.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
zhaosongzs: I was thinking that if we fix SKARA-2016 in the future, then `forceUpdate` would be removed. So three booleans is reasonable.

erikj79: 3 is already a lot, especially if certain combinations aren't actually valid. I was already thinking this constructor was bad before this change. 

When reading the source in intellij, it shows the parameter name at each call site, which alleviates some of the problem, but that isn't happening when reviewing on GitHub, or when merging using other tools.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -69,14 +69,49 @@ class CheckWorkItem extends PullRequestWorkItem {
...
+     * Create CheckWorkItem spawned from IssueBot
+     */
+    public static CheckWorkItem fromIssueBot(PullRequestBot bot, String prId, Consumer<RuntimeException> errorHandler, ZonedDateTime triggerUpdatedAt) {
+        return new CheckWorkItem(bot, prId, errorHandler, triggerUpdatedAt, true, false, true, false);
+    }
+
+    /**
+     * Create Normal CheckWorkItem
+     */
+    public static CheckWorkItem normal(PullRequestBot bot, String prId, Consumer<RuntimeException> errorHandler, ZonedDateTime triggerUpdatedAt) {
```
erikj79, line 111: Maybe call it `fromWorkItem`.

erikj79 (approved):

zhaosongzs:
Thanks for the review!
/integrate

PR #1556, opened 2023-09-07
827: Copyright year differences can be regarded as a clean backport

zhaosongzs:
As many users requested, when determining whether a backport is clean, our bot should ignore copyright year difference.

---------
### Progress

### Issue
 * [SKARA-827](https://bugs.openjdk.org/browse/SKARA-827): Copyright year differences can be regarded as a clean backport (**Enhancement** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1556/head:pull/1556` \
`$ git checkout pull/1556`

`$ git checkout pull/1556` \
`$ git pull https://git.openjdk.org/skara.git pull/1556/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1556`

`$ git pr show -t 1556`

</details>

<a href="https://git.openjdk.org/skara/pull/1556.diff">https://git.openjdk.org/skara/pull/1556.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1556#issuecomment-1710785729)

### DIFF
--- a/vcs/src/main/java/org/openjdk/skara/vcs/DiffComparator.java
+++ b/vcs/src/main/java/org/openjdk/skara/vcs/DiffComparator.java
@@ -64,18 +64,18 @@ public class DiffComparator {
     }
 
     private static boolean areFuzzyEqual(Patch a, Patch b) {
-        Pattern copyRightPattern = Pattern.compile("""
+        Pattern copyrightPattern = Pattern.compile("""
                 -(.)*Copyright \\(c\\) (?:\\d|\\s|,)* Oracle and/or its affiliates\\. All rights reserved\\.
                 \\+(.)*Copyright \\(c\\) (?:\\d|\\s|,)* Oracle and/or its affiliates\\. All rights reserved\\.
                 """);
 
         var aHunks = a.asTextualPatch().hunks()
                 .stream()
-                .filter(hunk -> !copyRightPattern.matcher(hunk.toString()).find())
+                .filter(hunk -> !copyrightPattern.matcher(hunk.toString()).find())
                 .toList();
         var bHunks = b.asTextualPatch().hunks()
                 .stream()
-                .filter(hunk -> !copyRightPattern.matcher(hunk.toString()).find())
+                .filter(hunk -> !copyrightPattern.matcher(hunk.toString()).find())
                 .toList();
 
         if (aHunks.size() != bHunks.size()) {
--- a/vcs/src/main/java/org/openjdk/skara/vcs/DiffComparator.java
+++ b/vcs/src/main/java/org/openjdk/skara/vcs/DiffComparator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2020, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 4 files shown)


vcs/src/main/java/org/openjdk/skara/vcs/DiffComparator.java
```diff
@@ -63,8 +64,20 @@ public static boolean areFuzzyEqual(Diff a, Diff b) {
     }
 
     private static boolean areFuzzyEqual(Patch a, Patch b) {
-        var aHunks = a.asTextualPatch().hunks();
-        var bHunks = b.asTextualPatch().hunks();
+        Pattern copyrightPattern = Pattern.compile("""
+                -(.)*Copyright \\(c\\) (?:\\d|\\s|,)* Oracle and/or its affiliates\\. All rights reserved\\.
+                \\+(.)*Copyright \\(c\\) (?:\\d|\\s|,)* Oracle and/or its affiliates\\. All rights reserved\\.
+                """);
```
erikj79, line 70: This pattern should be compiled in a static variable to avoid recompiling it.

erikj79 (commented):
I'm not sure about changing this code without having some kind of unit level testing for it. I'm quite surprised that we don't have any. I think we need to add basic coverage either before or with this change.

You could extract some patches from real backported commits to use for these tests.

zhaosongzs:
> I'm not sure about changing this code without having some kind of unit level testing for it. I'm quite surprised that we don't have any. I think we need to add basic coverage either before or with this change.
> 
> You could extract some patches from real backported commits to use for these tests.

Sure, I didn't add some unit tests because I tested it with some real backport cases. And I agree with you, it's good to have some unit tests.

vcs/src/main/java/org/openjdk/skara/vcs/DiffComparator.java
```diff
@@ -22,12 +22,16 @@
...
-import java.io.*;
-import java.nio.file.*;
 import java.util.*;
 import java.util.regex.Pattern;
 
 public class DiffComparator {
+
+    private static final Pattern COPYRIGHT_PATTERN = Pattern.compile("""
+            -(.)*Copyright \\(c\\) (?:\\d|\\s|,)* Oracle and/or its affiliates\\. All rights reserved\\.
+            \\+(.)*Copyright \\(c\\) (?:\\d|\\s|,)* Oracle and/or its affiliates\\. All rights reserved\\.
```
erikj79, line 32: If the only difference between these lines is the first character, we should only need one line.
```suggestion
            [-+](.)*Copyright \\(c\\) (?:\\d|\\s|,)* Oracle and/or its affiliates\\. All rights reserved\\.
```

zhaosongzs: But if we only use one line like you suggested, this will introduce a problem. In the backport PR, the user could make any change to the first line of copyright.  For example, the user could change the first line to "Random line" and the backport pr would still be considered as clean backport.  

Because the hunk will be 
```
@@ -2,1 +2,1 @@
- * Copyright (c) 2020, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Random line.
```
and this hunk would match the regex pattern and be filtered out.


I think we still need two lines to make sure the user **only** changes the copyright years.

erikj79: Ah I get it now, this is a multiline pattern, not an either this line or that line. I thought you needed to tell something in the Java regex API that it was multiline, but maybe you don't if the pattern has a newline in it?

erikj79: I think this should work as is. The `Pattern.MULTILINE` and `Pattern.DOTALL` are used to change the meaning of `^`, `$` and `.` respectively to work differently with newlines. I don't think that's relevant for this pattern which contains an explicit newline. Sorry for the confusion!

erikj79 (approved):

vcs/src/main/java/org/openjdk/skara/vcs/DiffComparator.java
zhaosongzs: I think having a newline in the pattern is enough, so we don't need to tell the java regex api that it is a multiline pattern.

zhaosongzs:
Thanks for the review!
/integrate

PR #1560, opened 2023-09-22
2033: Require all bugs to be accessible before marking PR 'rfr'

zhaosongzs:
If one or more issues associated with the pull request is not accessible, then the pull request should be considered as invalid. Also, no 'rft' label will be added to this pull request and this could help prevent emails from being sent out while the pull request is associated with invalid issues.

---------
### Progress

### Issue
 * [SKARA-2033](https://bugs.openjdk.org/browse/SKARA-2033): Require all bugs to be accessible before marking PR 'rfr' (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1560/head:pull/1560` \
`$ git checkout pull/1560`

`$ git checkout pull/1560` \
`$ git pull https://git.openjdk.org/skara.git pull/1560/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1560`

`$ git pr show -t 1560`

</details>

<a href="https://git.openjdk.org/skara/pull/1560.diff">https://git.openjdk.org/skara/pull/1560.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1560#issuecomment-1734151322)

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -76,6 +76,7 @@ class CheckRun {
     private final boolean reviewCleanBackport;
     private final boolean reviewMerge;
     private final Approval approval;
+    private boolean allIssuesAccessible = true;
 
     private Duration expiresIn;
 
@@ -144,6 +145,7 @@ class CheckRun {
                     var issueTrackerIssue = workItem.issueTrackerIssue(i.shortId());
                     if (issueTrackerIssue.isEmpty()) {
                         log.info("Failed to retrieve issue " + i.id());
+                        allIssuesAccessible = false;
                         setExpiration(Duration.ofMinutes(10));
                     }
                     map.put(i, issueTrackerIssue);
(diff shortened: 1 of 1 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -409,6 +411,11 @@ private void updateCheckBuilder(CheckBuilder checkBuilder, PullRequestCheckIssue
     }
 
     private boolean updateReadyForReview(PullRequestCheckIssueVisitor visitor, List<String> additionalErrors) {
+        // All the issues must be accessible
+        if (!allIssuesAccessible) {
```
erikj79, line 415: I would like to avoid having a boolean field that gets modified at some arbitrary point during a run. We can just send the `regularIssuesMap` to this method and we can check if any of the values is an empty Optional.

zhaosongzs: Sure

erikj79 (approved):
If I understand this correctly, this will only prevent PRs not yet ready for review from being marked `rfr`, but will not remove `rfr` if an issue is suddenly unreachable. I think that is what we want for this.

zhaosongzs:
Thanks for the review!
/integrate

PR #1561, opened 2023-09-25
2040: Temporary workaround of SKARA-2025

zhaosongzs:
The final solution for [SKARA-2025](https://bugs.openjdk.org/browse/SKARA-2025) should be for us to optimize the HostedRepositoryPool#materializeClone method, but this will take some time. 

And some users would use /tag command every week, to make their experiences better, we should fix it with the temporary workaround as soon as possible. 

As Erik said in [SKARA-2025](https://bugs.openjdk.org/browse/SKARA-2025), we only need to fetch from the remote repo into local repo and the problem could be solved.

---------
### Progress

### Issue
 * [SKARA-2040](https://bugs.openjdk.org/browse/SKARA-2040): Temporary workaround of SKARA-2025 (**Sub-task** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1561/head:pull/1561` \
`$ git checkout pull/1561`

`$ git checkout pull/1561` \
`$ git pull https://git.openjdk.org/skara.git pull/1561/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1561`

`$ git pr show -t 1561`

</details>

<a href="https://git.openjdk.org/skara/pull/1561.diff">https://git.openjdk.org/skara/pull/1561.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1561#issuecomment-1736393337)

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/TagCommand.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/TagCommand.java
@@ -94,6 +94,7 @@ public class TagCommand implements CommandHandler {
             var localRepo = bot.hostedRepositoryPool()
                                .orElseThrow(() -> new IllegalStateException("Missing repository pool for PR bot"))
                                .materialize(bot.repo(), localRepoDir);
+            localRepo.fetchAll(bot.repo().authenticatedUrl());
 
             var existingTagNames = localRepo.tags().stream().map(Tag::name).collect(Collectors.toSet());
             if (existingTagNames.contains(tagName)) {
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/TagCommand.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/TagCommand.java
@@ -94,7 +94,7 @@ public class TagCommand implements CommandHandler {
             var localRepo = bot.hostedRepositoryPool()
                                .orElseThrow(() -> new IllegalStateException("Missing repository pool for PR bot"))
                                .materialize(bot.repo(), localRepoDir);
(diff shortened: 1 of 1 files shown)


zhaosongzs:
Repository#fetchAll doesn't work in this case because when cloning the repository, `git clone` typically only clones the default branch by default. However, in the fetchAll method, the `git fetch ` command includes the parameter '+refs/heads/:refs/heads/,' which will make the `git fetch` command fails.

erikj79:
> Repository#fetchAll doesn't work in this case because when cloning the repository, `git clone` typically only clones the default branch by default. However, in the fetchAll method, the `git fetch ` command includes the parameter '+refs/heads/:refs/heads/,' which will make the `git fetch` command fails.

Fetching just the commit will fix the reported problem, but it will not fix the issue of checking if the tag already exists. Maybe rewriting that check to use the remote repository instead of the local is a better solution for that.

erikj79:
> > Repository#fetchAll doesn't work in this case because when cloning the repository, `git clone` typically only clones the default branch by default. However, in the fetchAll method, the `git fetch ` command includes the parameter '+refs/heads/:refs/heads/,' which will make the `git fetch` command fails.
> 
> Fetching just the commit will fix the reported problem, but it will not fix the issue of checking if the tag already exists. Maybe rewriting that check to use the remote repository instead of the local is a better solution for that.

I was wrong. Fetching with includeTags=true will get all tags, so this should be fine.

erikj79 (approved):

zhaosongzs:
Thanks for the review!
/integrate

PR #1562, opened 2023-09-27
2045: Make maintainer approval feature compatible with dependent pr feature.

zhaosongzs:
A user reported that a dependent pr in a repo which configured with maintainer approval feature is not marked for approval. 

After investigation, I realized that the skara bot would determine whether this pr needs maintainer approval by checking whether merging into the target branch needs maintainer approval. 
  
In this case, we only configured that merging into master branch of jdk21u needs maintainer approval, however, for dependent pull requests, the target branch is pr/XXX. 

To fix this issue, we should let the skara bot be able to find the real target ref.

---------
### Progress

### Issue
 * [SKARA-2045](https://bugs.openjdk.org/browse/SKARA-2045): Make maintainer approval feature compatible with dependent pr feature. (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1562/head:pull/1562` \
`$ git checkout pull/1562`

`$ git checkout pull/1562` \
`$ git pull https://git.openjdk.org/skara.git pull/1562/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1562`

`$ git pr show -t 1562`

</details>

<a href="https://git.openjdk.org/skara/pull/1562.diff">https://git.openjdk.org/skara/pull/1562.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1562#issuecomment-1738068895)

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/ApprovalCommand.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/ApprovalCommand.java
@@ -23,6 +23,7 @@
 package org.openjdk.skara.bots.pr;
 
 import org.openjdk.skara.bots.common.SolvesTracker;
+import org.openjdk.skara.forge.PreIntegrations;
 import org.openjdk.skara.forge.PullRequest;
 import org.openjdk.skara.issuetracker.Comment;
 import org.openjdk.skara.vcs.openjdk.Issue;
@@ -48,8 +49,6 @@ public class ApprovalCommand implements CommandHandler {
 
     private static final Pattern APPROVAL_ARG_PATTERN = Pattern.compile("(([A-Za-z]+-)?[0-9]+)? ?(request|cancel)(.*?)?", Pattern.MULTILINE | Pattern.DOTALL);
 
-    private static final Pattern PRE_INTEGRATE_BRANCH_PATTERN = Pattern.compile("pr/(\\d+)");
-
     @Override
     public void handle(PullRequestBot bot, PullRequest pr, CensusInstance censusInstance, ScratchArea scratchArea, CommandInvocation command, List<Comment> allComments, PrintWriter reply) {
         if (!command.user().equals(pr.author())) {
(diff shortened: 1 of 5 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/ApprovalCommand.java
```diff
@@ -170,4 +172,15 @@ public boolean multiLine() {
...
+    public static String realTargetRef(PullRequest pr) {
+        var targetRef = pr.targetRef();
+        var matcher = PRE_INTEGRATE_BRANCH_PATTERN.matcher(targetRef);
+        if (!matcher.matches()) {
+            return targetRef;
+        }
+        String id = matcher.group(1);
+        var dependentPR = pr.repository().pullRequest(id);
+        return realTargetRef(dependentPR);
+    }
```
erikj79, line 185: There is a class `org.openjdk.skara.forge.PreIntegrations` that handles these things. I think this method should be added there (and be integrated with existing methods).

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -104,6 +105,7 @@ private CheckRun(CheckWorkItem workItem, PullRequest pr, Repository localRepo, L
                 workItem.bot.confOverrideName(),
                 workItem.bot.confOverrideRef(),
                 comments);
+        realTargetRef = ApprovalCommand.realTargetRef(pr);
```
erikj79, line 108: I would like to avoid calculating this for every pr targeting a pr/X branch. We could make this field lazy initialized in the `needsApproval` method. Something like:

```
if (approval != null) {
    if (realTargetRef == null) {
         realTargetRef = PreIntegration.realTargetRef(pr);
    }
    return approval.needsApproval(realTargetRef);
}
return false;
```

I would also add a comment on the field with something like
```
// Only set if approval is configured for the repo
```

zhaosongzs: Thanks for the suggestion. It will make this patch much better!

erikj79 (approved):

zhaosongzs:
Thanks!
/integrate

PR #1564, opened 2023-09-28
2043: Skara mistakenly tagged the wrong commit when processing a /tag command

erikj79:
The method `GitLabRepository::recentCommitComments` may sometimes return a comment associated with the wrong commit. We have seen it happen, which caused a tag to be set on the wrong commit. My theory is that this was caused by a race, which was probably made more potent due to GitLab sometimes being slow in returning up to date data. See bug for more details.

The main fix is to always verify that an event note about a recent commit comment actually matches a comment found on the commit in question. The current implementation has a quick return when only one potential commit is found.

In addition to that, I'm also reworking the implementation to be a bit more robust. @zhaosongzs found that there is an alternate API for retrieving commit comments called "discussions" and that gives us more data, notably the same ID field as an event has. This makes it possible to match an event note with a commit comment note by ID instead of by the combination of author and created_at. From this further simplifications naturally fell out.

A CommitComment returned from `GitLabRepository::addCommitComment` no longer contains a value for ID. I think this is more correct than making up a value. I verified that no code seems to be using the returned object currently, so this shouldn't impact anything.

The two new tests were run and inspected manually.

---------
### Progress

### Issue
 * [SKARA-2043](https://bugs.openjdk.org/browse/SKARA-2043): Skara mistakenly tagged the wrong commit when processing a /tag command (**Bug** - P2)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1564/head:pull/1564` \
`$ git checkout pull/1564`

`$ git checkout pull/1564` \
`$ git pull https://git.openjdk.org/skara.git pull/1564/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1564`

`$ git pr show -t 1564`

</details>

<a href="https://git.openjdk.org/skara/pull/1564.diff">https://git.openjdk.org/skara/pull/1564.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1564#issuecomment-1739807289)

### DIFF
--- a/forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabRepository.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabRepository.java
@@ -465,28 +465,55 @@ public class GitLabRepository implements HostedRepository {
                 .execute();
     }
 
+    // Handles results from both comments and discussions API
     private CommitComment toCommitComment(Hash hash, JSONValue o) {
-       var line = o.get("line").isNull()? -1 : o.get("line").asInt();
-       var path = o.get("path").isNull()? null : Path.of(o.get("path").asString());
-       // GitLab does not offer updated_at for commit comments
-       var createdAt = ZonedDateTime.parse(o.get("created_at").asString());
-       // GitLab does not offer an id for commit comments
-       var body = o.get("note").asString();
-       var user = gitLabHost.parseAuthorField(o);
-       var id = Integer.toString((hash.hex() + createdAt.toString() + user.id()).hashCode());
-       return new CommitComment(hash,
-                                path,
-                                line,
-                                id,
-                                body,
-                                gitLabHost.parseAuthorField(o),
-                                createdAt,
-                                createdAt);
+        if (o.contains("note")) {
+            var line = o.get("line").isNull() ? -1 : o.get("line").asInt();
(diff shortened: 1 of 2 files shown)


forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabRepository.java
```diff
@@ -512,26 +539,22 @@ private Set<Hash> commitsWithTitle(String commitTitle, Map<String, Set<Hash>> co
...
+        Optional<CommitComment> found = candidates.stream()
+                .flatMap(candidate -> commitComments(candidate).stream())
+                .filter(comment -> comment.id().equals(commentId))
+                .findFirst();
+        if (found.isEmpty()) {
+            log.warning("Did not find commit with title " + commitTitle + " for repository " + projectName);
         }
-
-        throw new RuntimeException("Did not find commit with title " + commitTitle + " for repository " + projectName);
+        return found;
```
zhaosongzs, line 557: Just want to confirm that if we stop throwing exceptions here, the commitComment will be handled again in the next CommitCommentsWorkItem.

zhaosongzs (approved):
Looks good.

PR #1568, opened 2023-10-12
2061: Checking for .jcheck/conf fails when pr/X branches are removed

zhaosongzs:
As Erik described in the issue, "The targetRefPRMap introduced in [SKARA-1937](https://bugs.openjdk.org/browse/SKARA-1937) isn't working well with the dependent PR feature. When a pr/X branch is removed, the map may still contains that branch which will cause getPeriodicItems to fail with exception when trying to get .jcheck/conf from the non existing branch.". 

To solve this issue, I think that every pr would only have one targetRef, so before the bot is trying to create an entry for the pr, the bot should delete any existing entry associated with that pr.

Since the operations are all about HashMap, I believe they wouldn't significantly impact the bot's performance.

---------
### Progress

### Issue
 * [SKARA-2061](https://bugs.openjdk.org/browse/SKARA-2061): Checking for .jcheck/conf fails when pr/X branches are removed (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1568/head:pull/1568` \
`$ git checkout pull/1568`

`$ git checkout pull/1568` \
`$ git pull https://git.openjdk.org/skara.git pull/1568/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1568`

`$ git pr show -t 1568`

</details>

<a href="https://git.openjdk.org/skara/pull/1568.diff">https://git.openjdk.org/skara/pull/1568.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1568#issuecomment-1759754673)

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestBot.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestBot.java
@@ -182,7 +182,7 @@ class PullRequestBot implements Bot {
                 var targetRef = pr.targetRef();
                 var prId = pr.id();
                 if (pr.isOpen()) {
-                    targetRefPRMap.keySet().forEach(key -> targetRefPRMap.get(key).remove(prId));
+                    targetRefPRMap.values().forEach(s -> s.remove(prId));
                     targetRefPRMap.computeIfAbsent(targetRef, key -> new HashSet<>()).add(prId);
                 } else {
                     if (targetRefPRMap.containsKey(targetRef)) {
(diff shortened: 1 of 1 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestBot.java
```diff
@@ -182,6 +182,7 @@ public List<WorkItem> getPeriodicItems() {
                 var targetRef = pr.targetRef();
                 var prId = pr.id();
                 if (pr.isOpen()) {
+                    targetRefPRMap.keySet().forEach(key -> targetRefPRMap.get(key).remove(prId));
```
erikj79, line 185: I think this could be simplified by just iterating over `targetRefPRMap.values()`.

zhaosongzs: Yes, it should be better.

erikj79 (approved):

zhaosongzs:
Thanks for the review!
/integrate

PR #1569, opened 2023-10-12
2063: Reduce the health check times of local repo instance

zhaosongzs:
In the method HostedRepositoryPool#materializeClone, when the bot is trying to reuse a local repo instance, the bot will always check whether the repo is still good by processing command "git fsck --connectivity-only".

Sometimes this command would be slow and Erik said that we should believe the bots are doing right things, so we should assume the local repos are good, so the health check is not always needed. But shutdown of the bot could make a local repo instance unhealthy, so we should at least do the health check once for each local repo instance after the bot is restarted.

To solve this issue, Erik suggested to maintain a static map for keeping track of paths of known checked/good repositories.

To leverage this patch to mlbridge bot, I also made some changes to WebrevStorage.

---------
### Progress

### Issue
 * [SKARA-2063](https://bugs.openjdk.org/browse/SKARA-2063): Reduce the health check times of local repo instance (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1569/head:pull/1569` \
`$ git checkout pull/1569`

`$ git checkout pull/1569` \
`$ git pull https://git.openjdk.org/skara.git pull/1569/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1569`

`$ git pr show -t 1569`

</details>

<a href="https://git.openjdk.org/skara/pull/1569.diff">https://git.openjdk.org/skara/pull/1569.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1569#issuecomment-1760382020)

### DIFF
--- a/bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/WebrevStorage.java
+++ b/bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/WebrevStorage.java
@@ -284,7 +284,8 @@ class WebrevStorage {
         throw new RuntimeException("No success response from " + uri + " within " + timeout);
     }
 
-    private URI createAndArchive(PullRequest pr, Repository localRepository, Path scratchPath, Diff diff, Hash base, Hash head, String identifier) {
+    private URI createAndArchive(PullRequest pr, Repository localRepository, Path scratchPath, Diff diff, Hash base, Hash head, String identifier,
+                                 Repository jsonLocalStorage, Repository htmlLocalStorage) {
         try {
             if (!generateHTML && !generateJSON) {
                 return null;
@@ -295,8 +296,6 @@ class WebrevStorage {
             URI uri = null;
 
             if (generateJSON) {
-                var jsonLocalStorage = Repository.materialize(scratchPath, jsonStorage.authenticatedUrl(),
-                                                              "+" + storageRef + ":mlbridge_webrevs");
                 if (Files.exists(outputFolder)) {
                     clearDirectory(outputFolder);
                 }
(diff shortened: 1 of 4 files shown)


forge/src/main/java/org/openjdk/skara/forge/HostedRepositoryPool.java
```diff
@@ -147,7 +148,7 @@ private Repository materializeClone(Path path, boolean allowStale, boolean bare)
                 return cloneSeeded(path, allowStale, bare);
             } else {
                 var localRepoInstance = localRepo.get();
-                if (!localRepoInstance.isHealthy()) {
+                if (!isHealthy(localRepoInstance, path)) {
                     removeOldClone(path, "unhealthy");
                     return cloneSeeded(path, allowStale, bare);
```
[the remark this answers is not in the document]
zhaosongzs: Yes, good idea.

bots/mlbridge/src/main/java/org/openjdk/skara/bots/mlbridge/WebrevStorage.java
```diff
@@ -322,26 +322,45 @@ private URI createAndArchive(PullRequest pr, Repository localRepository, Path sc
...
         return new WebrevGenerator() {
+            Repository jsonLocalStorage = null;
+            Repository htmlLocalStorage = null;
+
+            private void initializeLocalStorage() {
+                if (jsonLocalStorage == null && !(jsonStorage == null)) {
+                    try {
+                        jsonLocalStorage = hostedRepositoryPool.checkout(jsonStorage, storageRef, scratchPath);
+                    } catch (IOException e) {
+                        throw new RuntimeException(e);
```
erikj79, line 342: This would throw `UncheckedIOException` before so should still do that. Can also consider just having one try-catch in the method, surrounding all of it, to avoid repeating the catch clause.

zhaosongzs: You are right, thanks for catching it!

erikj79 (approved):

zhaosongzs:
Thanks for the review and suggestions!
/integrate

PR #1575, opened 2023-10-24
1962: JiraProject#queryIssues may return duplicate issues

zhaosongzs:
Since [SKARA-1912](https://bugs.openjdk.org/browse/SKARA-1912) deployed into prod, we have noticed occurrences of duplicate key exceptions in the IssueProjectPoller#updatedIssues.

After investigation, we found that `JiraProject#queryIssues` may return duplicate issues and I think the pagination in this method is the root cause.

The problem here is that, sometimes, issues will be updated during the pagination process. For example, the default page size is 50 and we assume that issue#50 is the last issue in the first page. And before querying the second page, another issue which not in the first page got updated, so issue#50 will also be the first issue in the second page. Therefore, the method may return duplicate issues.

Erik's solution is right. We just need to order by updatedTime DESC and remove the duplicate issues. Ordering by updatedTime DESC would make sure that even some issues are updated during the pagination, the issues will be processed in the future round.

---------
### Progress

### Issue
 * [SKARA-1962](https://bugs.openjdk.org/browse/SKARA-1962): JiraProject#queryIssues may return duplicate issues (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1575/head:pull/1575` \
`$ git checkout pull/1575`

`$ git checkout pull/1575` \
`$ git pull https://git.openjdk.org/skara.git pull/1575/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1575`

`$ git pr show -t 1575`

</details>

<a href="https://git.openjdk.org/skara/pull/1575.diff">https://git.openjdk.org/skara/pull/1575.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1575#issuecomment-1777738411)

### DIFF
--- a/issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraProject.java
+++ b/issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraProject.java
@@ -473,7 +473,7 @@ public class JiraProject implements IssueProject {
     @Override
     public List<IssueTrackerIssue> issues(ZonedDateTime updatedAfter) {
         var timeString = toTimeString(updatedAfter);
-        var jql = "project = " + projectName + " AND updated >= '" + timeString + "'";
+        var jql = "project = " + projectName + " AND updated >= '" + timeString + "' ORDER BY updated DESC";
         return queryIssues(jql);
     }
 
@@ -481,7 +481,7 @@ public class JiraProject implements IssueProject {
     @Override
     public List<IssueTrackerIssue> csrIssues(ZonedDateTime updatedAfter) {
         var timeString = toTimeString(updatedAfter);
-        var jql = "project = " + projectName + " AND updated >= '" + timeString + "' AND issuetype = CSR";
+        var jql = "project = " + projectName + " AND updated >= '" + timeString + "' AND issuetype = CSR ORDER BY updated DESC";
         return queryIssues(jql);
     }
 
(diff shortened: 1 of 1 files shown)


issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraProject.java
```diff
@@ -506,18 +506,20 @@ private String toTimeString(ZonedDateTime time) {
         return timeZoned.format(DateTimeFormatter.ofPattern("yyyy/MM/dd HH:mm"));
     }
 
-    private ArrayList<IssueTrackerIssue> queryIssues(String jql) {
-        var ret = new ArrayList<IssueTrackerIssue>();
+    private List<IssueTrackerIssue> queryIssues(String jql) {
+        var ret = new HashMap<String, IssueTrackerIssue>();
```
erikj79, line 510: I think we should try to preserve the order we received issues. We can do that if we use a `LinkedHashMap` here. It would also make more sense to me to use ascending order. Then we will include any concurrently updated issue in this call instead of waiting for next call. I also think it makes sense for the bot to process issues in ascending order, so the first updated issue is processed first.

erikj79: After offline discussion, I take this back, descending order is the correct solution. We could still preserve the order here, or maybe just sort the result on "updated" before returning it.

erikj79: Or just skip sorting.

erikj79 (approved):

erikj79 (commented):
I think we need a comment explaining why we need the sorting order.

zhaosongzs:
> I think we need a comment explaining why we need the sorting order.

Sure

erikj79 (approved):

zhaosongzs:
Thanks for the review!
/integrate

PR #1578, opened 2023-10-27
2082

edvbld:
Hi all,

please review this patch that ensures that we pass the correct `.jcheck/conf` to [AdditionalConfiguration.java](https://github.com/openjdk/skara/blob/master/bots/pr/src/main/java/org/openjdk/skara/bots/pr/AdditionalConfiguration.java). The `.jcheck/conf` should either come from the "overriding" `.jcheck/conf` (if the PR bot is configured to use that) or the pull request's target branch.

I have added that test that passes with this patch but fails without it.

Thanks,
Erik


bots/pr/src/main/java/org/openjdk/skara/bots/pr/AdditionalConfiguration.java
```diff
@@ -31,14 +31,14 @@
 import java.util.*;
 
 public class AdditionalConfiguration {
-    static List<String> get(ReadOnlyRepository repository, Hash hash, HostUser botUser, List<Comment> comments, boolean reviewMerge) throws IOException {
+    static List<String> get(Optional<JCheckConfiguration> conf, HostUser botUser, List<Comment> comments, boolean reviewMerge) throws IOException {
```
erikj79, line 34: Using Optional as parameter isn't recommended, especially since we throw if it's empty. Is it truly optional here? I would think we should just `.orElseThrow()` at the parse call.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -1265,9 +1267,10 @@ private void checkStatus() {
             } else {
                 // Determine current status
                 jcheckType = "jcheck";
-                var additionalConfiguration = AdditionalConfiguration.get(localRepo, localHash,
+                var conf = checkablePullRequest.parseJCheckConfiguration(localRepo, jcheckConfHash);
+                var additionalConfiguration = AdditionalConfiguration.get(conf,
                         pr.repository().forge().currentUser(), comments, reviewMerge);
-                checkablePullRequest.executeChecks(localHash, censusInstance, visitor, additionalConfiguration, checkablePullRequest.targetHash());
+                checkablePullRequest.executeChecks(localHash, censusInstance, visitor, additionalConfiguration, jcheckConfHash);
                 // Don't need to run the second round if confOverride is set.
```
erikj79, line 1274: Strictly speaking, should we regenerate the AdditionalConfiguration for the second run where we use the merged .jcheck/conf?

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -1254,7 +1254,9 @@ private void checkStatus() {
                 additionalErrors = List.of(e.getMessage());
                 localHash = baseHash;
             }
-            PullRequestCheckIssueVisitor visitor = checkablePullRequest.createVisitor(checkablePullRequest.targetHash());
+
+            Hash jcheckConfHash = checkablePullRequest.targetHash();
+            PullRequestCheckIssueVisitor visitor = checkablePullRequest.createVisitor(jcheckConfHash);
```
erikj79, line 1259: Any particular reason to spell out the types here?

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
zhaosongzs: Oh, right, I overlooked it when I was implementing the second run of jcheck feature.

edvbld:
Ok, so I spent quite a bit of time going through the places where we parse `.jcheck/conf`. I figured out the following:

- for merge PRs we want to always test each commit in the PR with _both_ the target branch's `.jcheck/conf` and, if any commit in the PR updates `.jcheck/conf`, the `.jcheck/conf` present in the commit. This could be optimized to only test with the `.jcheck/conf` in the commit if the current commit, or any parents of the current commit, updated `.jcheck/conf`. But that seemed like a premature optimization given that the number of merge PR that changes `.jcheck/conf` most likely will be very few.
- for regular pull requests we want to ensure that we are always using the `.jcheck/conf` from the target branch _with_ any additional configuration added (or using an overriding `.jcheck/conf` is that is set up). To enable this I moved this logic into `CheckablePullRequest.parseJCheckConfiguration`.

The remainder of the patch is dealing with fall-out from the above two changes and aligning the code with them.

Thanks,
Erik

PS. The tests failed due to JBS being down

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
edvbld: Nope, thanks for catching

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
edvbld: This has been re-worked in the latest version

bots/pr/src/main/java/org/openjdk/skara/bots/pr/AdditionalConfiguration.java
edvbld: Thanks, removed the use of `Optional` in the latest version.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckablePullRequest.java
```diff
@@ -202,25 +204,30 @@ Hash amendManualReviewers(Hash commit, Namespace namespace, Hash original) throw
...
-                       List<String> additionalConfiguration, Hash hash) throws IOException {
-        Optional<JCheckConfiguration> conf;
-        if (confOverride != null) {
-            conf = JCheck.parseConfiguration(confOverride, additionalConfiguration);
-        } else {
-            conf = JCheck.parseConfiguration(localRepo, hash, additionalConfiguration);
-        }
-        if (conf.isEmpty()) {
-            throw new RuntimeException("Failed to parse jcheck configuration at: " + hash + " with extra: " + additionalConfiguration);
+    Optional<JCheckConfiguration> parseJCheckConfiguration(Hash hash) throws IOException {
```
erikj79, line 212: Every caller of this method is throwing IllegalStateException if jcheck conf isn't found. Did you consider just throwing here and not return Optional? I realize there is a bit more information in the exception message, but on the other hand, whoever reads it will have a stacktrace anyway.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckablePullRequest.java
```diff
@@ -202,25 +204,30 @@ Hash amendManualReviewers(Hash commit, Namespace namespace, Hash original) throw
...
-                                       conf.get())) {
+
+        var botUser = pr.repository().forge().currentUser();
+        var additional = AdditionalConfiguration.get(original.get(), botUser, comments, reviewMerge);
+        return confOverride == null ?
+            JCheck.parseConfiguration(localRepo, hash, additional) :
+            JCheck.parseConfiguration(confOverride, additional);
+    }
+
+    void executeChecks(Hash localHash, CensusInstance censusInstance, PullRequestCheckIssueVisitor visitor, JCheckConfiguration conf) throws IOException {
```
erikj79, line 228: The parameter name `localHash` threw me off for a bit, maybe just `hash` in this context, unless I'm missing some context?

erikj79 (commented):
Overall I really like this change. It makes dealing with jcheck configs a lot clearer.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -1209,25 +1208,40 @@ private void checkStatus() {
...
                         jcheckType = "merge jcheck in commit " + hash.hex();
-                        PullRequestCheckIssueVisitor visitor = checkablePullRequest.createVisitor(hash);
-                        checkablePullRequest.executeChecks(hash, censusInstance, visitor, List.of(), hash);
-                        mergeJCheckMessage.addAll(visitor.messages().stream()
+                        var targetVisitor = checkablePullRequest.createVisitor(targetJCheckConf, hash);
+                        checkablePullRequest.executeChecks(hash, censusInstance, targetVisitor, targetJCheckConf);
+
+                        var commitJCheckConf = checkablePullRequest.parseJCheckConfiguration(hash)
+                                                .orElseThrow(() -> new IllegalStateException("No .jcheck/conf present in tree for commit " + hash));
+                        var commitVisitor = checkablePullRequest.createVisitor(commitJCheckConf, hash);
```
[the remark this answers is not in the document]
erikj79: Maybe we shouldn't concat the streams and instead differentiate the message between the different jcheck runs and include information about which jcheck conf was used.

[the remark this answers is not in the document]
zhaosongzs: I agree with Erik here, it's better to include the jcheck conf information

zhaosongzs (commented):
This patch looks pretty good, just some minor issues to address.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -1254,8 +1268,9 @@ private void checkStatus() {
                 additionalErrors = List.of(e.getMessage());
                 localHash = baseHash;
             }
-            PullRequestCheckIssueVisitor visitor = checkablePullRequest.createVisitor(checkablePullRequest.targetHash());
-            boolean needUpdateAdditionalProgresses = false;
+
+            var visitor = checkablePullRequest.createVisitor(targetJCheckConf, localHash);
```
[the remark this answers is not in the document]
edvbld: I agree, fixed in latest version

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckablePullRequest.java
edvbld: Sure, renamed it t to `hash` in the latest version

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckablePullRequest.java
```diff
@@ -202,25 +204,30 @@ Hash amendManualReviewers(Hash commit, Namespace namespace, Hash original) throw
...
         }
-        visitor.setConfiguration(conf.get());
-        try (var issues = JCheck.check(localRepo, censusInstance.census(), CommitMessageParsers.v1, localHash,
-                                       conf.get())) {
+
+        var botUser = pr.repository().forge().currentUser();
+        var additional = AdditionalConfiguration.get(original.get(), botUser, comments, reviewMerge);
+        return confOverride == null ?
+            JCheck.parseConfiguration(localRepo, hash, additional) :
+            JCheck.parseConfiguration(confOverride, additional);
```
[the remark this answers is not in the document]
edvbld: Good idea, fixed in most recent change.

jcheck/src/main/java/org/openjdk/skara/jcheck/JCheck.java
```diff
@@ -295,4 +295,16 @@ public static Set<Check> checksFor(ReadOnlyRepository repository, Hash hash) thr
...
+        var jcheck = new JCheck(repository,
+                                CommitMessageParsers.v1,
+                                repository.range(hash),
+                                Pattern.compile(".*"),
+                                Pattern.compile(".*"),
+                                List.of(),
+                                conf,
+                                null);
+        return jcheck.checksForRange();
+    }
```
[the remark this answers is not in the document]
edvbld: I think I would like to keep it for the tests since those tests will be exercising slightly different code paths (especially with the most recent version)

edvbld:
@erikj79 and @zhaosongzs I'm finally looping back to this and _think_ I addressed all or your comments and suggestions. Please have a look at this PR again, I think it is shaping up quite niceliy.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -1275,27 +1294,30 @@ private void checkStatus() {
...
-                    checkablePullRequest.executeChecks(localHash, censusInstance, visitor2, additionalConfiguration, pr.headHash());
-                    secondJCheckMessage.addAll(visitor2.messages().stream()
+                jcheckType = "target jcheck";
+                checkablePullRequest.executeChecks(localHash, censusInstance, visitor, targetJCheckConf);
+
+                // If the PR updates .jcheck/conf then Need to run JCheck again using the configuration
+                // from the resulting commit. Not needed if we are overriding the JCheck configuration since
+                // then we won't use the one in the repo anyway.
+                if (workItem.bot.confOverrideRepository().isEmpty() &&
+                    (isFileUpdated(Path.of(".jcheck").resolve("conf"), localHash) || isJCheckConfUpdatedInMergePR)) {
```
erikj79, line 1304: Style nit. I think this way of creating a path looks a bit overly complex. I would suggest either just `Path.of(".jcheck/conf")` or split the elements like this `Path.of(".jcheck", "conf")`.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -1219,30 +1218,49 @@ private void checkStatus() {
...
+                        mergeJCheckMessageWithTargetConf.addAll(targetVisitor.messages().stream()
                                 .map(StringBuilder::new)
-                                .map(e -> e.append(" (in commit ").append(hash.hex()).append(")"))
+                                .map(e -> e.append(" (in commit `").append(hash.hex()).append("` with target configuration)"))
                                 .map(StringBuilder::toString)
                                 .toList());
+
+                        var commitJCheckConf = checkablePullRequest.parseJCheckConfiguration(hash);
+                        var commitVisitor = checkablePullRequest.createVisitor(commitJCheckConf);
+                        if (isJCheckConfUpdatedInMergePR) {
```
erikj79, line 1254: Shouldn't this be conditional on the repo using an external jcheck conf?

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -1275,27 +1294,30 @@ private void checkStatus() {
...
+
+                // If the PR updates .jcheck/conf then Need to run JCheck again using the configuration
+                // from the resulting commit. Not needed if we are overriding the JCheck configuration since
+                // then we won't use the one in the repo anyway.
+                if (workItem.bot.confOverrideRepository().isEmpty() &&
+                    (isFileUpdated(Path.of(".jcheck").resolve("conf"), localHash) || isJCheckConfUpdatedInMergePR)) {
+                    jcheckType = "source jcheck";
+                    var localJCheckConf = checkablePullRequest.parseJCheckConfiguration(localHash);
+                    var localVisitor = checkablePullRequest.createVisitor(localJCheckConf);
+                    log.info("Run jcheck against localHash with JCHeck configuration from localHash");
```
erikj79, line 1308:
```suggestion
                    log.info("Run JCheck against localHash with configuration from localHash");
```

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -1275,27 +1294,30 @@ private void checkStatus() {
...
+                            .map(e -> e.append(" (failed with updated jcheck configuration in pull request)"))
                             .map(StringBuilder::toString)
                             .toList());
                 }
                 additionalErrors = botSpecificChecks(isCleanBackport);
                 needUpdateAdditionalProgresses = true;
             }
 
-            var confFile = localRepo.lines(Path.of(".jcheck/conf"), localHash);
+            var confFile = localRepo.lines(Path.of(".jcheck").resolve("conf"), localHash);
```
erikj79, line 1320: Same style nit as above.

erikj79 (commented):
I like how this is shaping out. Just some minor issues left.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
edvbld: Fixed in latest

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
edvbld: Fixed in latest

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
edvbld: Sure, fixed in latest

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -1219,30 +1218,49 @@ private void checkStatus() {
...
+                        checkablePullRequest.executeChecks(hash, censusInstance, targetVisitor, targetJCheckConf);
+                        mergeJCheckMessageWithTargetConf.addAll(targetVisitor.messages().stream()
                                 .map(StringBuilder::new)
-                                .map(e -> e.append(" (in commit ").append(hash.hex()).append(")"))
+                                .map(e -> e.append(" (in commit `").append(hash.hex()).append("` with target configuration)"))
                                 .map(StringBuilder::toString)
                                 .toList());
+
+                        var commitJCheckConf = checkablePullRequest.parseJCheckConfiguration(hash);
+                        var commitVisitor = checkablePullRequest.createVisitor(commitJCheckConf);
```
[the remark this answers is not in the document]
edvbld: Agree, fixed in latest

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
erikj79: This one wasn't changed. :)

erikj79 (approved):

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
edvbld: Now it is 🙈

edvbld:
Thanks @erikj79 for the thorough review! @zhaosongzs would you mind taking a look as well?

zhaosongzs:
> Thanks @erikj79 for the thorough review! @zhaosongzs would you mind taking a look as well?

I reviewed your latest patch yesterday and thought it pretty good. However, I hesitated to approve it because I thought ErikJ might discover some minor issues.

zhaosongzs (approved):
Looks very good!

erikj79 (approved):

edvbld:
Thanks @erikj79 and @zhaosongzs for reviewing!

/integrate

PR #1583, opened 2023-11-03
2084: Add a `/branch` command similar to `/tag` to create a new branch from a commit

edvbld:
Hi all,

please review this patch that adds a `/branch` commit command. The `/branch` commit command allows an integrator to create a _new_ branch in a repository by adding the comment `/branch <name>` on a commit's web page. The command will check that no branch with the given name exists (i.e. it won't update a branch with that name to that commit).

The implementation is very similar to the `/tag` command, but I don't think it makes sense to try to share code between them. They are both only around 60 lines of code, so trying to share code won't give us much.

I also added a couple of new unit tests.

Thanks,
Erik

---------
### Progress

### Issue
 * [SKARA-2084](https://bugs.openjdk.org/browse/SKARA-2084): Add a `/branch` command similar to `/tag` to create a new branch from a commit (**Enhancement** - P3)

### Reviewers
 * [Zhao Song](https://openjdk.org/census#zsong) (@zhaosongzs - **Reviewer**) ⚠️ Review applies to [d32d69ba](https://git.openjdk.org/skara/pull/1583/files/d32d69ba9df903e6789468fc250aaf0f3cb8d89e)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1583/head:pull/1583` \
`$ git checkout pull/1583`

`$ git checkout pull/1583` \
`$ git pull https://git.openjdk.org/skara.git pull/1583/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1583`

`$ git pr show -t 1583`

</details>

<a href="https://git.openjdk.org/skara/pull/1583.diff">https://git.openjdk.org/skara/pull/1583.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1583#issuecomment-1792923147)

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/BranchCommand.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/BranchCommand.java
@@ -119,8 +119,8 @@ public class BranchCommand implements CommandHandler {
             }
 
             var jcheckConf = JCheckConfiguration.from(localRepo, commit.hash());
-            var branchPattern = jcheckConf.isPresent() ? jcheckConf.get().repository().branches() : null;
-            if (branchPattern != null && !branchName.matches(branchPattern)) {
+            var branchPattern = jcheckConf.map(c -> c.repository().branches());
+            if (branchPattern.isPresent() && !branchName.matches(branchPattern.get())) {
                 reply.println("The given branch name `" + branchName + "` is not of the form `" + branchPattern + "`.");
                 return;
             }
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/BranchCommand.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/BranchCommand.java
@@ -1,0 +1,133 @@
+/*
+ * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package org.openjdk.skara.bots.pr;
+
+import org.openjdk.skara.forge.HostedCommit;
+import org.openjdk.skara.issuetracker.Comment;
+import org.openjdk.skara.vcs.*;
+import org.openjdk.skara.jcheck.JCheckConfiguration;
+
+import java.io.PrintWriter;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.util.List;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+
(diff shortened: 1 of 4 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/BranchCommand.java
```diff
@@ -0,0 +1,133 @@
...
+            var branchName = parts[0];
+
+            var localRepoDir = scratchArea.get(this)
+                    .resolve(bot.repo().name());
+            var localRepo = bot.hostedRepositoryPool()
+                               .orElseThrow(() -> new IllegalStateException("Missing repository pool for PR bot"))
+                               .materialize(bot.repo(), localRepoDir);
+            localRepo.fetch(bot.repo().authenticatedUrl(), commit.hash().toString(), true);
+
+            var remoteBranches = localRepo.remoteBranches(bot.repo().authenticatedUrl().toString());
```
erikj79, line 102: I'm not sure this will contain all branches. We have had problems with repos materialized from HostedRepositoryPool, probably a bug in there (specifically in refreshSeed). Most uses fetch explicitly after to work around it, which you do here, but this usecase is different because we need every branch to be present, not just the main commit.

Instead of working around it here, it would be better to fix HostedRepositoryPool before integrating this.

If we could trust materialize, then the fetch after wouldn't be needed.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/BranchCommand.java
```diff
@@ -0,0 +1,133 @@
...
+            var branchPattern = jcheckConf.isPresent() ? jcheckConf.get().repository().branches() : null;
+            if (branchPattern != null && !branchName.matches(branchPattern)) {
+                reply.println("The given branch name `" + branchName + "` is not of the form `" + branchPattern + "`.");
+                return;
+            }
+
+            var branch = localRepo.branch(commit.hash(), branchName);
+            log.info("Pushing branch '" + branch + "' to refer to commit: " + commit.hash().hex());
+            localRepo.push(commit.hash(), bot.repo().authenticatedUrl(), branch.name(), false, false);
+            log.info(String.join(", ", localRepo.branches().stream().map(Branch::name).collect(Collectors.toList())));
```
erikj79, line 127: This will potentially be a rather long log message. Not sure we need to list all branches in the repo.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/BranchCommand.java
```diff
@@ -0,0 +1,133 @@
...
+
+            var remoteBranches = localRepo.remoteBranches(bot.repo().authenticatedUrl().toString());
+            var remoteBranchNames = remoteBranches.stream()
+                                                  .map(Reference::name)
+                                                  .collect(Collectors.toSet());
+            if (remoteBranchNames.contains(branchName)) {
+                var ref = remoteBranches.stream().filter(r -> r.name().equals(branchName)).findFirst();
+                if (ref.isEmpty()) {
+                    throw new IllegalStateException("Cannot find hash for remote branch '" + branchName + "'");
+                }
```
erikj79, line 110: I think this turned out a little awkward, especially the exception that should never actually happen.
```suggestion
            var existingBranch = remoteBranches.stream()
                    .filter(b -> b.name().equals(branchName))
                    .findAny();
            if (existingBranch.isPresent()) {
```

bots/pr/src/main/java/org/openjdk/skara/bots/pr/BranchCommand.java
edvbld: Yeah, sorry, this was a left-over debug message.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/BranchCommand.java
edvbld: Sure, I rewrote it a bit in the latest version.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/BranchCommand.java
edvbld: We actually don't need the `fetch` (we are about to push a new ref), I just left it there to "follow the pattern". I agree that we should fix the problem `HostedRepositoryPool`, but I would prefer to integrate this first, just to get if off my plate (there is no reason to hold this PR hostage over that bug 😄). 

I opted to use `HostedRepository::branches` in the latest version which hopefully will make the code a bit clearer.

erikj79 (approved):

zhaosongzs (approved):

erikj79 (approved):

PR #1584, opened 2023-11-08
2101: Change confusing requirement text

magicus:
The checkbox list in "Progress" in the PRs body says:

"Commit message must refer to an issue"

but it is not really talking about the commit message, but the PR title. The "refer to an issue" is also vague; especially since the only time you will be hit by this problem is when you are a newcomer, and do not understand what is expected of you.

The text should be rephrased to be more helpful to newcomers.

---------
### Progress
- [ ] Change must be properly reviewed (2 reviews required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer), 1 [Author](https://openjdk.org/bylaws#author))

### Issue
 * [SKARA-2101](https://bugs.openjdk.org/browse/SKARA-2101): Change confusing requirement text (**Bug** - P5)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1584/head:pull/1584` \
`$ git checkout pull/1584`

`$ git checkout pull/1584` \
`$ git pull https://git.openjdk.org/skara.git pull/1584/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1584`

`$ git pr show -t 1584`

</details>

<a href="https://git.openjdk.org/skara/pull/1584.diff">https://git.openjdk.org/skara/pull/1584.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1584#issuecomment-1802360731)

### DIFF
--- a/jcheck/src/main/java/org/openjdk/skara/jcheck/IssuesCheck.java
+++ b/jcheck/src/main/java/org/openjdk/skara/jcheck/IssuesCheck.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -68,6 +68,6 @@ public class IssuesCheck extends CommitCheck {
 
     @Override
     public String description() {
-        return "Commit message must refer to an issue";
+        return "PR title must begin with a JBS issue number";
     }
 }


zhaosongzs (approved):
Looks good to me, but it's better to get approval from Erik.
/reviewers 2

erikj79:
If you run the command line jcheck and get this warning, the new message will look very confusing. The problem here is that we are running jcheck, which is it's own thing as part of validating a PR and posting the jcheck errors directly in the PR body. Even if running jcheck in a PR is more common than running the cmdline, I don't think just changing the message like this is a good idea.

magicus:
Oh, I see. That makes it quite more tricky to resolve without resorting to ugly hacks. :-(

Back to the drawing board, then...

magicus:
This simple attempt did not work. Closing this PR.

magicus:
People are still getting confused on this. I would argue that the potential confusion factor for someone who runs jcheck locally is magnitudes lower than the confusion factor when the bots run jcheck on open PRs. So while integrating this PR will create a slight regression for people running jcheck locally, I think it is acceptable compared to the win.

If you run it locally, 
1) you are an expert user,
2) you *know* you don't have a PR, so you would most likely think "well, Skara, that was a bit incorrect about you, but I get the point".

For jcheck run on PRs,
1) several users have no knowledge about how jcheck and Skara bots work,
2) you *do* have commits, and you worry that you might need to force-push an updated commit, which from the warning, seems like what is requested.

@erikj79 I'd ask you to reconsider and accept this as a patch. At least as a temporary improvement; if you want we can file a follow-up bug to make the much more complicated solution that can tell the difference between jcheck being run locally or by bots.

PR #1585, opened 2023-11-08
2089: Warn when bug fixVersion doesn't match .jcheck/conf version in PR

zhaosongzs:
This patch tries to alert the user when the fixVersion in the issue differs from the fixVersion in .jcheck/conf, thereby preventing the creation of unwanted backports.

---------
### Progress

### Issue
 * [SKARA-2089](https://bugs.openjdk.org/browse/SKARA-2089): Warn when bug fixVersion doesn't match .jcheck/conf version in PR (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1585/head:pull/1585` \
`$ git checkout pull/1585`

`$ git checkout pull/1585` \
`$ git pull https://git.openjdk.org/skara.git pull/1585/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1585`

`$ git pr show -t 1585`

</details>

<a href="https://git.openjdk.org/skara/pull/1585.diff">https://git.openjdk.org/skara/pull/1585.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1585#issuecomment-1802886998)

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestBotFactory.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestBotFactory.java
@@ -244,8 +244,8 @@ public class PullRequestBotFactory implements BotFactory {
                 botBuilder.approval(approval);
             }
 
-            if (repo.value().contains("fixVersionOverride")) {
-                botBuilder.fixVersionOverride(repo.value().get("fixVersionOverride").asBoolean());
+            if (repo.value().contains("versionMismatchWarning")) {
+                botBuilder.versionMismatchWarning(repo.value().get("versionMismatchWarning").asBoolean());
             }
 
             var prBot = botBuilder.build();
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestBotFactory.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestBotFactory.java
@@ -244,6 +244,10 @@ public class PullRequestBotFactory implements BotFactory {
                 botBuilder.approval(approval);
             }
 
+            if (repo.value().contains("fixVersionOverride")) {
+                botBuilder.fixVersionOverride(repo.value().get("fixVersionOverride").asBoolean());
+            }
+
             var prBot = botBuilder.build();
             pullRequestBotMap.put(repository.name(), prBot);
             ret.add(prBot);
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -727,7 +727,7 @@ class CheckRun {
                             }
                             progressBody.append(")");
                         }
-                        if (!workItem.bot.fixVersionOverride() && issueTrackerIssue.get().isOpen()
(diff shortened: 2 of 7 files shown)


kevinrushforth:
There are valid cases where this could occur, so it might generate a bit of noise.

I have a few questions:
* Does this correctly handle a fix version of "tbd", or "NN-pool" or come other "pseudo" version? Those should be considered matching. 
* When you say ".jcheck/conf" version, what about repos that are configured to use a fix version from the Skara configuration?

I think the main case you want to handle are in repos (e.g., mainline jdk) where the `.jcheck/conf` version is used, and has a value of NN and the JBS mainline fix version has a concrete value of "MM". Anything else seems likely to generate more noise than be hepful.

erikj79:
> I have a few questions:
> 
> * Does this correctly handle a fix version of "tbd", or "NN-pool" or come other "pseudo" version? Those should be considered matching.
Yes, it's calling the same logic as the Notifier uses to find an existing backport record. If it finds a match, no warning is printed. If the main bug is not open, no warning is printed. I think this is doing what we want.

> * When you say ".jcheck/conf" version, what about repos that are configured to use a fix version from the Skara configuration?
The Skara configuration for fixVersion is only present in the notifier bot, so the PR bot doesn't know (unless we add this configuration option, which would be awkward to maintain for the admin). The logic here will only try to look for a mismatch if the fixVersion can be found through .jcheck/conf. This means that if a repo has a .jcheck/conf with a version defined, and we configure the Notifier with a different fixVersion, then this warning will be printed when it shouldn't be. We need some kind of configuration to prevent this. Another boolean flag in the configuration for PR bot.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestBotFactory.java
```diff
@@ -244,6 +244,10 @@ public List<Bot> create(BotConfiguration configuration) {
                 botBuilder.approval(approval);
             }
 
+            if (repo.value().contains("fixVersionOverride")) {
```
erikj79, line 247: I think the config option should reflect that we are turning off this warning specifically, rather than that we have configured a different fixVersion in the notifier. There are more situations where this warning doesn't apply (see bug comments).

How about `versionMismatchWarning` and I think we should even default it to `false` and selectively add it to the repos we want it on.

zhaosongzs: Sure, thanks

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestBotFactory.java
```diff
@@ -244,6 +244,10 @@ public List<Bot> create(BotConfiguration configuration) {
                 botBuilder.approval(approval);
             }
 
+            if (repo.value().contains("versionMismatchWarning")) {
```
erikj79, line 247: Can you add this to the factory tests?

zhaosongzs: Sure

erikj79 (approved):

zhaosongzs:
Thanks for the review!
/integrate

PR #1600, opened 2024-01-16
2143: Support selective tags mirroring

erikj79:
We need a way to mirror tags selectively. Here is how I think I want to solve this. We currently have

"branches": Is either a string of one branch pattern or an array of multiple branch patterns
"tags": Takes either "include" or "only", if value is "only", "branches" cannot also be set

Instead of trying to somehow bend the "tags" parameter into taking some kind of patterns, I want to introduce a new parameter "refspecs". The value is an array (or a single string for convenience). Each string constitutes a single refspec that will be given unmodified to a "git push" command (with a + for force if not already present). The configuration becomes a bit more verbose by using this format, but it's also much more powerful. It makes it possible to express more complex mappings using the mirror bot, which may very well be needed when we start using branches more actively.

When setting "refspecs", we do not accept either "branches" or "tags" in the same configuration.

We still pull everything to the local repository in the bot. The refspec is only applied to the push command. Keeping the local repo the same as the source repo reduced the risk of bot configs interfering with each other through conflicting configurations.

---------
### Progress

### Issue
 * [SKARA-2143](https://bugs.openjdk.org/browse/SKARA-2143): Support selective tags mirroring (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1600/head:pull/1600` \
`$ git checkout pull/1600`

`$ git checkout pull/1600` \
`$ git pull https://git.openjdk.org/skara.git pull/1600/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1600`

`$ git pr show -t 1600`

</details>

<a href="https://git.openjdk.org/skara/pull/1600.diff">https://git.openjdk.org/skara/pull/1600.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1600#issuecomment-1894693601)

### DIFF
--- a/vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
+++ b/vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
@@ -650,6 +650,16 @@ public class GitRepository implements Repository {
     }
 
     @Override
+    public void push(String refspec, URI uri, boolean force) throws IOException {
+        if (force && !refspec.equals("+")) {
+            refspec = "+" + refspec;
+        }
+        try (var p = capture("git", "push", uri.toString(), refspec)) {
+            await(p);
+        }
+    }
+
+    @Override
     public void push(Branch branch, String remote, boolean setUpstream) throws IOException {
         var cmd = new ArrayList<String>();
         cmd.addAll(List.of("git", "push", remote, branch.name()));
--- a/vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
+++ b/vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
@@ -650,10 +650,7 @@ public class GitRepository implements Repository {
     }
 
     @Override
-    public void push(String refspec, URI uri, boolean force) throws IOException {
-        if (force && !refspec.equals("+")) {
-            refspec = "+" + refspec;
-        }
+    public void push(String refspec, URI uri) throws IOException {
         try (var p = capture("git", "push", uri.toString(), refspec)) {
             await(p);
         }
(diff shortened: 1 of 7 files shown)


vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
```diff
@@ -649,6 +649,16 @@ public void push(Tag tag, URI uri, boolean force) throws IOException {
         }
     }
 
+    @Override
+    public void push(String refspec, URI uri, boolean force) throws IOException {
+        if (force && !refspec.equals("+")) {
```
magicus, line 654: Are you sure you did not mean 
```suggestion
        if (force && !refspec.beginsWith("+")) {
```
?

magicus (approved):
Looks good as far as I can tell.

vcs/src/main/java/org/openjdk/skara/vcs/git/GitRepository.java
edvbld: I would have opted to just skip the `boolean force` parameter. If we are already entrusting the user to supply a correct refspec then they will probably know that they should prefix the refspec with a `+` if they want to force push.

erikj79: You are right, I meant beginsWith. But having thought about it, I think Erik has a point. We shouldn't modify the refspec given by the user.

zhaosongzs (approved):
LGTM

erikj79:
Test comment.

magicus (approved):

PR #1610, opened 2024-01-30
2167: Unify search methods in IssueProject

edvbld:
Hi all,

this patch unifies the various ways of searching that exist within `JiraProejct`. I also chose to expose the `search` method in the `IssueProject` since I will make use of it in later patches. This is the first patch of _many_ that will extract the JBS specific parts from `JiraProject` so that `JiraProject` will be about JIRA and JBS specific functionality will be moved elsewhere. This refactoring is needed to allow other parts of the Skara to use JBS functionality with only an `IssueProject` instance (without turning `JiraProject` into effectively `JbsProject`).

Thanks,
Erik

---------
### Progress

### Issue
 * [SKARA-2167](https://bugs.openjdk.org/browse/SKARA-2167): Unify search methods in IssueProject (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1610/head:pull/1610` \
`$ git checkout pull/1610`

`$ git checkout pull/1610` \
`$ git pull https://git.openjdk.org/skara.git pull/1610/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1610`

`$ git pr show -t 1610`

</details>

<a href="https://git.openjdk.org/skara/pull/1610.diff">https://git.openjdk.org/skara/pull/1610.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1610#issuecomment-1916894138)

### DIFF
--- a/issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraProject.java
+++ b/issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraProject.java
@@ -447,27 +447,13 @@ public class JiraProject implements IssueProject {
 
     @Override
     public Optional<IssueTrackerIssue> jepIssue(String jepId) {
-        var issues = request.post("search")
-                .body("jql", "project = " + projectName + " AND \"JEP Number\" ~ \"" + jepId + "\"")
-                .execute();
-        if (issues.get("issues").asArray().size() == 0) {
-            return Optional.empty();
-        } else {
-            var json = issues.get("issues").asArray().get(0);
-            return Optional.of(new JiraIssue(this, generateIssueRequest(json), json));
-        }
+        var issues = search("project = " + projectName + " AND \"JEP Number\" ~ \"" + jepId + "\"");
+        return issues.isEmpty() ? Optional.empty() : Optional.of(issues.get(0));
     }
 
     @Override
     public List<IssueTrackerIssue> issues() {
-        var ret = new ArrayList<IssueTrackerIssue>();
-        var issues = request.post("search")
-                            .body("jql", "project = " + projectName + " AND status in (Open, New)")
-                            .execute();
-        for (var issue : issues.get("issues").asArray()) {
-            ret.add(new JiraIssue(this, generateIssueRequest(issue), issue));
-        }
-        return ret;
+        return search("project = " + projectName + " AND status in (Open, New)");
     }
 
     // Need to sort by updated time in descending order to guarantee that no issues are missed, see SKARA-1962 for details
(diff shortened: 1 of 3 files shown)


issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraProject.java
```diff
@@ -523,10 +506,14 @@ private List<IssueTrackerIssue> queryIssues(String jql) {
 
             if (count < issues.get("total").asInt()) {
                 startAt += issues.get("issues").asArray().size();
```
erikj79, line 508: Aren't `count` and `startAt` basically equivalent at this point? Seems a bit weird to have two values for basically the same thing but calculated differently.

issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraProject.java
```diff
@@ -523,10 +506,14 @@ private List<IssueTrackerIssue> queryIssues(String jql) {
...
-                issues = request.get("search")
-                        .param("jql", jql)
-                        .param("startAt", String.valueOf(startAt))
-                        .execute();
+                req = request.get("search")
+                             .param("jql", jql)
+                             .param("startAt", String.valueOf(startAt));
+                if (limit > 0) {
+                    req = req.param("maxResults", Integer.toString(limit));
```
erikj79, line 513: Is the limit supposed to be the total limit or max size per page (would be great with javadoc specifying this btw)? The `maxResults` parameter is results returned per page. I would think a global limit to be more useful.

edvbld: Yeah, it was meant to be a global limit (we only have one caller currently passing a limit other than -1, and that caller passes 1, so it would have worked, but I agree that the current code is wrong). Will update!

issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraProject.java
edvbld: Yes, I just tried to change the existing code as little as possible to try to keep the patch about refactoring. Since I will be touch this a bit more based on your other comment then I can clean this up as well.

issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraProject.java
```diff
@@ -484,49 +470,65 @@ public List<IssueTrackerIssue> issues(ZonedDateTime updatedAfter) {
...
+                }
             }
 
-            if (count < issues.get("total").asInt()) {
-                startAt += issues.get("issues").asArray().size();
-                issues = request.get("search")
-                        .param("jql", jql)
-                        .param("startAt", String.valueOf(startAt))
-                        .execute();
+            if (count < issues.get("total").asInt() && count < limit) {
```
erikj79, line 522: If `limit == -1` then we will never try to request more pages. Is that intended?

erikj79 (commented):
This looks like it's doing the right thing. It would be good if you could add a manual test that verifies that it at least doesn't blow up, similar to existing manual tests for API calls.

PR #1630, opened 2024-04-09
2226: Make it possible to avoid "forward backports"

edvbld:
Hi all,

please review this patch that makes it possible to configure the PR bot to avoid creating a "forward backport" when integrating a pull request. A "forward backport" gets created when the issue being fixed a PR has a `fixVersion` that is less than the `fixVersion` for the repository (which is specified by `.jcheck/conf`). This patch detects this scenario and instead sets the `fixVersion` for the issue to the repository's `fixVersion` and creates a backport with the `fixVersion` of the original issue. Avoiding "forward backport"s is configurable and will only be enabled for the [jdk](https://github.com/openjdk/jdk) repository (it would be a bug to have this behaviour for update release repositories).

For the patch itself, there are two larger changes that are almost just moving code: the creation of the new class `OverridingJCheckConfigurationParser` and the addition of `createBackport` to `Backports` (and the removal of `JbsBackports`). `OverridingJCheckConfigurationParser` is useful for parsing a `.jcheck/conf` that can be overridden _without_ a local repository (only with a `HostedRepository`). `Backports.createBackport` is almost a clean refactoring of `JbsBackport` - no functionality has been changed.

There is one detail in the patch that might look weird:
```
issue.get().setProperty("fixVersions",
                        JSON.array().add(targetFixVersion.get().raw()));
```
In `IssueNotifier` the code is setting `fixVersions` to a value of a single JSON string. However, this value gets transformed by `JiraProject.encodeProperty`. The problem with not using an array (which the property value should be) is that tests gets confused (since `TestIssueTrackerIssue.setProperty` is *not* transforming its input). I therefore opted to use an array in the production code to make it easier to test.

I also added a bunch of new unit tests as well as slightly updated the integration tests for backports.

Thanks,
Erik

---------
### Progress

### Issue
 * [SKARA-2226](https://bugs.openjdk.org/browse/SKARA-2226): Make it possible to avoid "forward backports" (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1630/head:pull/1630` \
`$ git checkout pull/1630`

`$ git checkout pull/1630` \
`$ git pull https://git.openjdk.org/skara.git pull/1630/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1630`

`$ git pr show -t 1630`

</details>

<a href="https://git.openjdk.org/skara/pull/1630.diff">https://git.openjdk.org/skara/pull/1630.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1630#issuecomment-2046054568)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -654,6 +657,65 @@ public Collection<WorkItem> prRun(ScratchArea scratchArea) {
...
+                                             "', repository fixVersion is '" + targetFixVersion.get().raw() + "'");
+                                    // This scenario would result in a "forward-backport" being created. Avoid it by
+                                    // updating the fixVersion of the original issue and creating a backport with
+                                    // the fixVersion from the original issue.
+                                    issue.get().setProperty("fixVersions",
+                                                            JSON.array().add(targetFixVersion.get().raw()));
+                                    var assignees = issue.get().assignees();
+                                    var backport = assignees.isEmpty() ?
+                                        Backports.createBackport(issue.get(), issueFixVersion.get().raw()) :
+                                        Backports.createBackport(issue.get(), issueFixVersion.get().raw(), assignees.get(0).username());
```
erikj79, line 702: The order of operations here risks leaving a mess if the bot is interrupted at some point. The fixVersion could be updated in the bug, but no backport created and no comment posted. On a rerun nothing would happen. The backport could be created but no comment be posted. On a rerun, no comment would be posted.

Operations like this need to be carefully crafted to guarantee that the end state is eventually reached. The simplest fix would be to create the backport first. Then we would also need to check for a backport with the same fixVersion as the main bug and only create the backport if that doesn't already exist. After that we change the fixVersion of the main bug.

That still won't guarantee that we will post the comment. For a complete solution we will need to post an interim comment with the wording "going to" added before performing any operations. We can look for this comment in addition to the other conditions to detect an incomplete existing attempt. Then when we have performed all operations, we finalize the comment by rewriting it.

We also need tests that verify that the end state is reached when the bot is presented with a PR in any of these incomplete states.

This may seem pedantic, but I'm very sure this will actually happen, more often than one might think. I went through a similar exercise with the /integrate command soon after taking over Skara because it was rather common for the bot to be interrupted mid integration.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -654,6 +657,65 @@ public Collection<WorkItem> prRun(ScratchArea scratchArea) {
...
+
+            // Check if the fixVersion of the issue is older than the fixVersion in the target branch. If so, update the
+            // fixVersion of the issue to fixVersion of the target branch and create a backport with the fixVersion
+            // from the original issue.
+            //
+            // NOTE: The current implementation only works for the issue in the title, *NOT* for all issues solved by the PR
+            if (bot.avoidForwardports()) {
+                var m = ISSUE_ID_PATTERN.matcher(pr.title());
+                Optional<IssueTrackerIssue> issue = m.matches() ? issueTrackerIssue(getMatchGroup(m, "id")) : Optional.empty();
+                if (issue.isPresent()) {
```
erikj79, line 669: Don't we need to check that the main issue is open?

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -654,6 +657,65 @@ public Collection<WorkItem> prRun(ScratchArea scratchArea) {
...
+                                                                            bot.confOverrideRef());
+                        var jcheckConf = parser.parse(pr.targetHeadHash());
+                        var targetFixVersion = jcheckConf.flatMap(c -> c.general().version())
+                                                        .flatMap(v -> JdkVersion.parse(v));
+                        if (targetFixVersion.isPresent()) {
+                            var willCreateBackport = Backports.findIssue(issue.get(), targetFixVersion.get()).isEmpty();
+                            if (willCreateBackport) {
+                                var issueFixVersion = Backports.mainFixVersion(issue.get());
+                                if (issueFixVersion.isPresent() &&
+                                    targetFixVersion.get().compareTo(issueFixVersion.get()) > 0) {
```
erikj79, line 691: This check should be made before `willCreateBackport` as it's much cheaper.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -654,6 +657,65 @@ public Collection<WorkItem> prRun(ScratchArea scratchArea) {
                 }
             }
 
+
+            // Check if the fixVersion of the issue is older than the fixVersion in the target branch. If so, update the
+            // fixVersion of the issue to fixVersion of the target branch and create a backport with the fixVersion
+            // from the original issue.
+            //
+            // NOTE: The current implementation only works for the issue in the title, *NOT* for all issues solved by the PR
+            if (bot.avoidForwardports()) {
```
zhaosongzs, line 666: I am not sure if we should take action with forward port at this early stage. I mean if a user had a typo in the pr title(the issue id is wrong) and this made the pr like a forward port, then Skara bot would set the fixVersion to null for the incorrect issue and create a backport. Even if the user fixes the typo in the pr title, Skara bot will not withdraw the previous action. The user will need to fix it manually.
I think maybe it's better to take action at least when the pr is marked as "ready".

And if you want, I think we can make it work for all the issues associated with this pr.
We can get all the issue ids via `BotUtils.parseAllIssues(prBody)`.


I am also wondering why don't we put the whole logic of `avoidForwardports` in `issueNotifier`, it's where the backports will be created.

erikj79:
Test comment

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
erikj79: You bring up a good point with user mistakes having the potential to create lots of bad backports without any automatic cleanup from the bot. We could potentially handle that by doing it in the `IssueNotifier::onNewIssue` and `::onRemovedIssue`, but it would be brittle. Their current use, which is to add links in the issue, is easily revertible, but backport creation is definitely messier.

The core of the issue is that we are trying to create the backport at PR creation rather than at PR integration. This was a design decision we made a while back, before considering the implementation, as we thought it would be better from a JBS user point of view. The issues would be updated to reflect the correct fixVersion and backports earlier.

@edvbld and I had a discussion and we think that it's probably better to instead do any backport creation at PR integration (which I think means in the `issueNotifier`). However, we need to make it clear to the user that this is going to happen, so the PR bot would still need to maintain comment (likely a section in the merge ready comment) describing exactly which fixVersion would be changed and what backports would be created. I think this description could be useful for all PRs, regardless of using this feature or not, so that users know what to expect with regards to fixVersion and backport creation. To achieve that, we probably need to implement some more common logic between the two bots.

zhaosongzs: Thanks for the details! I was thinking the users might be confused when they found in forward port cases, backports are created at PR creation while they already got used to see backports created at PR integration when they are using /backport command.

edvbld:
@zhaosongzs and @erikj79 thanks for great reviews and discussion! I will split this work into two patches, one for [SKARA-2227 - Describe planned modifications to issues](https://bugs.openjdk.org/browse/SKARA-2227) and one for [SKARA-2226 - Make it possible to avoid "forward backports"](https://bugs.openjdk.org/browse/SKARA-2226).

The implementation of [SKARA-2226](https://bugs.openjdk.org/browse/SKARA-2226) will change to avoid the creation of "forward backports" _once_ a PR has been integrated (and yes, those changes will primarily be made to the `IssueNotifier`). As @erikj79 mentioned the concept of describing planned changes to issues is a general useful concept, so we can apply it to all issues (not only those that happen to have a fixVersiont that differs from the target repository's). That work will be done as part of [SKARA-2227](https://bugs.openjdk.org/browse/SKARA-2227).

I will close this PR since it no longer represents how we want to solve [SKARA-2226](https://bugs.openjdk.org/browse/SKARA-2226) and will open a new one PR with a solution focused on avoiding "forward backports" after a PR has been integrated.

PR #1656, opened 2024-06-04
2278: Issuestitle Check shouldn't treat special character as leading lowercase letter

zhaosongzs:
Currently, in IssuesTitleCheck::hasLeadingLowercaseLetter, the method checks if the leading character is uppercase, if so, it returns false. Otherwise, it will treat the character as lowercase. This is wrong because it treats special characters as lowercase letters. The solution is to make the method to check if the leading character is a lowercase letter.

---------
### Progress

### Issue
 * [SKARA-2278](https://bugs.openjdk.org/browse/SKARA-2278): Issuestitle Check shouldn't treat special character as leading lowercase letter (**Bug** - P4)

### Reviewers

### Contributors
 * Kevin Rushforth `<kcr@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1656/head:pull/1656` \
`$ git checkout pull/1656`

`$ git checkout pull/1656` \
`$ git pull https://git.openjdk.org/skara.git pull/1656/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1656`

`$ git pr show -t 1656`

</details>

<a href="https://git.openjdk.org/skara/pull/1656.diff">https://git.openjdk.org/skara/pull/1656.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1656#issuecomment-2148030237)

### DIFF
--- a/jcheck/src/main/java/org/openjdk/skara/jcheck/IssuesTitleCheck.java
+++ b/jcheck/src/main/java/org/openjdk/skara/jcheck/IssuesTitleCheck.java
@@ -35,7 +35,7 @@ import java.util.regex.Pattern;
 public class IssuesTitleCheck extends CommitCheck {
     private final Logger log = Logger.getLogger("org.openjdk.skara.jcheck.issuesTitle");
     private final static List<String> VALID_WORD_WITH_TRAILING_PERIOD = List.of("et al.", "etc.", "...");
-    private final static Pattern FILE_OR_FUNCTION_PATTERN = Pattern.compile(".*[()/._:].*");
+    private final static Pattern ALL_LOWER_CASE_PATTERN = Pattern.compile("[a-z]+");
 
     @Override
     Iterator<Issue> check(Commit commit, CommitMessage message, JCheckConfiguration conf, Census census) {
@@ -88,11 +88,6 @@ public class IssuesTitleCheck extends CommitCheck {
     }
 
     private boolean hasLeadingLowerCaseLetter(String description) {
-        if (!Character.isLowerCase(description.charAt(0))) {
-            return false;
-        }
-        var firstWord = description.split(" ")[0];
-        // If first word contains special character, it's very likely a reference to file or function, ignore it
-        return !FILE_OR_FUNCTION_PATTERN.matcher(firstWord).matches();
+        return ALL_LOWER_CASE_PATTERN.matcher(description.split(" ")[0]).matches();
     }
 }
--- a/jcheck/src/main/java/org/openjdk/skara/jcheck/IssuesTitleCheck.java
+++ b/jcheck/src/main/java/org/openjdk/skara/jcheck/IssuesTitleCheck.java
@@ -35,7 +35,7 @@ import java.util.regex.Pattern;
 public class IssuesTitleCheck extends CommitCheck {
     private final Logger log = Logger.getLogger("org.openjdk.skara.jcheck.issuesTitle");
     private final static List<String> VALID_WORD_WITH_TRAILING_PERIOD = List.of("et al.", "etc.", "...");
-    private final static Pattern ALL_LOWER_CASE_PATTERN = Pattern.compile("[a-z]+");
+    private final static Pattern FIRST_WORD_ALL_LOWER_CASE_PATTERN = Pattern.compile("[a-z]+(?:\\h.*)?");
 
     @Override
     Iterator<Issue> check(Commit commit, CommitMessage message, JCheckConfiguration conf, Census census) {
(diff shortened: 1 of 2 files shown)


kevinrushforth:
This change will fix this specific problem, but I wonder if it would be better to step back and rethink this check. The main thing this check is trying to catch is the case of an all lower-case initial word, so I would test for _that_ rather than checking that the initial word starts with a lower-case letter and doesn't contain certain other characters.

Something like this:

```
Pattern ALL_LOWER_CASE = Pattern.compile("[a-z]+");
...

    boolean hasLeadingLowerCaseWord(String description) {
        var firstWord = description.split(" ")[0];
        return ALL_LOWER_CASE.matcher(description).matches();
    }
```

Then you can get rid of the `FILE_OR_FUNCTION_PATTERN` regex.

This has the benefit of not warning on any camel case word (e.g., "macOS") and also means you won't have to keep adding special characters to the `FILE_OR_FUNCTION_PATTERN` (which I predict you will otherwise have to do at some point)

zhaosongzs:
> This change will fix this specific problem, but I wonder if it would be better to step back and rethink this check. The main thing this check is trying to catch is the case of an all lower-case initial word, so I would test for _that_ rather than checking that the initial word starts with a lower-case letter and doesn't contain certain other characters.
> 
> Something like this:
> 
> ```
> Pattern ALL_LOWER_CASE = Pattern.compile("[a-z]+");
> ...
> 
>     boolean hasLeadingLowerCaseWord(String description) {
>         var firstWord = description.split(" ")[0];
>         return ALL_LOWER_CASE.matcher(description).matches();
>     }
> ```
> 
> Then you can get rid of the `FILE_OR_FUNCTION_PATTERN` regex.
> 
> This has the benefit of not warning on any camel case word (e.g., "macOS") and also means you won't have to keep adding special characters to the `FILE_OR_FUNCTION_PATTERN` (which I predict you will otherwise have to do at some point)

Seems like it's a good idea. I will think of it. Thank you!

jcheck/src/main/java/org/openjdk/skara/jcheck/IssuesTitleCheck.java
```diff
@@ -88,11 +88,6 @@ private boolean hasTrailingPeriod(String description) {
     }
 
     private boolean hasLeadingLowerCaseLetter(String description) {
-        if (Character.isUpperCase(description.charAt(0))) {
-            return false;
-        }
-        var firstWord = description.split(" ")[0];
-        // If first word contains special character, it's very likely a reference to file or function, ignore it
-        return !FILE_OR_FUNCTION_PATTERN.matcher(firstWord).matches();
+        return ALL_LOWER_CASE_PATTERN.matcher(description.split(" ")[0]).matches();
```
erikj79, line 91: This could all be captured in the regex instead of splitting the string first. Something like `"[a-z]+(:?\\h.*)?"` should do it I think.

kevinrushforth, line 91: I presume that `description` has been `trim`med? If so, then this should be fine without additional checks.

kevinrushforth (commented):
Looks good, but I would wait for Erik. I left one question.

jcheck/src/main/java/org/openjdk/skara/jcheck/IssuesTitleCheck.java
kevinrushforth: Is the `:?` needed? It seems redundant, but I'm not a regex expert.

zhaosongzs: Yes, the description has been trimmed

zhaosongzs: I think Erik has a typo here, should be `"[a-z]+(?:\\h.*)?"`?

kevinrushforth: On closer look, the "`:?`" won't do what we want. It will match one of the patterns we explicitly do not want to match: `"abc: def"`. So I think that should just be: `"[a-z]+(\\h.*)?"`

zhaosongzs: I think Erik was trying to use the non-capturing pattern `?:...`

kevinrushforth: Ah, you may be right (in which case _that_ would be fine).

kevinrushforth (approved):

erikj79 (approved):

jcheck/src/main/java/org/openjdk/skara/jcheck/IssuesTitleCheck.java
erikj79: That was indeed what I was going for.

zhaosongzs:
Thank you both for the review!
/integrate

PR #1660, opened 2024-06-13
2295: Rephrase error messages in PullRequestCheckIssueVisitor

zhaosongzs:
PullRequestCheckIssueVisitor is responsible for generating error messages when jcheck fails. Currently, jcheck could be configured as different severity(error or warning). However, when generating error messages for some jchecks like binary check, the error message would be like "Binary files are not allowed", so if the binary check was configured as warning, this message seems too hard. Therefore, I would like to rephrase some error messages.

Erik suggested that we should generate different messages depending on the severity, but I think maybe it's unnecessary. The user does not need to judge the severity depending on the error message. The user only needs to check which section is the message under(Integration Blocker or Warning).

---------
### Progress

### Issue
 * [SKARA-2295](https://bugs.openjdk.org/browse/SKARA-2295): Rephrase error messages in PullRequestCheckIssueVisitor (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1660/head:pull/1660` \
`$ git checkout pull/1660`

`$ git checkout pull/1660` \
`$ git pull https://git.openjdk.org/skara.git pull/1660/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1660`

`$ git pr show -t 1660`

</details>

<a href="https://git.openjdk.org/skara/pull/1660.diff">https://git.openjdk.org/skara/pull/1660.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1660#issuecomment-2168514343)

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
@@ -149,9 +149,10 @@ class PullRequestCheckIssueVisitor implements IssueVisitor {
     }
 
     @Override
-    public void visit(SelfReviewIssue issue)
-    {
-        addMessage(issue.check(), "Self-reviews are not allowed", issue.severity());
+    public void visit(SelfReviewIssue issue) {
+        var message = issue.severity().equals(Severity.ERROR) ? "Self-reviews are not allowed" :
+                "Self-reviews are not recommended";
+        addMessage(issue.check(), message, issue.severity());
         setNotReadyForReviewOnError(issue.severity());
     }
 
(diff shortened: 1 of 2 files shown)


zhaosongzs:
> I meant that each visit method needs to check the severity and print the appropriate error message for the severity.

Yes, but as I said in the PR body, maybe it's too complicated? Or you think it's necessary...

erikj79:
> Yes, but as I said in the PR body, maybe it's too complicated? Or you think it's necessary...

Sorry, I didn't read that part fully until now. You have a point, but after having thought about this for a bit I still think it would be better to write separate messages. If it's an error it needs to be clear that it's an error.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
```diff
@@ -280,19 +280,19 @@ public void visit(IssuesIssue issue) {
 
     @Override
     public void visit(ExecutableIssue issue) {
-        addMessage(issue.check(), String.format("Executable files are not allowed (file: %s)", issue.path()), issue.severity());
+        addMessage(issue.check(), String.format("Patch contains an executable file (file: %s)", issue.path()), issue.severity());
```
erikj79, line 283: I think for the warning variant, leave out the "file:" part since the sentence is already ending with "file". It looks a bit weird repeating the word.

erikj79 (approved):
Nice, I like these more nuanced messages.

zhaosongzs:
Thanks for the review!
/integrate

PR #1662, opened 2024-06-24
2307: PullRequestCheckIssueVisitor generates only one error message per Issue type

zhaosongzs:
When testing [SKARA-2303](https://bugs.openjdk.org/browse/SKARA-2303), I found that even when there are few binary files in a diff, the pr body only warns about one binary file. Then I found that in PullRequestCheckIssueVisitor, we are using HashMap<Check, String> to store the error messages, so if there are few issues with the same type, the HashMap is only able to save the last error message.

---------
### Progress

### Issue
 * [SKARA-2307](https://bugs.openjdk.org/browse/SKARA-2307): PullRequestCheckIssueVisitor generates only one error message per Issue type (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1662/head:pull/1662` \
`$ git checkout pull/1662`

`$ git checkout pull/1662` \
`$ git pull https://git.openjdk.org/skara.git pull/1662/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1662`

`$ git pr show -t 1662`

</details>

<a href="https://git.openjdk.org/skara/pull/1662.diff">https://git.openjdk.org/skara/pull/1662.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1662#issuecomment-2187681163)

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -77,6 +77,7 @@ class CheckRun {
     private static final Set<String> PRIMARY_TYPES = Set.of("Bug", "New Feature", "Enhancement", "Task", "Sub-task");
     protected static final String CSR_PROCESS_LINK = "https://wiki.openjdk.org/display/csr/Main";
     private static final Path JCHECK_CONF_PATH = Path.of(".jcheck", "conf");
+    private static final int MESSAGE_LIMIT = 50;
     private final Set<String> newLabels;
     private final boolean reviewCleanBackport;
     private final Approval approval;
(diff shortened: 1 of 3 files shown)


erikj79 (commented):
I wonder if this was done on purpose to limit the number of issues listed. 

Is there a limitation on the number of errors shown in the presentation layer? 

Do all checks produce unique messages (like which file has the problem) for each instance? If not, then having multiple instances with the same message isn't going to be very helpful.

zhaosongzs:
> I wonder if this was done on purpose to limit the number of issues listed.

Maybe not intended since this behavior is only in PullRequestCheckIssueVisitor. In JCheckCLIVisitor, it will print all the messages.

> Is there a limitation on the number of errors shown in the presentation layer?

No.

> Do all checks produce unique messages (like which file has the problem) for each instance? If not, then having multiple instances with the same message isn't going to be very helpful.

Yes, all checks will generate unique messages otherwise it will only generate one issue.

erikj79:
>> Is there a limitation on the number of errors shown in the presentation layer?
> No.

The previous implementation did provide an implicit limitation to the number of errors/warnings printed. I think we should add a new explicit limit if we fix this. Something like max 50 or so to avoid flooding the PR body with it.

zhaosongzs:
> > > Is there a limitation on the number of errors shown in the presentation layer?
> 
> > No.
> 
> The previous implementation did provide an implicit limitation to the number of errors/warnings printed. I think we should add a new explicit limit if we fix this. Something like max 50 or so to avoid flooding the PR body with it.

Okay, 50 for all checks or 50 for each check?

erikj79:
I meant 50 total.

erikj79 (approved):

zhaosongzs:
Thanks for the review!
/integrate

PR #1682, opened 2024-07-29
2340: CommitCommentsWorkItem executing in GitLab takes a lot of time

zhaosongzs:
Currently, if a user comments under a commit named "Merge" in GitLab, the CommitCommentsWorkItem will take about 8 hours to complete.
After investigation, I think there is a bug in GitLabRepository#getCommitTitleToCommitsMap

For GitLab repos, we need to build a hash map for mapping commit title to commits. The implementation of the map is Map<String, LinkedHashSet<Hash>>. We use LinkedHashSet here because we want the bot to check latest commits first.

If the map has already been built, then the bot will query new commits in the repo and adds the commits into the map. However, when adding the new commit hash to the map, the bot add it to the end of the LinkedHashSet. So new commits will be checked very late.

For commit titles like "Merge", sometimes, there can be around 50K commits in a repo. If the latest "Merge" commit is added to the tail of the LinkedHashSet, the bot must check all previous commits first, which takes a lot of time.

---------
### Progress

### Issue
 * [SKARA-2340](https://bugs.openjdk.org/browse/SKARA-2340): CommitCommentsWorkItem executing in GitLab takes a lot of time (**Bug** - P2)

### Reviewers
 * [Erik Joelsson](https://openjdk.org/census#erikj) (@erikj79 - **Reviewer**) ⚠️ Review applies to [db3c7597](https://git.openjdk.org/skara/pull/1682/files/db3c75979be82f30363e348674f893d7e41c97ed)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1682/head:pull/1682` \
`$ git checkout pull/1682`

`$ git checkout pull/1682` \
`$ git pull https://git.openjdk.org/skara.git pull/1682/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1682`

`$ git pr show -t 1682`

</details>

<a href="https://git.openjdk.org/skara/pull/1682.diff">https://git.openjdk.org/skara/pull/1682.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1682#issuecomment-2269693969)

### DIFF
--- a/forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabRepository.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabRepository.java
@@ -635,28 +635,18 @@ public class GitLabRepository implements HostedRepository {
                 .param("since", lastCommitTime.format(DateTimeFormatter.ISO_DATE_TIME))
                 .param("all", "true")
                 .execute()
-                .asArray();
-
-        Map<String, Set<Hash>> tempCommitTitleToCommits = new HashMap<>();
+                .asArray()
+                .stream()
+                .toList()
+                .reversed();
 
         for (var commit : commits) {
             var hash = new Hash(commit.get("id").asString());
             var title = commit.get("title").asString();
-            tempCommitTitleToCommits.computeIfAbsent(title, t -> new LinkedHashSet<>()).add(hash);
+            ((LinkedHashSet<Hash>) commitTitleToCommits.computeIfAbsent(title, t -> new LinkedHashSet<>())).addFirst(hash);
             var authored = ZonedDateTime.parse(commit.get("authored_date").asString());
             if (lastCommitTime.isBefore(authored)) {
                 lastCommitTime = authored;
-            }
-        }
-
-        for (var entry : tempCommitTitleToCommits.entrySet()) {
-            if (commitTitleToCommits.containsKey(entry.getKey())) {
-                Set<Hash> temp = new LinkedHashSet<>();
-                temp.addAll(entry.getValue());
-                temp.addAll(commitTitleToCommits.get(entry.getKey()));
-                commitTitleToCommits.put(entry.getKey(), temp);
-            } else {
-                commitTitleToCommits.put(entry.getKey(), entry.getValue());
             }
         }
 
--- a/forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabRepository.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabRepository.java
@@ -529,7 +529,7 @@ public class GitLabRepository implements HostedRepository {
                       .collect(Collectors.toList());
     }
 
-    private Set<Hash> commitsWithTitle(String commitTitle, Map<String, Set<Hash>> commitTitlesToHashes) {
+    private Set<Hash> commitsWithTitle(String commitTitle, Map<String, SequencedSet<Hash>> commitTitlesToHashes) {
         if (commitTitlesToHashes.containsKey(commitTitle)) {
             return commitTitlesToHashes.get(commitTitle);
         }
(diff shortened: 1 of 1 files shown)


forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabRepository.java
```diff
@@ -556,10 +556,15 @@ private Optional<CommitComment> findComment(String commitTitle,
         // exists on that commit before we try to process it. If this fails it's
         // most likely due to inconsistent data from GitLab, which should
         // eventually clear up on subsequent tries.
-        Optional<CommitComment> found = candidates.stream()
-                .flatMap(candidate -> commitComments(candidate).stream())
-                .filter(comment -> comment.id().equals(commentId))
-                .findFirst();
+        Optional<CommitComment> found = Optional.empty();
+        for (var candidate : candidates) {
```
erikj79, line 560: Are you sure this makes a difference? Streams aren't supposed to be eager, so I was expecting the current implementation to do exactly the same thing.

zhaosongzs: I am not sure what do you mean by eager. I think in the original implementation, it will retrieve all the commit comments before filtering, so with so many candidates, it's inefficient.

zhaosongzs: Ah, I did some tests, you are right, need to think more about it

zhaosongzs:
I have updated the PR body and pushed a new commit

forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabRepository.java
```diff
@@ -636,15 +636,30 @@ private Map<String, Set<Hash>> getCommitTitleToCommitsMap(ReadOnlyRepository loc
...
                 .execute()
                 .asArray();
+
+        Map<String, Set<Hash>> tempCommitTitleToCommits = new HashMap<>();
+
         for (var commit : commits) {
             var hash = new Hash(commit.get("id").asString());
             var title = commit.get("title").asString();
-            commitTitleToCommits.computeIfAbsent(title, t -> new LinkedHashSet<>()).add(hash);
+            tempCommitTitleToCommits.computeIfAbsent(title, t -> new LinkedHashSet<>()).add(hash);
```
erikj79, line 645: I read the bug description and I think I understand the problem now. The initial creation of the map iterates commits from newest to oldest, while this loop iterates from oldest to newest.

Instead of creating a temporary map, couldn't we just call `LinkedHashSet::addFirst` to get the new commits added at the front in the correct order?

zhaosongzs: Yes, it's better to use `LinkedHashSet::addFirst`

forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabRepository.java
```diff
@@ -635,16 +635,21 @@ private Map<String, Set<Hash>> getCommitTitleToCommitsMap(ReadOnlyRepository loc
...
+                .asArray()
+                .stream()
+                .toList()
+                .reversed();
+
         for (var commit : commits) {
             var hash = new Hash(commit.get("id").asString());
             var title = commit.get("title").asString();
-            commitTitleToCommits.computeIfAbsent(title, t -> new LinkedHashSet<>()).add(hash);
+            ((LinkedHashSet<Hash>) commitTitleToCommits.computeIfAbsent(title, t -> new LinkedHashSet<>())).addFirst(hash);
```
erikj79, line 646: Instead of casting to LinkedHashSet, we can use the type directly in the type declaration for `commitTitleToCommits`. We can even use the new shiny `SequencedSet` interface to avoid specifying the concrete class.

zhaosongzs: Sure

erikj79 (approved):

zhaosongzs:
Thanks for the suggestions and review!
/integrate

PR #1696, opened 2024-11-18
2400: Avoid overwriting mbox archive

zhaosongzs:
Sometimes, the ArchiveWorkItem fails to retrieve the current contents of an mbox file while processing a PR. When this happens, it will resend all emails. Erik pointed out that in GitLab, creating a new file uses a POST request, while updating an existing file uses PUT. Similarly, GitHub requires the current "SHA" to update a file.

The ArchiveWorkItem can determine whether it found existing content in the mbox archive and choose the right method accordingly. If it incorrectly fails to retrieve the existing content from GitLab or GitHub, it will attempt to create a new file. This attempt should fail, triggering a retry of the WorkItem, which would ideally succeed in retrieving the existing content.

---------
### Progress

### Issue
 * [SKARA-2400](https://bugs.openjdk.org/browse/SKARA-2400): Avoid overwriting mbox archive (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1696/head:pull/1696` \
`$ git checkout pull/1696`

`$ git checkout pull/1696` \
`$ git pull https://git.openjdk.org/skara.git pull/1696/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1696`

`$ git pr show -t 1696`

</details>

<a href="https://git.openjdk.org/skara/pull/1696.diff">https://git.openjdk.org/skara/pull/1696.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1696#issuecomment-2484225353)
</details>

### DIFF
--- a/forge/src/main/java/org/openjdk/skara/forge/HostedRepository.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/HostedRepository.java
@@ -108,7 +108,7 @@ public interface HostedRepository {
      * @param authorName  Name of author and committer for commit
      * @param authorEmail Email of author and committer for commit
      */
-    void writeFileContents(String filename, String content, Branch branch, String message, String authorName, String authorEmail);
+    void writeFileContents(String filename, String content, Branch branch, String message, String authorName, String authorEmail, boolean createNewFile);
     String namespace();
     Optional<WebHook> parseWebHook(JSONValue body);
     HostedRepository fork();
--- a/forge/src/main/java/org/openjdk/skara/forge/HostedRepository.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/HostedRepository.java
@@ -107,6 +107,11 @@ public interface HostedRepository {
      * @param message     Commit message
      * @param authorName  Name of author and committer for commit
      * @param authorEmail Email of author and committer for commit
+     * @param createNewFile Determines the file operation mode
+     *                      If set to `true`, the operation attempts to create a new file and write contents to it.
+     *                      The operation will fail if the file already exists.
+     *                      If set to `false`, the operation attempts to update an existing file.
+     *                      The operation will fail if the file does not exist.
      */
     void writeFileContents(String filename, String content, Branch branch, String message, String authorName, String authorEmail, boolean createNewFile);
     String namespace();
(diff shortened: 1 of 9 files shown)


forge/src/main/java/org/openjdk/skara/forge/HostedRepository.java
```diff
@@ -108,7 +108,7 @@ PullRequest createPullRequest(HostedRepository target,
      * @param authorName  Name of author and committer for commit
      * @param authorEmail Email of author and committer for commit
      */
-    void writeFileContents(String filename, String content, Branch branch, String message, String authorName, String authorEmail);
+    void writeFileContents(String filename, String content, Branch branch, String message, String authorName, String authorEmail, boolean createNewFile);
```
erikj79, line 111: Please update the javadoc with the new parameter and make it clear what should happen when set to true or false. If set to true, the call will fail if the file exists. If set to false, the call will fail if the file does not exist.

forge/src/main/java/org/openjdk/skara/forge/github/GitHubRepository.java
```diff
@@ -291,7 +291,7 @@ public Optional<String> fileContents(String filename, String ref) {
     }
 
     @Override
-    public void writeFileContents(String filename, String content, Branch branch, String message, String authorName, String authorEmail) {
+    public void writeFileContents(String filename, String content, Branch branch, String message, String authorName, String authorEmail, boolean createNewEmail) {
```
erikj79, line 294: This variable name doesn't make sense here.

test/src/main/java/org/openjdk/skara/test/TestHostedRepository.java
```diff
@@ -229,7 +229,7 @@ public Optional<String> fileContents(String filename, String ref) {
     }
 
     @Override
-    public void writeFileContents(String filename, String content, Branch branch, String message, String authorName, String authorEmail) {
+    public void writeFileContents(String filename, String content, Branch branch, String message, String authorName, String authorEmail, boolean createNewFile) {
```
erikj79, line 232: This implementation should also enforce the new behavior so we catch any mistakes in testing.

forge/src/main/java/org/openjdk/skara/forge/HostedRepository.java
zhaosongzs: Will do

forge/src/main/java/org/openjdk/skara/forge/github/GitHubRepository.java
zhaosongzs: Thanks for catching it! I was in a hurry when working on it

test/src/main/java/org/openjdk/skara/test/TestHostedRepository.java
zhaosongzs: Will fix it

erikj79 (approved):

zhaosongzs:
Thank you for the review!
/integrate

PR #1698, opened 2024-11-20
2411: Host::user should throw when failing

erikj79:
The implementations of Host::user are all currently returning Optional::empty on any error return code. This means a that a bot checking for the existence of a user encounters an error with a host, e.g. JBS, it will think the user doesn't exist and take actions based on that assumption. This would be correct if the server response was 404, but in all other cases, it would be better if the WorkItem failed to be automatically retried later.

This patch changes the onError lambda for these methods so that it only handles the 404 case, and in all other cases lets the RestException be thrown. This matches the behavior of most other methods in these classes.

---------
### Progress

### Issue
 * [SKARA-2411](https://bugs.openjdk.org/browse/SKARA-2411): Host::user should throw when failing (**Bug** - P2)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1698/head:pull/1698` \
`$ git checkout pull/1698`

`$ git checkout pull/1698` \
`$ git pull https://git.openjdk.org/skara.git pull/1698/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1698`

`$ git pr show -t 1698`

</details>

<a href="https://git.openjdk.org/skara/pull/1698.diff">https://git.openjdk.org/skara/pull/1698.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1698#issuecomment-2488858043)
</details>

### DIFF
--- a/forge/src/main/java/org/openjdk/skara/forge/github/GitHubHost.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/github/GitHubHost.java
@@ -323,7 +323,7 @@ public class GitHubHost implements Forge {
     @Override
     public Optional<HostUser> user(String username) {
         var details = request.get("users/" + URLEncoder.encode(username, StandardCharsets.UTF_8))
-                             .onError(r -> Optional.of(JSON.of()))
+                             .onError(r -> r.statusCode() == 404 ? Optional.of(JSON.of()) : Optional.empty())
                              .execute();
         if (details.isNull()) {
             return Optional.empty();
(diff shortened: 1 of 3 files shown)


kevinrushforth (approved):
This looks good. to me.

One question about the Description:

> The implementations of Host::user are all currently returning Optional::empty on any error return code

I think you mean they are all currently returning `Optional.of(JSON.of())` right? And that after this change you will continue to return that for 404, and return `Optional.empty()` (which will throw) for all other cases.

erikj79:
> One question about the Description:
> 
> > The implementations of Host::user are all currently returning Optional::empty on any error return code
> 
> I think you mean they are all currently returning `Optional.of(JSON.of())` right? And that after this change you will continue to return that for 404, and return `Optional.empty()` (which will throw) for all other cases.

No, the description is about the `Host::user` method, which in turn uses `RestRequest`, providing a lambda for `onError`. What you are describing is what that lambda does, which the Host::user methods will translate to `Optional<HostUser>` that either contains an instance of HostUser, or is empty.

kevinrushforth:
Thanks for the clarification.

zhaosongzs (approved):

erikj79:
Thanks!

/integrate

PR #1699, opened 2024-11-21
SKARA-2415

zhaosongzs:
As Pavel pointed out in the pull request of [SKARA-2410](https://bugs.openjdk.org/browse/SKARA-2410)(https://github.com/openjdk/skara/pull/1697), with current implementation , GitCommits::close may have created multiple processes and only destroy one of them.

As Erik and Pavel suggested, we should make sure close() is called on each process.

### DIFF
--- a/vcs/src/main/java/org/openjdk/skara/vcs/git/GitCommits.java
+++ b/vcs/src/main/java/org/openjdk/skara/vcs/git/GitCommits.java
@@ -135,10 +135,24 @@ class GitCommits implements Commits, AutoCloseable {
             }
         }
 
+        RuntimeException runtimeException = null;
+
         for (var i = 0; i < processes.size(); i++) {
             var p = processes.get(i);
             var command = commands.get(i);
-            close(p, command);
+            try {
+                close(p, command);
+            } catch (Exception e) {
+                if (runtimeException == null) {
+                    runtimeException = new RuntimeException(e);
+                } else {
+                    runtimeException.addSuppressed(e);
+                }
+            }
+        }
+
+        if (runtimeException != null) {
+            throw runtimeException;
         }
     }
 


vcs/src/main/java/org/openjdk/skara/vcs/git/GitCommits.java
```diff
@@ -135,10 +135,24 @@ public void close() throws IOException {
...
+
         for (var i = 0; i < processes.size(); i++) {
             var p = processes.get(i);
             var command = commands.get(i);
-            close(p, command);
+            try {
+                close(p, command);
+            } catch (Exception e) {
+                if (runtimeException == null) {
+                    runtimeException = new RuntimeException(e);
```
erikj79, line 147: This converts an IOException to a RuntimeException, which I don't think is ideal. I would suggest just saving the first exception, which should be either a RuntimeException or IOException, and then add any additional exceptions to the suppressed set on the saved exception.

erikj79 (approved):

vcs/src/main/java/org/openjdk/skara/vcs/git/GitCommits.java
```diff
@@ -135,10 +135,28 @@ public void close() throws IOException {
...
 
+        Exception exception = null;
+
         for (var i = 0; i < processes.size(); i++) {
             var p = processes.get(i);
             var command = commands.get(i);
-            close(p, command);
+            try {
+                close(p, command);
+            } catch (IOException | RuntimeException e) {
```
pavelrappo, line 145: What kind of runtime exception are you expecting here?

erikj79: I suggested catching it mostly to be defensive. The Skara code base uses RuntimeExceptions quite a lot. In this case, there is no obvious place where one would be thrown though

pavelrappo: > I suggested catching it mostly to be defensive. The Skara code base uses RuntimeExceptions quite a lot. In this case, there is no obvious place where one would be thrown though

I've had a look at those; I see what you mean.

A separate nit for later is calling those exceptions _errors_; this is not conventional in Java:

    class GitHubApplicationError extends RuntimeException

"Error" strongly suggests a subclass of `java.lang.Error`. If you need some other word, but "Exception" is undesirable, you may want to call it "Failure".

zhaosongzs: I think we can improve it later, but not in this pr.

zhaosongzs:
Thank you for the review!
/integrate

vcs/src/main/java/org/openjdk/skara/vcs/git/GitCommits.java
pavelrappo: I see you have already integrated the PR; okay.

Erik, I couldn't see where those Skara-specific runtime exceptions could be thrown from in this particular case, but maybe you see it as some sort of future proofing?

erikj79: > Erik, I couldn't see where those Skara-specific runtime exceptions could be thrown from in this particular case, but maybe you see it as some sort of future proofing?

> In this case, there is no obvious place where one would be thrown though

I meant that there is no explicit throwing of RuntimeException in this code, but if we would ever get an NPE or similar, I think we should still make sure all processes are killed.

PR #1702, opened 2024-12-05
SKARA-1357

zhaosongzs:
This patch introduces a copyright jcheck.
For pull requests, the check checks the copyright header in all modified files(if the file name matches the regex in .jcheck/conf)

### DIFF
--- a/cli/src/main/java/org/openjdk/skara/cli/JCheckCLIVisitor.java
+++ b/cli/src/main/java/org/openjdk/skara/cli/JCheckCLIVisitor.java
@@ -335,6 +335,9 @@ class JCheckCLIVisitor implements IssueVisitor {
             if (!i.filesWithCopyrightYearIssue().isEmpty()) {
                 println(i, "Found outdated copyright year in [" + String.join(", ", i.filesWithCopyrightYearIssue()) + "]");
             }
+            if (!i.filesWithCopyrightMissingIssue().isEmpty()) {
+                println(i, "Can't find copyright header in [" + String.join(", ", i.filesWithCopyrightMissingIssue()) + "]");
+            }
             for (var line : i.commit().message()) {
                 System.out.println("> " + line);
             }
--- a/cli/src/main/java/org/openjdk/skara/cli/JCheckCLIVisitor.java
+++ b/cli/src/main/java/org/openjdk/skara/cli/JCheckCLIVisitor.java
@@ -326,6 +326,22 @@ class JCheckCLIVisitor implements IssueVisitor {
         }
     }
 
+    @Override
+    public void visit(CopyrightIssue i) {
+        if (!ignore.contains(i.check().name()) && !isLax) {
+            if (!i.filesWithCopyrightFormatIssue().isEmpty()) {
+                println(i, "Found copyright format issue in [" + String.join(", ", i.filesWithCopyrightFormatIssue()) + "]");
+            }
+            if (!i.filesWithCopyrightYearIssue().isEmpty()) {
+                println(i, "Found outdated copyright year in [" + String.join(", ", i.filesWithCopyrightYearIssue()) + "]");
+            }
+            for (var line : i.commit().message()) {
+                System.out.println("> " + line);
+            }
+            hasDisplayedErrors = i.severity() == Severity.ERROR;
+        }
+    }
+
     public boolean hasDisplayedErrors() {
         return hasDisplayedErrors;
     }
--- a/jcheck/src/main/java/org/openjdk/skara/jcheck/CopyrightCheck.java
+++ b/jcheck/src/main/java/org/openjdk/skara/jcheck/CopyrightCheck.java
@@ -51,6 +51,7 @@ public class CopyrightCheck extends CommitCheck {
 
         var filesWithCopyrightFormatIssue = new ArrayList<String>();
         var filesWithCopyrightYearIssue = new ArrayList<String>();
+        var filesWithCopyrightMissingIssue = new ArrayList<String>();
 
         for (var diff : commit.parentDiffs()) {
             for (var patch : diff.patches()) {
@@ -61,8 +62,10 @@ public class CopyrightCheck extends CommitCheck {
                 if (pattern.matcher(path.toString()).matches()) {
                     try {
                         var lines = repo.lines(path, commit.hash()).orElse(List.of());
+                        var copyrightFound = false;
                         for (String line : lines) {
                             if (line.contains("Copyright (c)") && line.contains("Oracle")) {
+                                copyrightFound = true;
                                 var matcher = COPYRIGHT_PATTERN.matcher(line);
                                 if (matcher.matches()) {
                                     int minYear = Integer.parseInt(matcher.group(1));
@@ -76,6 +79,9 @@ public class CopyrightCheck extends CommitCheck {
                                 }
                             }
                         }
+                        if (!copyrightFound) {
+                            filesWithCopyrightMissingIssue.add(path.toString());
+                        }
                     } catch (IOException e) {
                         throw new RuntimeException(e);
                     }
@@ -83,8 +89,8 @@ public class CopyrightCheck extends CommitCheck {
             }
         }
 
-        if (!filesWithCopyrightFormatIssue.isEmpty() || !filesWithCopyrightYearIssue.isEmpty()) {
-            return iterator(new CopyrightIssue(metadata, filesWithCopyrightFormatIssue, filesWithCopyrightYearIssue));
+        if (!filesWithCopyrightFormatIssue.isEmpty() || !filesWithCopyrightYearIssue.isEmpty() || !filesWithCopyrightMissingIssue.isEmpty()) {
+            return iterator(new CopyrightIssue(metadata, filesWithCopyrightFormatIssue, filesWithCopyrightYearIssue, filesWithCopyrightMissingIssue));
         }
 
         return iterator();
--- a/jcheck/src/main/java/org/openjdk/skara/jcheck/CopyrightCheck.java
+++ b/jcheck/src/main/java/org/openjdk/skara/jcheck/CopyrightCheck.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright (c) 2024, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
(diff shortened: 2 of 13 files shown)


jddarcy:
From looking over the logic, I think some additional accommodations will be needed, at least for update trains. For example, if a patch from a feature release authored in 2022 is backported unchanged to an update release, the affected files should have a copyright year end date of 2022 and *not* 2024.

zhaosongzs:
> From looking over the logic, I think some additional accommodations will be needed, at least for update trains. For example, if a patch from a feature release authored in 2022 is backported unchanged to an update release, the affected files should have a copyright year end date of 2022 and _not_ 2024.

You're right, maybe this is why Erik said "not make any attempt at verifying that files have the correct header, or that the copyright years are correct" in the issue description

erikj79:
From the bug description:

> Note that this feature is only about verifying the format of Oracle copyright headers. It will, for example, not make any attempt at verifying that files have the correct header, or that the copyright years are correct. These are things that need human inspection.

We have a pretty well established agreement from discussing this feature over the years that we should not try to automate copyright year values. The copyright year should only be updated if a "significant" change is made to the copyrighted material. Because of this it's a judgement call each time a file is edited. The issue Joe brings up also makes it hard to automate. We simply don't have the information about when the copyrighted material was created.

jcheck/src/main/java/org/openjdk/skara/jcheck/CopyrightCheck.java
```diff
@@ -0,0 +1,108 @@
...
+                if (patch.target().path().isEmpty() || patch.isBinary()) {
+                    continue;
+                }
+                var path = patch.target().path().get();
+                if (pattern.matcher(path.toString()).matches()) {
+                    try {
+                        var lines = repo.lines(path, commit.hash()).orElse(List.of());
+                        var copyrightFound = false;
+                        for (String line : lines) {
+                            if (line.contains("Copyright (c)") && line.contains("Oracle")) {
```
erikj79: I think it would be nice if we could work with pairs of regular expressions. One for identifying a line that should match and one that verifies the format. This way we could make it possible to add more variants through configuration.

We could also consider not having any default and put the Oracle config in .jcheck/conf.

jcheck/src/main/java/org/openjdk/skara/jcheck/CopyrightConfiguration.java
```diff
@@ -0,0 +1,53 @@
...
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package org.openjdk.skara.jcheck;
+
+import org.openjdk.skara.ini.Section;
+
+public class CopyrightConfiguration {
+    static final CopyrightConfiguration DEFAULT =
+            new CopyrightConfiguration(".*\\.cpp|.*\\.hpp|.*\\.c|.*\\.h|.*\\.java");
```
erikj79: I would like to see more filetypes included by default. At least `Makefile`, `*.gmk`, `configure`, `*.m4`, `*.md` and `*.properties`. Probably also `*.cc`, `*.hh` and `*.m`.

We could also consider not defining defaults here and have it all in .jcheck/conf for better visibility.

zhaosongzs: Ah, I didn't think about the default filetypes, just copied from the default of whitespaceConfiguration. I was thinking we can configure it in .jcheck/conf later like the configuration for  whitespace check https://github.com/openjdk/jdk/blob/41c8971cbcc1d281ffae907c952637ebff945ebb/.jcheck/conf#L19

dholmes-ora:
> The copyright year should only be updated if a "significant" change is made to the copyrighted material.

The copyright year is supposed to be always updated and we are not supposed to try to assess significance.

jcheck/src/main/java/org/openjdk/skara/jcheck/CopyrightConfiguration.java
magicus: I definitely agree that file patterns that should be checked should be defined in .jcheck/conf.

jcheck/src/main/java/org/openjdk/skara/jcheck/CopyrightCheck.java
magicus: I'm not sure its better to have Oracle in the configuration file. It kind of just highlights it as a special case. If/when some other company requests special copyright checks, then we can generalize and move it to the configuration file. Also, it's never the case that this is just a per-repo setting: for Oracle-copyrighted code, this rule always applies.

magicus:
> > The copyright year should only be updated if a "significant" change is made to the copyrighted material.
> 
> The copyright year is supposed to be always updated and we are not supposed to try to assess significance.

I've heard both statements before, but I think what David says is the latest bid. But let's not continue this discussion here; that's really an Oracle-internal discussion. 

I suggest that for this patch, we do not look at what year is used, only that it conforms to the special syntax required. We can come back and update the functionality later on if we agree that the rules permit an automatic updating.

kevinrushforth:
> I suggest that for this patch, we do not look at what year is used, only that it conforms to the special syntax required. We can come back and update the functionality later on if we agree that the rules permit an automatic updating.

I strongly agree with this. The reason for this PR and bug fix is to prevent the all too common error in formatting (usually a missing comma). Please let's not expand the scope to be more than that.

jcheck/src/main/java/org/openjdk/skara/jcheck/CopyrightConfiguration.java
kevinrushforth: I also think it would be better to not have any defaults and just define the patterns in `.jcheck/conf`

zhaosongzs: Yes, I discussed with Erik offline yesterday, and we agreed on removing the default setting for this check.

zhaosongzs:
It's my PR for testing
https://github.com/openjdk/playground/pull/224
It's the .jcheck/conf for testing
https://github.com/openjdk/playground/blob/testSKARA/.jcheck/conf

jddarcy:
> From looking over the logic, I think some additional accommodations will be needed, at least for update trains. For example, if a patch from a feature release authored in 2022 is backported unchanged to an update release, the affected files should have a copyright year end date of 2022 and _not_ 2024.

PS FWIW, some related thoughts on automating checks of copyrights and licenses:

https://mail.openjdk.org/pipermail/core-libs-dev/2024-December/137090.html

jcheck/src/main/java/org/openjdk/skara/jcheck/CopyrightCheck.java
erikj79: I don't want defaults because once we set them, we have to live with them. Changing the default configuration is risky as it applies globally to all commits that have already been checked as well. I think it's generally a better design having the configuration in the affected repo for increased transparency on what is actually being checked.

cli/src/main/java/org/openjdk/skara/cli/JCheckCLIVisitor.java
```diff
@@ -326,6 +326,26 @@ public void visit(IssuesTitleIssue i) {
         }
     }
 
+    @Override
+    public void visit(CopyrightFormatIssue i) {
+        if (!ignore.contains(i.check().name()) && !isLax) {
+            if (!i.filesWithCopyrightFormatIssue().isEmpty()) {
```
erikj79, line 332: No need to check empty since we are iterating over it anyway. It doesn't hurt to do it, but I think it's more readable without the extra if statement.

jcheck/src/main/java/org/openjdk/skara/jcheck/CopyrightFormatCheck.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @Override
+    Iterator<Issue> check(Commit commit, CommitMessage message, JCheckConfiguration conf, Census census) {
+        var metadata = CommitIssue.metadata(commit, message, conf, this);
+        var copyrightConf = conf.checks().copyright();
+        if (copyrightConf == null) {
+            return iterator();
+        }
+        var pattern = Pattern.compile(copyrightConf.files());
```
erikj79, line 52: Could use a more descriptive name.
```suggestion
        var filesPattern = Pattern.compile(copyrightConf.files());
```

jcheck/src/main/java/org/openjdk/skara/jcheck/CopyrightFormatCheck.java
```diff
@@ -0,0 +1,110 @@
...
+        var copyrightSingleCheckConfigurations = copyrightConf.singleCheckConfigurations();
```
erikj79, line 53: This variable name is very long. How about just `copyrightConfigs`?

jcheck/src/main/java/org/openjdk/skara/jcheck/CopyrightFormatCheck.java
```diff
@@ -0,0 +1,110 @@
...
+                                }
+                            }
+                            if (singleConf.required && !copyrightFound) {
+                                filesWithCopyrightMissingIssue
+                                        .computeIfAbsent(singleConf.name, k -> new ArrayList<>())
+                                        .add(path.toString());
+                            }
+                        }
+                    } catch (IOException e) {
+                        throw new RuntimeException(e);
```
erikj79, line 88: I would suggest throwing `UncheckedIOException` instead.

jcheck/src/main/java/org/openjdk/skara/jcheck/CopyrightSingleCheckConfiguration.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package org.openjdk.skara.jcheck;
+
+import java.util.regex.Pattern;
+
+public class CopyrightSingleCheckConfiguration {
```
erikj79, line 27: Maybe move this to an inner class of `CopyrightFormatConfiguration` and shorten the name to something more manageable?

cli/src/main/java/org/openjdk/skara/cli/JCheckCLIVisitor.java
zhaosongzs: Yes, fixed it

jcheck/src/main/java/org/openjdk/skara/jcheck/CopyrightFormatConfiguration.java
```diff
@@ -29,20 +29,54 @@
 import java.util.regex.Pattern;
 
 public class CopyrightFormatConfiguration {
+
+    public static class CopyrightConfiguration {
```
erikj79, line 33: Did you consider making this a record? Looks like a good fit.

zhaosongzs: Yes, it makes sense. I have made it a record

erikj79 (approved):

zhaosongzs:
Thank you all for comments and reviews!
/integrate

AlanBateman:
Is there follow-up work to enable this in the "jdk" repo?

erikj79:
> Is there follow-up work to enable this in the "jdk" repo?

There was: https://github.com/openjdk/jdk/pull/22721

AlanBateman:
> > Is there follow-up work to enable this in the "jdk" repo?
> 
> There was: [openjdk/jdk#22721](https://github.com/openjdk/jdk/pull/22721)

I missed that, is it effective?

kevinrushforth:
It doesn't seem like it is working as expected.

zhaosongzs:
> It doesn't seem like it is working as expected.

Did you find any bug?

zhaosongzs:
> [PR23577](https://github.com/openjdk/jdk/pull/23577) is one from yesterday.

Thank you, I will investigate what's the issue.

kevinrushforth:
> > [PR23577](https://github.com/openjdk/jdk/pull/23577) is one from yesterday.
> 
> Thank you, I will investigate what's the issue.

To clarify, the above PR is the one that fixed the mistakenly introduced (missing comma) copyright line. The PR that introduce it was PR openjdk/jdk#23550

zhaosongzs:
Opened https://github.com/openjdk/jdk/pull/23598 to fix the issue

PR #1706, opened 2025-03-03
2451: Skara bot fails to transition bug to Resolved due to empty response from api

zhaosongzs:
A user reported that one backport issue was not automatically resolved by the Skara bot. From the logs, Erik found that the bot didn't actually send the request to transition the bug to resolved and he noticed that around that time, the bot was renewing the access token for jbs.

When updating the state of an issue, the Skara bot first attempts to fetch the available transitions for that issue by calling this API:
https://bugs.openjdk.org/rest/api/2/issue/issue_id/transitions.

I called the API without authentication, I was expecting a 401 Unauthorized response. However, I received an empty response:
`{"expand":"transitions","transitions":[]}.`

Due to this empty response, the Skara bot did not attempt to transition the bug to any state.

In my opinion, availableTransitions would be empty only when requesting the API without authentication,  so skara bot would throw an exception in this case.

---------
### Progress

### Issue
 * [SKARA-2451](https://bugs.openjdk.org/browse/SKARA-2451): Skara bot fails to transition bug to Resolved due to empty response from api (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1706/head:pull/1706` \
`$ git checkout pull/1706`

`$ git checkout pull/1706` \
`$ git pull https://git.openjdk.org/skara.git pull/1706/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1706`

`$ git pr show -t 1706`

</details>

<a href="https://git.openjdk.org/skara/pull/1706.diff">https://git.openjdk.org/skara/pull/1706.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1706#issuecomment-2695148403)
</details>

### DIFF
--- a/issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraIssue.java
+++ b/issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraIssue.java
@@ -250,6 +250,10 @@ public class JiraIssue implements IssueTrackerIssue {
     public void setState(State state) {
         var availableTransitions = availableTransitions();
 
+        if (availableTransitions.isEmpty()) {
+            throw new RuntimeException("Available transition states are empty");
+        }
+
         // Handle special cases
         if (state == State.RESOLVED) {
             if (!availableTransitions.containsKey("Resolved")) {
@@ -261,6 +265,7 @@ public class JiraIssue implements IssueTrackerIssue {
                     }
                 } else {
                     // The issue is most likely closed - skip transitioning
+                    log.warning("Can't transition the issue to Resolved or Open");
                     return;
                 }
             }
--- a/issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraIssue.java
+++ b/issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraIssue.java
@@ -251,7 +251,7 @@ public class JiraIssue implements IssueTrackerIssue {
         var availableTransitions = availableTransitions();
 
         if (availableTransitions.isEmpty()) {
-            throw new RuntimeException("Available transition states are empty");
+            throw new RuntimeException("Available transition states is empty");
         }
 
         // Handle special cases


issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraIssue.java
```diff
@@ -261,6 +265,7 @@ public void setState(State state) {
                     }
                 } else {
                     // The issue is most likely closed - skip transitioning
+                    log.warning("Can't transition the issue to Resolved or Open");
```
zhaosongzs, line 268: @erikj79 You suggested that 

> The else clause should at the very least log something so we know that is what happened. we should also verify that the state is either closed or resolved and throw RuntimeException if not. 

I thought about it, then I realized that the Skara bot should have already verified the state before calling Issue.setState(Issue.State.RESOLVED) like [this](https://github.com/openjdk/skara/blob/b8a99e211aae1cec3ae5b6477e9c0c91c0761284/bots/notify/src/main/java/org/openjdk/skara/bots/notify/issue/IssueNotifier.java#L211).

Also, it's will be a little hard to verify the Issue state since we need to change the interface of Issue.setState.

issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraIssue.java
```diff
@@ -250,6 +250,10 @@ private void performTransition(String id) {
     public void setState(State state) {
         var availableTransitions = availableTransitions();
 
+        if (availableTransitions.isEmpty()) {
+            throw new RuntimeException("Available transition states are empty");
```
erikj79, line 254:
```suggestion
            throw new RuntimeException("Available transition states is empty");
```

issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraIssue.java
erikj79: I think logging is good enough for now.

erikj79 (approved):

zhaosongzs:
Thank you for the review!
/integrate

PR #1714, opened 2025-04-17
2479: PR marked as ready with jcheck error

zhaosongzs:
Currently, when determining if a PR is ready, if the PR doesn't require a review, it also ignores all jcheck failures. Instead, only reviewers check should be ignored in this case.

---------
### Progress

### Issue
 * [SKARA-2479](https://bugs.openjdk.org/browse/SKARA-2479): PR marked as ready with jcheck error (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1714/head:pull/1714` \
`$ git checkout pull/1714`

`$ git checkout pull/1714` \
`$ git pull https://git.openjdk.org/skara.git pull/1714/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1714`

`$ git pr show -t 1714`

</details>

<a href="https://git.openjdk.org/skara/pull/1714.diff">https://git.openjdk.org/skara/pull/1714.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1714#issuecomment-2813639949)
</details>

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -1470,7 +1470,7 @@ class CheckRun {
             var commitMessage = String.join("\n", commit.message());
 
             var readyToPostApprovalNeededComment = readyForReview &&
-                    visitor.hasErrors(reviewNeeded) &&
+                    !visitor.hasErrors(reviewNeeded) &&
                     integrationBlockers.isEmpty() &&
                     !statusMessage.contains(TEMPORARY_ISSUE_FAILURE_MARKER);
 
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -1469,11 +1469,9 @@ class CheckRun {
             var commit = localRepo.lookup(amendedHash).orElseThrow();
             var commitMessage = String.join("\n", commit.message());
 
-            var onlyReviewersCheckFailed = visitor.errorFailedChecksMessages().stream()
-                    .allMatch(message -> message.contains("Too few reviewers"));
-
             var readyToPostApprovalNeededComment = readyForReview &&
-                    ((!reviewNeeded && onlyReviewersCheckFailed) || visitor.errorFailedChecksMessages().isEmpty()) &&
+                    ((!reviewNeeded && visitor.errorFailedCheckMessagesWithoutReviewersCheck().isEmpty()) ||
+                            visitor.errorFailedChecksMessages().isEmpty()) &&
                     integrationBlockers.isEmpty() &&
                     !statusMessage.contains(TEMPORARY_ISSUE_FAILURE_MARKER);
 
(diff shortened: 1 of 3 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -1469,8 +1469,11 @@ private void checkStatus() {
             var commit = localRepo.lookup(amendedHash).orElseThrow();
             var commitMessage = String.join("\n", commit.message());
 
+            var onlyReviewersCheckFailed = visitor.errorFailedChecksMessages().stream()
+                    .allMatch(message -> message.contains("Too few reviewers"));
```
erikj79, line 1473: This looks awkward. Can we extend the Visitor with an explicit method for this instead? The role of the Visitor implementation is to gather the jcheck errors and present them in a way that makes sense to the caller, in this case the pr bot classes.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
```diff
@@ -72,6 +72,14 @@ List<String> errorFailedChecksMessages() {
         return errorFailedChecks.values().stream().flatMap(List::stream).toList();
     }
 
+    List<String> errorFailedCheckMessagesWithoutReviewersCheck() {
```
erikj79, line 75: This is a mouthful. I was thinking something like `boolean tooFewReviewers()`.

zhaosongzs: Do you mean that `tooFewReviewers()` should return true when there is only ReviewersCheck Failure?

zhaosongzs: Or does `tooFewReviewers()` return true when there is a ReviewersCheck Failure in the list?
If so, the logic should be  like this
`            var readyToPostApprovalNeededComment = readyForReview &&
                    ((!reviewNeeded && visitor.errorFailedChecksMessages().size() == 1 && visitor.tooFewReviewers()) ||
                            visitor.errorFailedChecksMessages().isEmpty()) &&
                    integrationBlockers.isEmpty() &&
                    !statusMessage.contains(TEMPORARY_ISSUE_FAILURE_MARKER);`

erikj79: Right, I didn't think that through and this gets messy and I think it's because we aren't using the right abstraction. Can we just push something like this down to the visitor:
```
boolean hasErrors(boolean reviewNeeded)
```

zhaosongzs: Yes, sounds good

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -1470,7 +1470,7 @@ private void checkStatus() {
             var commitMessage = String.join("\n", commit.message());
 
             var readyToPostApprovalNeededComment = readyForReview &&
-                    (!reviewNeeded || visitor.errorFailedChecksMessages().isEmpty()) &&
+                    visitor.hasErrors(reviewNeeded) &&
```
erikj79, line 1473: Shouldn't this be negated?

zhaosongzs: Thanks for catching it. You are right. I remember I added "!" but Somehow it disappeared...

erikj79 (approved):

zhaosongzs:
Thanks for the review!
/integrate

PR #1717, opened 2025-05-02
2491: GitLabRepository::recentCommitComments can miss commits

erikj79:
The mechanism for polling GitLab for commit comments has a flaw that may cause it to start missing relevant comments. The underlying issue is that we are trusting timestamps on commits which may originate from a client machine, where the system clock may be wrong.

If a user changes their system clock to a date in the future and commits, that commit ends up having a future date. If this commit is then pushed to their personal fork and included in a merge request, then this commit will end up in a pr/X branch in the main/parent repository.

The rather convoluted mechanism for polling for commit comments (for the purpose of finding commit commands) for GitLab involves maintaining a mapping from commit title message to commit hash. This map is needed because the only way to find comments globally for a repository is to query for "events", and the event object only has the commit title, not the hash. Then for all possibly relevant events, we look up the commits in the map.

This map is initially generated from a local repository using a git command like this: "git rev-list <branch list>". After the first initialization, it is then updated using API calls to the server, querying for commits after a timestamp. The timestamp is the latest found already in the map. This is where the flaw lies. There is no guarantee that the timestamps are ever increasing, so once a commit with a faulty future timestamp ends up in the map, no new commits will ever be added when attempting to update it.

When I wrote this code, I think I assumed the timestamps would be strictly increasing because all official commits are generated by Skara. That still holds true for commits in repos and branches where we have strict Skara controlled pull requests. The problem here is that the API query does not limit commits to the specified branches, it gets all commits after the timestamp, which includes commits in pr/X branches as well as unreferenced commits.

A quick solution here would be to change the query to use branches. That would require doing a call per branch. I think a better solution would be to stick to one source for commits to the map, the local repository. It is already updated before each time we check for new commit comments. The challenge then is to identify the new commits to complement the map without having to rebuild it from scratch each time. I can't think of an exact way to accomplish this, but think we can get away with a good enough approximation. We can count the number of commits in the repo using "git rev-list --count <ref list>". Then when we iterate over the commits from commitMetaData iterator, we only go until the map contains the same number of commits as the repo has. At that point we know we have added all new commits. Most of the time, the rev-list loop (in topo-order) should return all the new commits first. It's only if we have a branched history within a branch that commits could be iterated in a different order, but worst case we just reprocess the unique ancestors of in each merge parent before we have passed all new commits.

While working on this I also realized that the current implementation has another flaw in that it assumes the list of branches given will always be the same. This is currently true in the affected bot, but that's not something the HostedRepository API should be assuming. I implemented a simple invalidation of the map in case the set of branches are different from the last call. This isn't an elegant solution but it at least keeps the API clean for now and if we actually need to handle this usecase, we can figure something else out then.

---------
### Progress

### Issue
 * [SKARA-2491](https://bugs.openjdk.org/browse/SKARA-2491): GitLabRepository::recentCommitComments can miss commits (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1717/head:pull/1717` \
`$ git checkout pull/1717`

`$ git checkout pull/1717` \
`$ git pull https://git.openjdk.org/skara.git pull/1717/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1717`

`$ git pr show -t 1717`

</details>

<a href="https://git.openjdk.org/skara/pull/1717.diff">https://git.openjdk.org/skara/pull/1717.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1717#issuecomment-2848230044)
</details>

### DIFF
--- a/forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabRepository.java
+++ b/forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabRepository.java
@@ -625,49 +625,63 @@ public class GitLabRepository implements HostedRepository {
     }
 
     /**
-     * Lazy fetching and caching of the commitTitleToCommits map. The first time
-     * this is called, the full map is built from the local repository. After that
-     * it's just refreshed from the server.
+     * Lazy fetching and caching of the commitTitleToCommits map. The map is only
+     * cached for one set of branches. If called with a different set, the old
+     * cache is invalidated and a new is rebuilt from scratch. In practice, an
+     * instance of this class is only ever called with the same set of branches.
+     * If that was to change we will need to rethink this.
+     * <p>
+     * The full map is built from the local repository. For incremental updates,
+     * it's refreshed until it contains the same number of commits as the local
+     * repository. The topo-order should guarantee that this gives us the same
+     * set of commits.
      */
     private final Map<String, SequencedSet<Hash>> commitTitleToCommits = new HashMap<>();
-    private boolean commitTitleToCommitsInitialized = false;
-    private ZonedDateTime lastCommitTime = ZonedDateTime.ofInstant(Instant.EPOCH, ZoneId.systemDefault());
+    private Set<Branch> commitMapBranchSet;
+    private int commitMapCount = 0;
 
     private Map<String, SequencedSet<Hash>> getCommitTitleToCommitsMap(ReadOnlyRepository localRepo, List<Branch> branches) {
-        if (!commitTitleToCommitsInitialized) {
-            try {
+        try {
+            Set<Branch> branchSet = Set.copyOf(branches);
+            if (!branchSet.equals(commitMapBranchSet)) {
+                log.info("Invalidating commitTitleToCommits map for branch set: " + branchSet + " old set: " + commitMapBranchSet);
+                commitTitleToCommits.clear();
+                commitMapCount = 0;
+                commitMapBranchSet = branchSet;
+            }
+            int newSize = localRepo.commitCount(branches);
+            if (newSize > commitMapCount) {
+                int sizeBefore = commitMapCount;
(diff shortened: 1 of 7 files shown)


forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabRepository.java
```diff
@@ -625,49 +625,63 @@ public List<CommitComment> recentCommitComments(ReadOnlyRepository localRepo, Se
...
+    private Set<Branch> commitMapBranchSet;
+    private int commitMapCount = 0;
 
     private Map<String, SequencedSet<Hash>> getCommitTitleToCommitsMap(ReadOnlyRepository localRepo, List<Branch> branches) {
-        if (!commitTitleToCommitsInitialized) {
-            try {
+        try {
+            Set<Branch> branchSet = Set.copyOf(branches);
+            if (!branchSet.equals(commitMapBranchSet)) {
+                log.info("Invalidating commitTitleToCommits map for branch set: " + branchSet + " old set: " + commitMapBranchSet);
```
zhaosongzs, line 647: Since commitMapBranchSet doesn't have an initial value, this log will also be printed during the initial run. Is this intentional?

zhaosongzs (commented):
Looks good overall!

forge/src/main/java/org/openjdk/skara/forge/gitlab/GitLabRepository.java
erikj79: Yes, kind of. I did note that it would happen and figured it was fine. I don't think it's worth complicating the code further to try to handle the first call differently.

zhaosongzs: I'm also ok with it

zhaosongzs (approved):

erikj79:
Thanks!

/integrate

PR #1719, opened 2025-05-28
2509

zhaosongzs:
In the pull request body, we have a 'Progress' part which will display the progresses of some jchecks, however, if the jchecks in the progress fails as warning, it will also be displayed in 'Warning' part. It will confuse the users.

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
@@ -413,6 +413,7 @@ class CheckRun {
     private void updateCheckBuilder(CheckBuilder checkBuilder, PullRequestCheckIssueVisitor visitor, List<String> additionalErrors) {
         if (visitor.isReadyForReview() && additionalErrors.isEmpty()) {
             checkBuilder.complete(true);
+
         } else {
             checkBuilder.title("Required");
             var summary = Stream.concat(visitor.errorFailedChecksMessages().stream().limit(MESSAGE_LIMIT), additionalErrors.stream().limit(MESSAGE_LIMIT))
@@ -726,7 +727,7 @@ class CheckRun {
             progressBody.append(warningListToText(integrationBlockers));
         }
 
-        var allWarnings = Stream.concat(visitor.warningFailedChecksMessages().stream(), warnings.stream()).toList();
+        var allWarnings = Stream.concat(visitor.hiddenWarningMessages().stream(), warnings.stream()).toList();
         if (!allWarnings.isEmpty()) {
             progressBody.append("\n\n### Warning");
             if (allWarnings.size() > 1) {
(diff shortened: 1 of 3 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
```diff
@@ -413,6 +413,14 @@ private List<String> botSpecificIntegrationBlockers(Map<Issue, Optional<IssueTra
     private void updateCheckBuilder(CheckBuilder checkBuilder, PullRequestCheckIssueVisitor visitor, List<String> additionalErrors) {
         if (visitor.isReadyForReview() && additionalErrors.isEmpty()) {
             checkBuilder.complete(true);
+            // It means whitespaceCheck failed as warning
+            if (!visitor.getAnnotations().isEmpty()) {
+                checkBuilder.title("Optional");
+                checkBuilder.summary("WhitespaceCheck failed as warning, it won't block the integration, but you'd better fix them.");
+                for (var annotation : visitor.getAnnotations()) {
+                    checkBuilder.annotation(annotation);
+                }
```
erikj79, line 422: I don't think we can assume that all `annotations` are whitespace warnings in this method. I don't think we need a summary text here, but it's hard to tell from just reading the code. The "Optional" title instead of "Required" and the list of annotations should be enough I think.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCheckIssueVisitor.java
```diff
@@ -86,6 +86,15 @@ List<String> warningFailedChecksMessages() {
         return warningFailedChecks.values().stream().flatMap(List::stream).toList();
     }
 
+    List<String> hiddenWarningMessages(){
```
erikj79, line 89:
```suggestion
    List<String> hiddenWarningMessages() {
```

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckRun.java
erikj79: I was trying to understand the role of the summary and it seems when there are errors, we are listing the 50 first errors in the summary. In GitHub, when creating a `check`, we are then adding the summary and each annotation as an `annotation`, so this makes sense. In GitLab, there is no concept of a `check` so we print in a comment, first the summary and then the annotations. This must result in duplicate listings for the 50 first whitespace errors in GitLab. That's what had me confused and why I thought the summary wasn't necessary.

So in this case, as you are pointing out, we need to put something there. How about:

```
"These warnings will not block integration"
```

zhaosongzs: Sure, I will just use what you said, it's cleaner.

zhaosongzs:
Thank you for the review!
/integrate

PR #1721, opened 2025-06-02
SKARA-2512

zhaosongzs:
Update windows-2019 to windows-2022 in .github/ci.yml

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,7 +105,7 @@ jobs:
 
   win:
     name: Windows x64
-    runs-on: "windows-2019"
+    runs-on: "windows-2022"
     needs: prerequisites
 
     steps:


.github/workflows/ci.yml
```diff
@@ -105,7 +105,7 @@ jobs:
 
   win:
     name: Windows x64
-    runs-on: "windows-2019"
+    runs-on: "windows-2022"
```
erikj79, line 108: I think we should go straight to the latest 2025 if it works.

zhaosongzs: Sure

erikj79: Looking at the documentation for these configurations, windows-2025 does not have Mercurial anymore. I think it's ok to not verify Mercurial dependent tests on Windows (or perhaps at all), so we could add assumes to those tests and check if Mercurial is available.

erikj79 (commented):
My idea was to try to execute `hg` once in `@BeforeClass` and if it failed, set a boolean to skip testcases with.

zhaosongzs:
> I think this looks good, but please wait for GHA to finish.

Sure, won't integrate it until GHA finished

zhaosongzs:
Thank you for the suggestions and review!
/integrate

PR #1726, opened 2025-07-17
SKARA-2527

zhaosongzs:
When determining the author of a commit, if the author of pr doesn't have OpenJDK Id, the bot falls back to use the author from the last commit of the pr, in this way, the bot may give credit to the wrong person. 
In this patch, if the author of pr doesn't have OpenJDK Id, the bot will try to parse full name and email from the forge first, if it fails, then it will fall back to use the author of the last commit.


zhaosongzs:
Thank you for the review!
/integrate

PR #1730, opened 2025-08-07
SKARA-2557

zhaosongzs:
This patch is trying to let Skara  bot  explicitly set resolution to "Fixed" when resolving Jira issues.

### DIFF
--- a/issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraIssue.java
+++ b/issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraIssue.java
@@ -48,6 +48,7 @@ public class JiraIssue implements IssueTrackerIssue {
     private static final List<String> VALID_RESOLUTIONS = List.of("Fixed", "Delivered");
 
     private List<Label> labels;
+    private Map<String, String> availableTransitions;
 
     JiraIssue(JiraProject jiraProject, RestRequest request, JSONValue json) {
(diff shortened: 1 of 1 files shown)


issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraIssue.java
```diff
@@ -237,18 +238,24 @@ private Map<String, String> availableTransitions() {
...
 
-    private void performTransition(String id) {
+    private void performTransition(String state) {
+        var id = availableTransitions.get(state);
         var query = JSON.object()
-                        .put("transition", JSON.object()
-                                               .put("id", id));
+                .put("transition", JSON.object()
+                        .put("id", id));
+        if (state.equals("RESOLVED")) {
```
erikj79, line 246: This method is called with `"Resolved"` not `"RESOLVED"`. To avoid relying on string comparisons here, could we make all the calls to this method use constants, or perhaps even an internal enum?

erikj79 (commented):
I think this change warrants a new manual integration test.

issuetracker/src/main/java/org/openjdk/skara/issuetracker/jira/JiraIssue.java
zhaosongzs: Sure

erikj79 (approved):

PR #1735, opened 2025-08-25
2065: Update PR labels when new files are touched

zhaosongzs:
This patch is trying to make the pr bot be able to update PR labels when new files are touched.
The main idea is from Erik. For a new PR, the bot will run a labelerWorkItem to auto label the PR first, and the commit hash will be stored in a comment.

With this patch, here are some new behaviors of the bot:
(1) The user didn't issue manual label command before auto labeling, LabelerWorkItem will do the initial auto labeling and store the commit hash in the comment. 
(2) If the user issued manual label command before auto labeling, the initial auto labeling will be skipped, the bot would still post a comment and store the hash in the comment.
(3) The user pushes a new commit or few commits to a pr that already auto labeled, the bot will evaluate the diff between stored hash and current head, then add new labels or upgrading labels to group labels, in the end, update the stored hash in the comment.
(4) The user force pushes to the pr that already auto labeled, the bot will evaluate the diff between baseHash and current head.
(5) The user issues a command to add a label to the pr(or even the user add the label via the forge UI), the bot will check if the labels can be upgraded to group labels.

The side effect of this feature I can imagine is that a user thinks a file is not related to a component and removed it manually, but later, every time when he touches the file, the label will be added back.

---------
### Progress

### Issue
 * [SKARA-2065](https://bugs.openjdk.org/browse/SKARA-2065): Update PR labels when new files are touched (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1735/head:pull/1735` \
`$ git checkout pull/1735`

`$ git checkout pull/1735` \
`$ git pull https://git.openjdk.org/skara.git pull/1735/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1735`

`$ git pr show -t 1735`

</details>

<a href="https://git.openjdk.org/skara/pull/1735.diff">https://git.openjdk.org/skara/pull/1735.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1735#issuecomment-3229478499)
</details>

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/LabelerWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/LabelerWorkItem.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -26,16 +26,22 @@ import java.time.ZonedDateTime;
 import org.openjdk.skara.bot.WorkItem;
 import org.openjdk.skara.forge.*;
 import org.openjdk.skara.issuetracker.Comment;
+import org.openjdk.skara.vcs.Hash;
 import org.openjdk.skara.vcs.Repository;
 
 import java.io.*;
 import java.util.*;
 import java.util.function.Consumer;
+import java.util.logging.Logger;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 public class LabelerWorkItem extends PullRequestWorkItem {
     protected static final String INITIAL_LABEL_MESSAGE = "<!-- PullRequestBot initial label help comment -->";
     private static final String LABEL_COMMIT_MARKER = "<!-- PullRequest Bot label commit '%s' -->";
+    protected static final Pattern LABEL_COMMIT_PATTERN = Pattern.compile("<!-- PullRequest Bot label commit '(.*?)' -->");
+    private static final Logger log = Logger.getLogger("org.openjdk.skara.bots.pr");
 
     LabelerWorkItem(PullRequestBot bot, String prId, Consumer<RuntimeException> errorHandler,
             ZonedDateTime prUpdatedAt) {
@@ -123,15 +129,73 @@ public class LabelerWorkItem extends PullRequestWorkItem {
 
     @Override
     public Collection<WorkItem> prRun(ScratchArea scratchArea) {
-        if (bot.isAutoLabelled(pr)) {
-            return List.of();
-        }
-
+        // If no label configuration, return early
         if (bot.labelConfiguration().allowed().isEmpty()) {
             bot.setAutoLabelled(pr);
             return List.of();
         }
 
+        // Updating labels when new files are touched
+        if (bot.isAutoLabelled(pr)) {
+            try {
+                var labelAdded = false;
+                var oldLabels = new HashSet<>(pr.labelNames());
+                var newLabels = new HashSet<>(pr.labelNames());
+
+                var path = scratchArea.get(pr.repository());
+                var seedPath = bot.seedStorage().orElse(scratchArea.getSeeds());
+                var hostedRepositoryPool = new HostedRepositoryPool(seedPath);
+                var localRepo = PullRequestUtils.materialize(hostedRepositoryPool, pr, path);
+
+                var labelComment = findComment(prComments(), INITIAL_LABEL_MESSAGE);
+
+                if (labelComment.isPresent()) {
+                    var line = labelComment.get().body().lines()
+                            .map(LABEL_COMMIT_PATTERN::matcher)
+                            .filter(Matcher::find)
+                            .findFirst();
+
+                    if (line.isPresent()) {
+                        var evaluatedCommitHash = line.get().group(1);
+                        var changedFiles = PullRequestUtils.changedFiles(pr, localRepo, new Hash(evaluatedCommitHash));
+                        var newLabelsNeedToBeAdded = bot.labelConfiguration().label(changedFiles);
+                        newLabels.addAll(newLabelsNeedToBeAdded);
+
+                        var upgradedLabels = bot.labelConfiguration().upgradeLabelsToGroups(newLabels);
+                        newLabels.addAll(upgradedLabels);
+                        newLabels.removeIf(label -> !upgradedLabels.contains(label));
+                    }
+
+                    for (var newLabel : newLabels) {
+                        if (!oldLabels.contains(newLabel)) {
+                            log.info("Adding label " + newLabel);
+                            pr.addLabel(newLabel);
+                            labelAdded = true;
+                        }
+                    }
+
+                    for (var oldLabel : oldLabels) {
+                        if (!newLabels.contains(oldLabel)) {
+                            log.info("Removing label " + oldLabel);
+                            pr.removeLabel(oldLabel);
+                        }
+                    }
+
+                    pr.updateComment(labelComment.get().id(), labelComment.get().body().replaceAll(
+                            "(<!-- PullRequest Bot label commit ')[^']*(' -->)",
+                            "$1" + pr.headHash().toString() + "$2"
+                    ));
+                }
+                if (labelAdded) {
+                    return needsRfrCheck(new ArrayList<>(newLabels));
+                }
+            } catch (IOException e) {
+                throw new UncheckedIOException(e);
+            }
+            return List.of();
+        }
+
+        // Initial auto labeling
         var comments = prComments();
         var labelNames = pr.labelNames();
         var manuallyAdded = LabelTracker.currentAdded(pr.repository().forge().currentUser(), comments);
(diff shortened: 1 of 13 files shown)


zhaosongzs:
Another thing is that for the current pull requests in openjdk/jdk, if the auto labelling was skipped, after this patch is deployed, the bot will add the comment "A manual label command was issued before auto-labeling, so auto-labeling was skipped." to the pull requests as well.

zhaosongzs:
> I understand wanting to optimize for performance by putting this in CheckRun, but it really comes at a cost of maintainability IMO. Having two different WorkItems share the responsibility of adding labels in different situations seems bad. We also have the race situation between labels and comment which isn't easily solved in a nice way within CheckRun.
> 
> Are you sure we would need to spawn another CheckWorkItem after the labeler has run? I can see that it would be forced if `rfr` wasn't set yet and that it would be triggered if we actually added a label, but for most situations where nothing changed, the only thing that would happen is that a comment got updated. That comment would have the bot user as author, and those are filtered out already for the updated check, aren't they? I may be missing something.
> 
> If we do this in the LabelerWorkItem, we can more easily control the order of label changes and comment update, to avoid the risk of inconsistent states.

Yeah, I was thinking the same, we need another run after the label is added, now I am agree with you, seems like it's not that expensive. I will move the logic to LabelerWorkItem. Thanks!

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
```diff
@@ -188,12 +188,17 @@ public Collection<WorkItem> prRun(ScratchArea scratchArea) {
...
-                return List.of(new LabelerWorkItem(bot, prId, errorHandler, triggerUpdatedAt));
-            } else {
+            // If there is no label configuration, don't generate LabelerWorkItem
+            if (bot.labelConfiguration().allowed().isEmpty()) {
+                bot.setAutoLabelled(pr);
                 return List.of();
             }
+
+            if (!pr.isClosed()) {
+                return List.of(new LabelerWorkItem(bot, prId, errorHandler, triggerUpdatedAt));
```
erikj79, line 198: At this stage in PullRequestCommandWorkItem, we have already loaded all PR comments, so I think it would be worth checking if there is any need for scheduling a LabelerWorkItem. What I mean is, we can check if the head hash of the PR is different from the last labeler commit marker. This could be implemented in a static method on `LabelereWorkItem` to keep the details in one place.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/LabelerWorkItem.java
```diff
@@ -71,69 +78,135 @@ private void updateLabelMessage(List<Comment> comments, List<String> newLabels)
...
+        // Updating labels when new files are touched
+        if (bot.isAutoLabelled(pr)) {
+            try {
+                var oldLabels = new HashSet<>(pr.labelNames());
+                var newLabels = new HashSet<>(pr.labelNames());
+
+                var path = scratchArea.get(pr.repository());
+                var seedPath = bot.seedStorage().orElse(scratchArea.getSeeds());
+                var hostedRepositoryPool = new HostedRepositoryPool(seedPath);
+                var localRepo = PullRequestUtils.materialize(hostedRepositoryPool, pr, path);
```
erikj79, line 152: This is repeated further down, maybe worth breaking out into a method `materializeRepo`?

bots/pr/src/main/java/org/openjdk/skara/bots/pr/LabelerWorkItem.java
```diff
@@ -71,69 +78,135 @@ private void updateLabelMessage(List<Comment> comments, List<String> newLabels)
...
+
+                var labelComment = findComment(prComments(), INITIAL_LABEL_MESSAGE);
+
+                if (labelComment.isPresent()) {
```
erikj79, line 156: If labelComment isn't present, something has gone wrong. Could we do something to try to recover or at least log it?

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
zhaosongzs: When I was working on it, I have the same thought with you, I added the check and did some tests, later I found there is a case which is that the user use the "/label" command to add the label, in this case, we need LabelerWorkItem to check whether the labels can be upgraded to group labels.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/LabelerWorkItem.java
zhaosongzs: Sure, will do.

zhaosongzs: I found there is already a method `IntegrateCommand#materializeLocalRepo`

bots/pr/src/main/java/org/openjdk/skara/bots/pr/PullRequestCommandWorkItem.java
zhaosongzs: Or there is another Option, we can add the check here. And in pullRequestWorkItem, after processing a label command, we can schedule a LabelerWorkItem

erikj79: Hm, that could work, but it would also mean that we need to schedule an initial LabelerWorkItem when bot is starting up, or rather, we need to let the very first one through regardless of commit hash. This is in case a LabelCommand was issued and handled right before bot was restarted. 

The alternative would be to compare dates on comments. If the label "command handled" comment has a later date than the last update of the auto labeler comment, then we need to run. If we do that, then the labeler needs to always touch the comment to mark when it was last run. Not sure which is best. If we end up spawning a lot of LabelereWorkItems, it could turn into a lot of comment updates.

zhaosongzs: I think the first one is pretty easy, we just need to check if the pr has been marked as auto labeled or not.

magicus:
I made some comments about this PR in the test bug https://github.com/openjdk/playground/pull/239#issuecomment-3236075737. I realized afterwards that this was kind of a stupid place to put it. :) So here it is again:

> @zhaosongzs A manual label command was issued before auto-labeling, so auto-labeling was skipped.

@zhaosongzs Oh, I'm not so sure about this... If the bots are slow or get stuck, people might give `/label` commands before the bots can do the auto labeling. I don't think that should block auto-labeling. But we could give a warning, showing the diff between the manual labeling and the autolabeling. And if there is no diff, then we can just be quiet, since that might just have been someone getting tired of waiting for the bots. Otherwise we can say like "Manual labels set `build`, `core-libs`. Automatic labelling would not have set `build`, but is keeping it due to manual label. Automatical labelling added `client-libs`, `panama`." (but slightly better phrased).

And also, just to be super clear, giving manual `/label` commands at some point, should not block the automatic label update when new files are added.

erikj79:
What Magnus said got me thinking. Since we are now redoing auto labeling every time the code changes, I think the concept of having the bot class track which PRs have been auto labeled doesn't make much sense. A PR is either up to date with the labeler, or it isn't. There shouldn't be anything special about the first run really. In CheckRun, to check if the PR has been auto labeled or not, it can check for the existence of an auto label comment. In LabelerWorkItem, I think the code paths could be more unified to make it clear that there isn't a big difference.

I'm thinking that when running the LabelerWorkItem again, due to a new commit, if nothing new is found, we just silently update the commit hash in the existing (last) auto label comment. If however we find that a new label needs to be added, we may want to create a new comment to make it clearer that a new action was taken, without erasing/hiding the history of the previous comment. That would make it easier to incorporate the kind of messaging Magnus is asking for.

For the case of running LabelerWorkItem after executing a LabelCommand, just to collapse groups, I'm starting to think that it would be better to just have the LabelCommand automatically collapse groups for you. The implementation might be shared, but by having it happen through the LabelCommand, we don't need to worry about updating the auto label comments, or racing against bot restarts. In that case the LabelCommand reply would include a message of combining the groups.

I might have missed something, but these are my current thoughts.

zhaosongzs:
> What Magnus said got me thinking. Since we are now redoing auto labeling every time the code changes, I think the concept of having the bot class track which PRs have been auto labeled doesn't make much sense. A PR is either up to date with the labeler, or it isn't. There shouldn't be anything special about the first run really. In CheckRun, to check if the PR has been auto labeled or not, it can check for the existence of an auto label comment. In LabelerWorkItem, I think the code paths could be more unified to make it clear that there isn't a big difference.
> 
> I'm thinking that when running the LabelerWorkItem again, due to a new commit, if nothing new is found, we just silently update the commit hash in the existing (last) auto label comment. If however we find that a new label needs to be added, we may want to create a new comment to make it clearer that a new action was taken, without erasing/hiding the history of the previous comment. That would make it easier to incorporate the kind of messaging Magnus is asking for.
> 
> For the case of running LabelerWorkItem after executing a LabelCommand, just to collapse groups, I'm starting to think that it would be better to just have the LabelCommand automatically collapse groups for you. The implementation might be shared, but by having it happen through the LabelCommand, we don't need to worry about updating the auto label comments, or racing against bot restarts. In that case the LabelCommand reply would include a message of combining the groups.
> 
> I might have missed something, but these are my current thoughts.

Make sense to me. I am thinking about should we introduce a command to let user be able to disable and enable the auto labeling, like `/autolabel off` and `/autolabel on`.

PR #1738, opened 2025-09-16
SKARA-2586

erikj79:
In some projects, other than the JDK, we would like to use Skara in a simplified Jira workflow, where we use the multi value aspect of the Jira FixVersion field instead of creating backport records. To support this I propose to add a configuration option in the IssueNotifier "multifixversion", with default value false, which corresponds to the current behavior with backports. When set to true, then instead of creating backports, the notifier would just pile on additional fix versions when a commit is integrated into multiple different branches.

In addition to this, I would like to also add handling of a "branch-" prefix for a version strings. This would mimic the current "repo-" prefix which is already in use for JDK projects.


zhaosongzs (approved):
Looks good

erikj79:
Thanks! I've manually verified a full workflow now, but I will hold off on this until the labeling change is under control.

PR #1740, opened 2025-09-17
2588: Auto Labeling should ignore merge commits

zhaosongzs:
Currently, the auto labeling doesn't ignore merge commits, so when user merge master back into their feature branch, it's possible that unrelated labels are added to the pull request.
This patch is trying to let the bot ignore all the merge commits and the commits brought in by the merges.

---------
### Progress

### Issue
 * [SKARA-2588](https://bugs.openjdk.org/browse/SKARA-2588): Auto Labeling should ignore merge commits (**Bug** - P3)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1740/head:pull/1740` \
`$ git checkout pull/1740`

`$ git checkout pull/1740` \
`$ git pull https://git.openjdk.org/skara.git pull/1740/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1740`

`$ git pr show -t 1740`

</details>

<a href="https://git.openjdk.org/skara/pull/1740.diff">https://git.openjdk.org/skara/pull/1740.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1740#issuecomment-3303708871)
</details>


erikj79 (commented):
There are problems with this approach. 

1. Say the user pushes two commits, in the first it touches file A, then realizes it was a mistake and reverts that change in the second commit. If those two commits are pushed together, then before this patch, file A wouldn't appear in the diff, but with this patch it does.
2. I don't think we can make the assumption that the first parent will always be the correct one to follow in a merge commit.

I think I have figured out a better algorithm, but I'm not sure yet:
First find all commits that are ancestors of the PR head, but are not present in the target branch, this is the complete set of PR commits. Then iterate backwards from `to` until you find a merge commit or the `from` commit. Take the diff from all commits passed this way combined and extract changed files. Then if this was a merge commit, pick the parent that is in the complete set we collected earlier and continue this way until reaching the `from` commit.

This will group any consecutive commits between merges to at least alleviate most of the potential issues from 1, but not all, so I'm still unsure if this is good enough. It will guarantee that we only consider commits in the PR and not anything already present in the target branch. 

Maybe for the initial labeling, we can skip this algorithm completely and just diff PR head with mergebase. At that time, that diff is guaranteed to be correct at least.

erikj79:
> I think I have figured out a better algorithm, but I'm not sure yet:

I did some more thinking and I think I have a much simpler proposal (as I already shared in person, but writing it here for transparency).

Instead of creating diffs between series of commits in the PR branch, we should always just diff against the merge-base with the target branch. The first auto labeling is the base case, where we just do exactly this and add labels based on the changed files. Then for additional labeling, we first repeat the diff from the previous run, then we create the diff from the current PR head to its merge-base. This gives us two sets of changed files. We subtract the set from the previous labeling from the set from the current PR head and calculate additional labels from that.

erikj79:
I think this will require extra care with dependent PRs. Since the target branch of a PR can change (and will for dependent PRs), we will need to record the merge-base we used in addition to the head hash.

PR #1768, opened 2026-04-09
2708: Two-Reviewers requirement shouldn't apply to Backport/Merge PRs

zhaosongzs:
In [SKARA-2573](https://bugs.openjdk.org/browse/SKARA-2573), I implemented a feature that if the PR touches certain areas, the reviewers count will be automatically increased to two.

Now, this requirement also applies to Merge/Backport PRs, but it shouldn't.

If there is any user issued reviewers command, then the two reviewers check will be skipped.
The two reviewers requirement will be applied to PRs if the PR touches certain areas.
If the PR is converted to a Merge/Backport PR later, the requirement will be cleared.
If the PR is converted back to a normal PR again, the requirement will be added back.

I didn’t think it will be too complicated, but somehow, I made the logic a bit complicated…

---------
- [x] I confirm that I make this contribution in accordance with the [OpenJDK Interim AI Policy](https://openjdk.org/legal/ai).

---------
### Progress

### Issue
 * [SKARA-2708](https://bugs.openjdk.org/browse/SKARA-2708): Two-Reviewers requirement shouldn't apply to Backport/Merge PRs (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1768/head:pull/1768` \
`$ git checkout pull/1768`

`$ git checkout pull/1768` \
`$ git pull https://git.openjdk.org/skara.git pull/1768/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1768`

`$ git pr show -t 1768`

</details>

<a href="https://git.openjdk.org/skara/pull/1768.diff">https://git.openjdk.org/skara/pull/1768.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1768#issuecomment-4220438175)
</details>

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
@@ -182,7 +182,7 @@ class CheckWorkItem extends PullRequestWorkItem {
             } else if (latestTwoReviewersComment.get().body().contains(TWO_REVIEWERS_APPLIED_MARKER)) {
                 var marker = ReviewersTracker.setReviewersMarker(0, "authors", ReviewersTracker.Source.BOT);
                 var prType = PullRequestUtils.isMerge(pr) ? "merge" : "backport";
-                var reviewersClearedComment = pr.addComment("This PR is now a " + prType + " PR, the extra reviewers requirement has been cleared.\n"
+                var reviewersClearedComment = pr.addComment("This is now a " + prType + " PR, the extra reviewers requirement has been cleared.\n"
                         + marker + "\n" + TWO_REVIEWERS_CLEARED_MARKER);
                 return Stream.concat(comments.stream(), Stream.of(reviewersClearedComment)).toList();
             }
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
@@ -182,7 +182,8 @@ class CheckWorkItem extends PullRequestWorkItem {
                 return comments;
             } else if (latestTwoReviewersComment.get().body().contains(TWO_REVIEWERS_APPLIED_MARKER)) {
                 var marker = ReviewersTracker.setReviewersMarker(0, "authors", "bot");
-                var reviewersClearedComment = pr.addComment("This PR is now a backport or merge PR, so the extra reviewers requirement has been cleared.\n"
+                var prType = PullRequestUtils.isMerge(pr) ? "merge" : "backport";
+                var reviewersClearedComment = pr.addComment("This PR is now a " + prType + " PR, so the extra reviewers requirement has been cleared.\n"
                         + marker + "\n" + TWO_REVIEWERS_CLEARED_MARKER);
                 return Stream.concat(comments.stream(), Stream.of(reviewersClearedComment)).toList();
             }
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
@@ -182,7 +182,7 @@ class CheckWorkItem extends PullRequestWorkItem {
             } else if (latestTwoReviewersComment.get().body().contains(TWO_REVIEWERS_APPLIED_MARKER)) {
                 var marker = ReviewersTracker.setReviewersMarker(0, "authors", ReviewersTracker.Source.BOT);
                 var prType = PullRequestUtils.isMerge(pr) ? "merge" : "backport";
-                var reviewersClearedComment = pr.addComment("This PR is now a " + prType + " PR, so the extra reviewers requirement has been cleared.\n"
+                var reviewersClearedComment = pr.addComment("This PR is now a " + prType + " PR, the extra reviewers requirement has been cleared.\n"
                         + marker + "\n" + TWO_REVIEWERS_CLEARED_MARKER);
                 return Stream.concat(comments.stream(), Stream.of(reviewersClearedComment)).toList();
             }
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
@@ -63,6 +63,8 @@ class CheckWorkItem extends PullRequestWorkItem {
     private static final Pattern BACKPORT_ISSUE_TITLE_PATTERN = Pattern.compile("^Backport\\s*(?:(?<prefix>[A-Za-z][A-Za-z0-9]+)-)?(?<id>[0-9]+)\\s*$", Pattern.CASE_INSENSITIVE);
     private static final Pattern METADATA_COMMENTS_PATTERN = Pattern.compile(
             "<!-- (?:backport)|(?:(add|remove) (?:contributor|reviewer))|(?:summary: ')|(?:solves: ')|(?:additional required reviewers)|(?:jep: ')|(?:csr: ')|(?:trailer:)");
+    private static final String TWO_REVIEWERS_APPLIED_MARKER = "<!-- two-reviewers applied -->";
+    private static final String TWO_REVIEWERS_CLEARED_MARKER = "<!-- two-reviewers cleared -->";
     private static final String ELLIPSIS = "…";
     protected static final String FORCE_PUSH_MARKER = "<!-- force-push suggestion -->";
     protected static final String FORCE_PUSH_SUGGESTION= """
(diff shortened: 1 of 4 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/ReviewersTracker.java
```diff
@@ -28,15 +28,18 @@
...
 class ReviewersTracker {
-    private static final String REVIEWERS_MARKER = "<!-- additional required reviewers id marker (%d) (%s) -->";
+    private static final String REVIEWERS_MARKER = "<!-- additional required reviewers id marker (%d) (%s) (%s)-->";
     private static final Pattern REVIEWERS_MARKER_PATTERN = Pattern.compile(
-            "<!-- additional required reviewers id marker \\((\\d+)\\) \\((\\w+)\\) -->");
+            "<!-- additional required reviewers id marker \\((\\d+)\\) \\((\\w+)\\)(?: \\(([^)]*)\\))?\\s*-->");
 
     static String setReviewersMarker(int numReviewers, String role) {
-        return String.format(REVIEWERS_MARKER, numReviewers, role);
+        return setReviewersMarker(numReviewers, role, "user");
```
erikj79: Can we make "user" a constant?

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -157,33 +159,60 @@ private List<Comment> ensureTwoReviewersLabelMarker(List<Comment> comments) {
...
-
-        var marker = ReviewersTracker.setReviewersMarker(2, "authors");
-
-        var matchingLabelsList = matchingLabels.stream()
-                .map(label -> "`" + label + "`")
-                .collect(Collectors.joining(", "));
+        var latestTwoReviewersComment = comments.stream()
+                .filter(comment -> comment.author().equals(pr.repository().forge().currentUser()))
+                .filter(comment -> comment.body().contains(TWO_REVIEWERS_APPLIED_MARKER) || comment.body().contains(TWO_REVIEWERS_CLEARED_MARKER))
+                .reduce((a, b) -> b);
```
erikj79: It's not important, but I would probably run the stream in reverse and do a findFirst.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -157,33 +159,60 @@ private List<Comment> ensureTwoReviewersLabelMarker(List<Comment> comments) {
...
+                || BACKPORT_HASH_TITLE_PATTERN.matcher(pr.title()).matches() || PullRequestUtils.isMerge(pr)) {
+            // Backport or Merge PR
+            if (latestTwoReviewersComment.isEmpty()) {
+                return comments;
+            } else if (latestTwoReviewersComment.get().body().contains(TWO_REVIEWERS_CLEARED_MARKER)) {
+                return comments;
+            } else if (latestTwoReviewersComment.get().body().contains(TWO_REVIEWERS_APPLIED_MARKER)) {
+                var marker = ReviewersTracker.setReviewersMarker(0, "authors", "bot");
+                var reviewersClearedComment = pr.addComment("This PR is now a backport or merge PR, so the extra reviewers requirement has been cleared.\n"
+                        + marker + "\n" + TWO_REVIEWERS_CLEARED_MARKER);
```
erikj79: I think it's worth specializing this message for each case, merge or backport.

I also pictured that we would edit the existing message in this case, but I'm not sure if that is better. The most likely reason for this to happen is that the user made a mistake with the title when creating the PR, and in that case we should probably avoid adding even more extra comments.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -157,33 +159,60 @@ private List<Comment> ensureTwoReviewersLabelMarker(List<Comment> comments) {
...
+
+                var matchingLabelsList = matchingLabels.stream()
+                        .map(label -> "`" + label + "`")
+                        .collect(Collectors.joining(", "));
+
+                var reviewersAppliedComment = pr.addComment(
+                        "The total number of required reviews for this PR has been set to 2 based on the presence of " +
+                                labelsNoun + ": " + matchingLabelsList + ". " +
+                                "This can be overridden with the `/reviewers` command.\n" +
+                                marker + "\n" + TWO_REVIEWERS_APPLIED_MARKER);
```
erikj79, line 212: Same here about reusing the existing comment if it exists.

bots/pr/src/test/java/org/openjdk/skara/bots/pr/LabelerTests.java
```diff
@@ -553,6 +553,58 @@ void autoLabelAppliesTwoReviewersRule(TestInfo testInfo) throws IOException {
         }
     }
 
+    @Test
+    void twoReviewersRuleClearedForBackportPR(TestInfo testInfo) throws IOException {
```
erikj79, line 558: I would like to see more tests.
* Convert to Merge and back.
* Repeat the conversion to backport/merge but add an explicit /reviewers command before converting back to normal.

erikj79:
> I didn’t think it will be too complicated, but somehow, I made the logic a bit complicated…

I think the logic is pretty clearly laid out.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/ReviewersTracker.java
zhaosongzs: Yes, it's better, was thinking about making it a constant later but somehow forgot...

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
zhaosongzs: Is there any easy way to run the stream in reverse?
AFAIK, we need to reverse the collection first...

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
zhaosongzs: I also thought about editing the existing message and wrote a patch for it, but I found that solution has some issues
(1) We not only need to edit the existing comment in GitHub, we also need to edit that comment in "comments", it's a cache we will use later. (It's not hard but cumbersome)
(2) Without extra comments, the edited message it not that obvious to user

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
zhaosongzs: If we choose to reuse the existing comment, then we can just do findFirst()...

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
zhaosongzs: The other thing is that if we reusing the existing comment, it means the comment should be updated to present the current state of the PR(like merge ready comment), So if there is any user issued reviewers command in the PR, what should we say in the comment.. "User issued reviewers comment detected, the two reviewers requirement has been overridden" ?

erikj79: I think you have a point here, reusing the comment is just too complicated. I'm fine with creating a new one.

bots/pr/src/test/java/org/openjdk/skara/bots/pr/LabelerTests.java
zhaosongzs: We indeed need more tests... I found an issue when adding another test...

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
erikj79: `List::reversed` is new in JDK 21. It's not a copy, just a reversed view.
```
comments.reversed().stream()
```

zhaosongzs: Ah, thanks!

bots/pr/src/test/java/org/openjdk/skara/bots/pr/LabelerTests.java
```diff
@@ -555,6 +556,86 @@ void autoLabelAppliesTwoReviewersRule(TestInfo testInfo) throws IOException {
...
+            // "backport" the new file to the master branch
+            localRepo.checkout(localRepo.defaultBranch());
+            var editBranch = localRepo.branch(masterHash, "edit");
+            localRepo.checkout(editBranch);
+            var newFile2 = localRepo.root().resolve("hotspot.txt");
+            Files.writeString(newFile2, "hello");
+            localRepo.add(newFile2);
+            var editHash = localRepo.commit("Backport", "duke", "duke@openjdk.org");
+            localRepo.push(editHash, author.authenticatedUrl(), "refs/heads/edit", true);
+            //var pr = credentials.createPullRequest(author, "master", "edit", "Backport " + releaseHash.hex(), List.of());
```
erikj79: Remove?

bots/pr/src/test/java/org/openjdk/skara/bots/pr/LabelerTests.java
```diff
@@ -555,6 +556,86 @@ void autoLabelAppliesTwoReviewersRule(TestInfo testInfo) throws IOException {
...
+            var pr = credentials.createPullRequest(author, "master", "edit", "Wrong title", List.of("body"));
+
+            TestBotRunner.runPeriodicItems(bot);
```
erikj79: Are you sure that you need to run periodic items twice? Test will run quite a bit faster if we don't do this unnecessarily.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
```diff
@@ -157,33 +159,61 @@ private List<Comment> ensureTwoReviewersLabelMarker(List<Comment> comments) {
...
+                || BACKPORT_HASH_TITLE_PATTERN.matcher(pr.title()).matches() || PullRequestUtils.isMerge(pr)) {
+            // Backport or Merge PR
+            if (latestTwoReviewersComment.isEmpty()) {
+                return comments;
+            } else if (latestTwoReviewersComment.get().body().contains(TWO_REVIEWERS_CLEARED_MARKER)) {
+                return comments;
+            } else if (latestTwoReviewersComment.get().body().contains(TWO_REVIEWERS_APPLIED_MARKER)) {
+                var marker = ReviewersTracker.setReviewersMarker(0, "authors", ReviewersTracker.Source.BOT);
+                var prType = PullRequestUtils.isMerge(pr) ? "merge" : "backport";
+                var reviewersClearedComment = pr.addComment("This PR is now a " + prType + " PR, so the extra reviewers requirement has been cleared.\n"
```
erikj79:
```suggestion
                var reviewersClearedComment = pr.addComment("This is now a " + prType + " PR, the extra reviewers requirement has been cleared.\n"
```

bots/pr/src/test/java/org/openjdk/skara/bots/pr/LabelerTests.java
zhaosongzs: Ah, yes

bots/pr/src/test/java/org/openjdk/skara/bots/pr/LabelerTests.java
zhaosongzs: I was just thinking adding another round of periodic items to make sure the next round won't do something wrong...  I will remove the unnecessary runs.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/CheckWorkItem.java
erikj79: I think having `PR` twice in the sentence is a bit much, which is why I suggested removing the first one. I think that reads better.

zhaosongzs: Oh Sorry, I didn't notice you removed the first PR. Will fix it

erikj79 (approved):

zhaosongzs:
Thank you for the review and all suggestions!
/integrate

PR #1781, opened 2026-07-10
2762: Prevent individual external PR failures from blocking IssueBot processing

zhaosongzs:
IssueBot previously processed all PRs associated with an updated issue through a single stream. If fetching any one external PR failed, the exception terminated the whole stream. Therefore, valid PRs were not processed and no CheckWorkItems were created for them. The periodic batch also did not complete normally.

We should process each associated PR independently instead. When a PR query fails, log the failure and schedule the issue for retry in a later polling round. Continue creating CheckWorkItems for the remaining PRs in the current round, so a transient or broken external PR cannot block unrelated valid PRs.

---------
- [x] I confirm that I make this contribution in accordance with the [OpenJDK Interim AI Policy](https://openjdk.org/legal/ai).

---------
### Progress
- \[x\] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- \[x\] Change must not contain extraneous whitespace

### Issue
 * [SKARA-2762](https://bugs.openjdk.org/browse/SKARA-2762): Prevent individual external PR failures from blocking IssueBot processing (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/skara.git pull/1781/head:pull/1781` \
`$ git checkout pull/1781`

`$ git checkout pull/1781` \
`$ git pull https://git.openjdk.org/skara.git pull/1781/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1781`

`$ git pr show -t 1781`

</details>

<a href="https://git.openjdk.org/skara/pull/1781.diff">https://git.openjdk.org/skara/pull/1781.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/skara/pull/1781#issuecomment-4938051941)
</details>

### DIFF
--- a/bots/pr/src/main/java/org/openjdk/skara/bots/pr/IssueBot.java
+++ b/bots/pr/src/main/java/org/openjdk/skara/bots/pr/IssueBot.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2023, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -25,7 +25,6 @@ package org.openjdk.skara.bots.pr;
 import org.openjdk.skara.bot.Bot;
 import org.openjdk.skara.bot.WorkItem;
 import org.openjdk.skara.forge.HostedRepository;
-import org.openjdk.skara.issuetracker.Issue;
 import org.openjdk.skara.issuetracker.IssueProjectPoller;
 import org.openjdk.skara.issuetracker.IssueProject;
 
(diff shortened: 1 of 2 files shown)


bots/pr/src/main/java/org/openjdk/skara/bots/pr/IssueBot.java
```diff
@@ -87,17 +86,26 @@ public List<WorkItem> getPeriodicItems() {
...
+                        var pr = repository.pullRequest(record.prId());
+                        if (pr.isOpen()) {
+                            // This will mix time stamps from the IssueTracker and the Forge hosting PRs, but it's the
+                            // best we can do.
+                            items.add(CheckWorkItem.fromIssueBot(pullRequestBotMap.get(pr.repository().name()), pr.id(),
+                                    e -> poller.retryIssue(issue), issue.updatedAt()));
+                        }
+                    } catch (RuntimeException e) {
+                        log.warning("Failed to retrieve pull request " + record.prId() + " from " + repository.name()
+                                + " for issue " + issue.id() + "; will retry the issue: " + e.getMessage());
```
erikj79: We should log the exception object here so we get the stacktrace. This can be done using `log.log(Level level, String msg, Throwable thrown)`.

erikj79 (commented):
This approach looks ok, but wouldn't it have been enough to just catch exception on line 93, log, retry and return null? I don't have a strong preference, so this is fine too.

zhaosongzs:
> This approach looks ok, but wouldn't it have been enough to just catch exception on line 93, log, retry and return null? I don't have a strong preference, so this is fine too.

Yeah, you are correct, I think your approach would be better and we can keep the stream structure, I will change to it. Thanks

bots/pr/src/main/java/org/openjdk/skara/bots/pr/IssueBot.java
```diff
@@ -90,9 +90,18 @@ public List<WorkItem> getPeriodicItems() {
...
                             .filter(r -> r.name().equals(record.repoName()))
-                            .map(r -> r.pullRequest(record.prId()))
+                            .map(r -> {
+                                try {
+                                    return r.pullRequest(record.prId());
+                                } catch (RuntimeException e) {
+                                    log.log(Level.WARNING, "Failed to retrieve pull request " + record.prId() + " from "
+                                            + r.name() + " for issue " + issue.id() + "; will retry the issue", e);
+                                    poller.retryIssue(issue);
+                                    return null;
```
erikj79: I was mistaken, this should return `Stream.empty()` not `null`. Then you shouldn't need a null check below either.

zhaosongzs: Right now I am using map() here, so I think return `null ` is correct, if we want to use `Stream.empty()`, we should use `flatMap()`

erikj79 (approved):
I was confused in my last comment and didn't realize there were two levels of mapping here. This solution works fine, as would the previous one.

bots/pr/src/main/java/org/openjdk/skara/bots/pr/IssueBot.java
```diff
@@ -90,7 +93,16 @@ public List<WorkItem> getPeriodicItems() {
...
                     .flatMap(record -> repositories.stream()
                             .filter(r -> r.name().equals(record.repoName()))
-                            .map(r -> r.pullRequest(record.prId()))
+                            .flatMap(r -> {
+                                try {
+                                    return Stream.of(r.pullRequest(record.prId()));
+                                } catch (RuntimeException e) {
+                                    log.log(Level.WARNING, "Failed to retrieve pull request " + record.prId() + " from "
+                                            + r.name() + " for issue " + issue.id() + "; will retry the issue", e);
+                                    poller.retryIssue(issue);
```
zhaosongzs: I just realized there is an issue here, we can't retry the issue before `poller.lastBatchHandled`, `poller.lastBatchHandled` will remove every issue in the current batch from the retry map including the one needs retry.

erikj79 (approved):

zhaosongzs:
Thank you!
/integrate

